PR #2, opened 2018-01-13
Possible solution to the map problem

Quantum64:
Add the same item reference to the map two times. One time keyed with the id, and another with the name.


draknyte1:
Why not store Map<meta, map<string, itemRef>>?

Quantum64:
Because that doesn't help us get the item if we don't know what the meta is.

PR #26, opened 2018-04-18
a couple of config todos

temp1011:
Just a couple of config todos I found. The range on the energy/damage use one probably isn't great. Let me know if I've done anything wrong.


Exidex:
There is discord server. Invite is in readme. 

I meant that it energy config should still be multiplier but just a int instead of double

temp1011:
Ah woops sorry. I'll fix it.

temp1011:
Thanks! I'll see what else I can do to help.

PR #90, opened 2018-06-09
reduced flint durability to 30

pyure:


Archengius:
i'm not really sure if 30 is right value - in my opinion it should be 50 at least

Archengius:
reduced to 50 myself

TedKnudsen:
Sounds good to me.

Issue #87, opened 2018-06-09
The Cheapest Anvil


EmosewaPixel:
There is a recipe that uses a single iron ingot to make an anvil. There is also one that uses 31, which is the correct one.
![image](https://user-images.githubusercontent.com/25333164/41190645-6be3b8a0-6be3-11e8-9bf4-2961ad373490.png)
![image](https://user-images.githubusercontent.com/25333164/41190654-8c7eaef8-6be3-11e8-8f43-e8459eeb1797.png)

PR #91, opened 2018-06-09
fix cheap anvil

temp1011:
Closes https://github.com/GregTechCE/GregTech/issues/87.


temp1011:
The actual recipe is just below it...

temp1011:
That's why the issue says there is both a proper and cheap recipe.

temp1011:
anvil

PR #94, opened 2018-06-09
don't register plate recipes twice

### LINKED ISSUE #89: Duplicating Plate Recipes
It's not at all a major issue. There simply are 2 identical crafting recipes for every plate.
image
image

temp1011:
Closes #89.
There is the same code twice in `OreProcessingHandler`.
Process ingot: https://github.com/GregTechCE/GregTech/blob/1a6122fefae1a99e0354a9b78bb6398b985fe36c/src/main/scala/gregtech/loaders/oreprocessing/OreProcessingHandler.java#L473-L474.

Process plate: https://github.com/GregTechCE/GregTech/blob/1a6122fefae1a99e0354a9b78bb6398b985fe36c/src/main/scala/gregtech/loaders/oreprocessing/OreProcessingHandler.java#L933-L935.

Correct me if I'm wrong.

### DIFF
--- a/src/main/scala/gregtech/loaders/oreprocessing/OreProcessingHandler.java
+++ b/src/main/scala/gregtech/loaders/oreprocessing/OreProcessingHandler.java
@@ -470,9 +470,6 @@ public class OreProcessingHandler {
                 .EUt(16).duration((int) (material.getMass() / 2L))
                 .buildAndRegister();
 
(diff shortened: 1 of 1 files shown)


temp1011:
I think the plate recipes are not the only thing that is duplicated inside this class. Someone who understands it better than me probably needs to take a look. I found one bit twice in same function. Hopefully this is right now at least.

PR #96, opened 2018-06-10
fix crash when calling with wildcard values

### LINKED ISSUE #95: Crash with buildcraft in post-init
See BuildCraft/BuildCraft#4194 and [#4195](https://github.com/BuildCraft/BuildCraft/issues/4195) - full crash report link: https://github.com/BuildCraft/BuildCraft/files/2087248/crash-2018-06-09_21.47.58-client.txt.

Essentially I think what's happening is that one of the stacks calls `BlockLogGT.func_176203_a` (`BlockLogGT.getStateFromMeta`) with a meta value equal to `OreDictionary.WILDCARD_VALUE` (32767). This is then divided by 4 to give `8191`. I think this might be fixable by changing the line from
````java
.withProperty(LOG_AXIS, EnumAxis.values()[meta / 4])
````
to
````java
.withProperty(LOG_AXIS, EnumAxis.values()[meta / 4 % EnumAxis.values().length])
````
to ensure that strange positive values for meta don't go outside the range. Alternately you could handle it  inside of 
https://github.com/GregTechCE/GregTech/blob/1a6122fefae1a99e0354a9b78bb6398b985fe36c/src/main/scala/gregtech/common/CommonProxy.java#L110-L117

and bring the value of `stack.getMetadata()` to between 0 to 15. ~~However that wouldn't handle the case where `meta = 15` as then `meta/4 = 3`, which is outside the bounds of the size-3 array.~~

temp1011:
Closes #95. The code now matches what vanilla does. From what I could tell in my world things looked the same. I hope there won't be any problems with logs suddenly being rotated :smile: .

### DIFF
--- a/src/main/scala/gregtech/common/blocks/wood/BlockLogGT.java
+++ b/src/main/scala/gregtech/common/blocks/wood/BlockLogGT.java
@@ -3,9 +3,6 @@ package gregtech.common.blocks.wood;
 import gregtech.api.GregTechAPI;
 import gregtech.common.items.MetaItems;
 import net.minecraft.block.BlockLog;
-import net.minecraft.block.BlockNewLog;
-import net.minecraft.block.BlockPlanks;
-import net.minecraft.block.properties.IProperty;
 import net.minecraft.block.properties.PropertyBool;
 import net.minecraft.block.properties.PropertyEnum;
 import net.minecraft.block.state.BlockStateContainer;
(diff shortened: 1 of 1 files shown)


temp1011:
nvm, this is incompatible with existing worlds it seems. I'm guessing the order of things in the function needs changing.

temp1011:
Should be fixed now.

PR #101, opened 2018-06-10
fix crash when tools can't be used anymore

### LINKED ISSUE #82: Crash when Mortar runs out of durability
GTCE 0.2.0.2
Forge 2707

I had a Flint Mortar that is nearly broken and I had several Copper Ingots in the Crafting Table to turn them into dust, I shift clicked so that the mortar ran out of durability half way through, I got a crash.

https://hastebin.com/kopubogama.cs

temp1011:
Closes #82. The tool won't actually break because of a bug(?) somewhere else. But it won't be able to craft anymore and shouldn't crash. The only minor issue I ran into is trying to shift click the recipe output while the craft was in progress caused some error in thaumcraft to appear in logs. But I think this is fairly minor.


temp1011:
How would you suggest to fix it? As far as I can see modifying things in containerItem just doesn't work well.

PR #102, opened 2018-06-10
Allow non electric tools to break

temp1011:
If the last use is used on a hammer etc. it will disappear. It seems that when breaking ores with a hammer if it doesn't have enough durability left to break the ore then the ore block itself will be dropped rather than the crushed stuff. I don't think this is a bad thing.


src/main/scala/gregtech/api/items/toolitem/ToolMetaItem.java
```diff
@@ -342,6 +342,9 @@ public int getInternalDamage(ItemStack itemStack) {
     private void setInternalDamage(ItemStack itemStack, int damage) {
         NBTTagCompound statsTag = itemStack.getOrCreateSubCompound("GT.ToolStats");
         statsTag.setInteger("Damage", damage);
+        if(getInternalDamage(itemStack) >= getMaxInternalDamage(itemStack)) {
+            itemStack.setCount(0);
```
Archengius: would be better to call itemStack.shrink()

PR #109, opened 2018-06-11
Fix cracking recipes

temp1011:
Before there was a bunch of spam in logs about invalid exceptions. Now everything is just registered fine. The recipes are still there in JEI. I fixed the localisation of the tab title too. I would like to add a machine for the tab but idk which one to use.


Archengius:
no, there were different ways of cracking. instead of removing them, you should have changed maxInputs to 2

temp1011:
OK, someone told me it was one-to-one on discord.

PR #114, opened 2018-06-12
Fix some smelting recipe registration

temp1011:
To register a smelting recipe it seems you have to remove the old one first. I've sorted this out for a few but for some reason not all of them work. Not sure why.

Before:
<details>

```
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.oreRedstone@0 = 5xitem.redstone@0
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.thermalfoundation.ore@2 = 1xitem.meta_item@10062
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.thermalfoundation.ore@5 = 1xitem.meta_item@10044
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.oreGold@0 = 1xitem.ingotGold@0
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.thermalfoundation.ore@3 = 1xitem.meta_item@10035
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.for.resources@0 = 4xitem.for.apatite@0
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.thermalfoundation.ore@6 = 1xitem.meta_item@10051
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.oreCoal@0 = 1xitem.coal@0
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.oreLapis@0 = 6xitem.dyePowder@4
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.netherquartz@0 = 2xitem.netherquartz@0
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.for.resources@1 = 1xitem.for.ingot_copper@0
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.thermalfoundation.ore@0 = 1xitem.for.ingot_copper@0
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.for.resources@2 = 1xitem.for.ingot_tin@0
[01:08:26] [main/INFO]: Ignored smelting recipe with conflicting input: 1xtile.thermalfoundation.ore@1 = 1xitem.for.ingot_tin@0
[...]

PR #115, opened 2018-06-12
Fix registration for most smelting recipes

temp1011:
Second try. Results in log are the same as #114.


Archengius:
i don't see why it should be added at all.

temp1011:
Did you read the errors in the log?

PR #117, opened 2018-06-12
Better multiblock pattern

Antifluxfield:
- More flexible pattern direction
- Allow aisles repeate with variable times
Using distallation tower as an example.


Archengius:
btw, can you join discord?

PR #125, opened 2018-06-12
Smelting recipe duplicate fixes

temp1011:
I'm not proud of the disgusting stream stuff, simplification or other solution appreciated. Maybe there is a list of just GTCE blocks I can use to register smelting? The logging and removing duplicate dust smelt should be fine though.


temp1011:
Hmmm, maybe I should just only call this for blockOre.

Archengius:
nope

PR #126, opened 2018-06-12
Fix rubber tree spawning in water

temp1011:
tree_in_water

I think I will look into spawning trees by hooking into terrain decorate events and replacing the tree that minecraft spawns. Might make for cleaner code.


src/main/scala/gregtech/common/worldgen/WorldGenRubberTree.java
```diff
@@ -41,6 +41,12 @@ public void generate(Random random, int chunkX, int chunkZ, World world, IChunkG
             BlockSaplingGT sapling = MetaBlocks.SAPLING;
             if(solidBlockState.getBlock().canSustainPlant(solidBlockState, world, randomPos, EnumFacing.UP, sapling)) {
                 BlockPos abovePos = randomPos.up();
+                IBlockState aboveState = world.getBlockState(abovePos);
+                if(!(aboveState.getBlock().isAir(aboveState, world, abovePos) ||
```
Archengius: Dumb check. Top block will always be top solid block, and state above can be only not solid liquid.

temp1011:
Yep, my bad

temp1011:
Hmm, I'm not sure if vanilla will generate there actually.

PR #181, opened 2018-06-30
Cobalt now generates rods

pyure:


Archengius:
I don't get any reason why it should generate rod if it's not used in crafting

PR #238, opened 2018-07-23
Cable refactor

Antifluxfield:
- Abstract the logic of cable blocks, tiles, multiparts, and e-net for further code reusage (like pipe nets).
- Fix the cable fusing. It wasn't implemented correctly:
b
You may transfer multiple 1A current from different sources using 1A cable as the picture showed above.
- Some extra feature changes and bugfixes will be submitted in another PR


Exidex:
Ideally all these changes should be in separate PRs. Because reviewing PR that has `+3,775 −2,070` changes gonna be a nightmare

Antifluxfield:
Well, I can roll back several commits of feature changes and bugfixes like this, but most of the diffs comes form the abstraction and reimplementation of the cable system, and... well, how can I split this change?

Exidex:
If nothing can be done leave it as is.

Also PRs are squashed into 1 commit when merged.

src/main/java/gregtech/api/cover/ICoverableTile.java
```diff
@@ -0,0 +1,65 @@
...
+import net.minecraft.world.World;
+import net.minecraftforge.common.capabilities.Capability;
+import net.minecraftforge.common.capabilities.ICapabilityProvider;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public interface ICoverableTile extends ICapabilityProvider {
+
+    World getWorld();
```
Archengius: Will break in obfuscated environment and crash with NoSuchMethodError, just like getPos(). Name methods in a way different from vanilla MC.

src/main/java/gregtech/api/multipart/MultipartPipeLike.java
```diff
@@ -0,0 +1,361 @@
...
+
+    ///////////////////////////////// ITEM DROPPED /////////////////////////////////////////
+
+    protected ItemStack getItem() {
+        return block.getItem(baseProperty);
+    }
+
+    @Override
+    public Iterable<ItemStack> getDrops() {//TODO Tools
+        return Collections.singleton(getItem());//TODO Covers?
```
Archengius: covers dropping is handled entirely by multipart, only thing you need to override for tools is getStrength(EntityPlayer, CuboidRayTraceResult).

src/main/java/gregtech/api/pipelike/PipeLikeObjectFactory.java
```diff
@@ -0,0 +1,298 @@
...
+    /////////////////////////////// MULTIPART METHODS //////////////////////////////////////
+
+    @Optional.Method(modid = GTValues.MODID_FMP)
+    private static void registerMultipartFactory() {
+        MultiPartRegistry.registerParts((IPartFactory) (identifier, client) -> {
+            PipeLikeObjectFactory factory = multipartFactories.get(identifier);
+            if (factory == null) return null;
+            return factory.createMultipart();
+        }, multipartFactories.keySet());
+        MultiPartRegistry.registerConverter(new IPartConverter() {
```
Archengius: Please don't use such big nested anonymous classes. It affects visibility quite bad. Better move these into some dedicated place.

src/main/java/gregtech/api/multipart/MultipartPipeLike.java
```diff
@@ -38,42 +42,47 @@
...
+    @Override
+    public PipeLikeObjectFactory<Q, P, C> getFactory() {
+        return factory;
+    }
+
     /////////////////////////////// BASE PROPERTIES ////////////////////////////////////////
 
     @Override
     public void save(NBTTagCompound tag) {
         tag.setString("material", block.material.toString());
```
Archengius: use UpperCamelCase for NBT naming

src/main/java/gregtech/api/worldobject/PipeNet.java
```diff
@@ -0,0 +1,396 @@
...
+            }
+        }
+        return false;
+    }
+
+    private void addPosToArea(BlockPos pos) {
+        area.put(new ChunkPos(pos), pos);
+    }
+
+    class MutableChunkPos extends ChunkPos {
```
Archengius: same applies here. move this class to utils package

src/main/java/gregtech/api/cover/ICoverableTile.java
Antifluxfield: Yep, I know, and this is fixed in commit 637ce6e59b5be

src/main/java/gregtech/api/pipelike/PipeLikeObjectFactory.java
Antifluxfield: Got it.

src/main/java/gregtech/api/multipart/MultipartPipeLike.java
Archengius: what do you mean by covers? I thought you meant FMP covers. IF you're talking about some GT-ish covers, i don't really seem them needed anyhow.

src/main/java/gregtech/api/cover/ICoverableTile.java
Archengius: you should push all things in one PR, that's the point. Otherwise it won't merge properly.

src/main/java/gregtech/api/multipart/MultipartPipeLike.java
Antifluxfield: I mean GT-ish covers. Well, just ignore it.

src/main/java/gregtech/api/cover/ICoverableTile.java
Antifluxfield: Hmmm... You mean I should put those commits in this pr again?

PR #239, opened 2018-07-23
Cable refactor 2

Antifluxfield:
Following by #238 , including:
- Cleanup & bugfix
- You should be able to color the multipart cables, though our spray behaves strangely...
- You can now directly put the cables into a multipart block
- Tool requirement for multipart cables.
- Improved TOP compact of cables.


Exidex:
Please don't do this. As I understand this PR is supposed to be merged after the first one but now if you'll make changes to the first one you'll need to make same changes to this one as well or it will be overridden. For now I'll close this PR and after first one is merged create followup pr.

Antifluxfield:
got it.

Antifluxfield:
@Exidex Seems it would be better to append these commits back to #238 , because there'e quite some important bugfixes & rewrites here, so currently #238 is totally not an complete version.

PR #267, opened 2018-08-14
Recipes, and other misc things

Antifluxfield:
- Recipes for pipes
- Change PE into ingot material
- Electric damage of bare wires
- Some l10n support
- Crafttweaker support of pipes & cables customization


Archengius:
no more

Issue #311, opened 2018-09-09
Twitch doesn't import ".worldgen_extracted" file properly


pyure:
As far as I can tell, the **".worldgen_extracted"** is a flag to prevent the regeneration of default worldgen.

Because the file doesn't import, packs are unable to ship custom worldgen without it simply getting destroyed.

After testing, our best guess is that it is (empty?) files beginning with "." which cause the issue.  Removing the "." does allow the file to import (but of course doesn't help prevent the destruction of the custom worldgen)

If confirmed, can we rename the file to remove the dot prefix?

pyure:
I should also clarify: this is only tested with packs that are shipped via the **Create Custom Profile -> Import** process in the twitch app.  I do not know explicitly if it impacts packs installed traditionally via twitch.

Rongmario:
Delete twitch

PR #312, opened 2018-09-09
renamed worldgen extract lock file to be twitch-import friendly

pyure:
Just allows any twitch users to leverage the custom-worldgen capabilities.  If the "." is present, these users will have their custom worldgen destroyed.

Only impacts twitch users as far as I know.

### DIFF
--- a/src/main/java/gregtech/api/worldgen/config/WorldGenRegistry.java
+++ b/src/main/java/gregtech/api/worldgen/config/WorldGenRegistry.java
@@ -93,13 +93,15 @@ public class WorldGenRegistry {
         oreVeinCache.clear();
         Path configPath = Loader.instance().getConfigDir().toPath().resolve(GTValues.MODID);
         Path worldgenRootPath = configPath.resolve("worldgen");
+        Path jarFileExtractLockOld = configPath.resolve(".worldgen_extracted");
         Path jarFileExtractLock = configPath.resolve("worldgen_extracted");
         if(!Files.exists(worldgenRootPath)) {
             Files.createDirectories(worldgenRootPath);
         }
         //attempt extraction if file extraction lock is absent or worldgen root directory is empty
-        if(!Files.exists(jarFileExtractLock) || !Files.list(worldgenRootPath).findFirst().isPresent()) {
-            if(!Files.exists(jarFileExtractLock)) {
+        // todo: "Old" is fugly.  Deprecate and/or replace with pattern-match on name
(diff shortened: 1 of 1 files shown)


pyure:
Pertains to https://github.com/GregTechCE/GregTech/issues/311.  SC, Exidex, don't merge this prematurely.  Arch has indicatedusers should "delete twitch".

pyure:
Good point.

pyure:
@Exidex I took a brief look at the Files class and didn't see a simple way to do pattern matching.

src/main/java/gregtech/api/worldgen/config/WorldGenRegistry.java
```diff
@@ -93,13 +93,15 @@ public void reinitializeRegisteredVeins() throws IOException {
...
         Path jarFileExtractLock = configPath.resolve("worldgen_extracted");
         if(!Files.exists(worldgenRootPath)) {
             Files.createDirectories(worldgenRootPath);
         }
         //attempt extraction if file extraction lock is absent or worldgen root directory is empty
-        if(!Files.exists(jarFileExtractLock) || !Files.list(worldgenRootPath).findFirst().isPresent()) {
-            if(!Files.exists(jarFileExtractLock)) {
+        // todo: "Old" is fugly.  Deprecate and/or replace with pattern-match on name
+        if((!Files.exists(jarFileExtractLock) && !Files.exists(jarFileExtractLockOld)) || !Files.list(worldgenRootPath).findFirst().isPresent()) {
+            if(!Files.exists(jarFileExtractLock) && !Files.exists(jarFileExtractLockOld)) {
```
Archengius: Wrong. Second if condition here shouldn't be the case.

Archengius:
and i won't take my words back. This is last time i fix twitch issue on my side. Comments like here will just make it happen with higher percentage.

src/main/java/gregtech/api/worldgen/config/WorldGenRegistry.java
pyure: Archengius: that's actually your own secondary if condition, sir.  If we remove it, then we need to also remove your OR logic that checks for an empty folder and creates the file.

Otherwise, if Files.list(worldgenRootPath).findFirst().isPresent() qualifies, we'll attempt to create the file.  If you're ok with that, I'll remove the condition.

Archengius: do you understand that with !Files.exists(jarFileExtractLockOld)) { it will not generate new lock file when there is deprecated one, resulting in worldgen extraction when support for deprecated file is removed,

PR #421, opened 2018-10-29
Add ability to disable GT ore gen on a per dimension basis

zmaster587:
Background:
There's a handful of mods that add dimensions to minecraft that are considered surface worlds.  In some cases modpack authors or users want to have gregtech generate its ore in the overworld and not in these dimensions added by these other mods.
Gregtech currently does not currently appear to fire an `OreGenEvent.GenerateMinable` event when checking to generate ores.  Additionally doing so might make disabling ore gen painful.

To solve this problem, I'd like to propose a config option that is a list of dimension ids where the user does not want gregtech to do oregen.  Where the values are checked against the dimid in which generation is about to occur in `gregtech.api.worldgen.generator.generate`.

I don't like the idea of iterating over a loop in `generate` to check the dimids, but unless the list is very large the performance impact is expected to be pretty small.

Thanks for your time!

### DIFF
--- a/src/main/java/gregtech/api/worldgen/generator/WorldGeneratorImpl.java
+++ b/src/main/java/gregtech/api/worldgen/generator/WorldGeneratorImpl.java
@@ -40,14 +40,21 @@ public class WorldGeneratorImpl implements IWorldGenerator {
     @SubscribeEvent(priority = EventPriority.HIGH)
     public void onOreGenerate(OreGenEvent.GenerateMinable event) {
         EventType eventType = event.getType();
+        World world = event.getWorld();
         if(ConfigHolder.disableVanillaOres &&
-            ORE_EVENT_TYPES.contains(eventType)) {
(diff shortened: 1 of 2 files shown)


Rongmario:
Or... You could just delete the jsons in the worldgen config folder...

Or if you want to change overworld ore veins to only spawn in 0, just assign in to 0.

zmaster587:
Deleting the jsons on the world folder would prevent the veins from spawning in the overworld.  The use case is for spawning veins in dimension 0 and dimensions 2 though 75, but not in 53, or 42, and having a non time consuming way of turning it off by dimension id.

Rongmario:
...Therefore you can assign which dimension you want individual ore veins to generate at.

zmaster587:
...If the dimensionType or the provider is different.  There are instances where mods create many worlds with the same provider and same dimension type; all of them surface worlds.

PR #430, opened 2018-11-05
Update zh_cn.lang

TartaricAcid:
This translation was originally produced by @Antifluxfield , but it is quite different from @Timozer. 
I don't know how to deal with it, so I hope that before merge, @Antifluxfield and @Timozer can make a check on it.

PR #431, opened 2018-11-05
update zh_cn.lang

Timozer:
fix some errors


Antifluxfield:
I've read your translation. Comparing to the the translation in #430 , your translation is:
* Incomplete. Many translatable (I mean, doesn't contain slang or memes that make it really hard to translate) entries are not translated.
* Containing extensive bad machine translation. Some entries are ungrammatical in Chinese. Some entries are literally correct but doesn't make sense in their place. And the most ludicrous example is that you translate a "chest buffer" to ["胸部缓冲液"](https://github.com/Timozer/GregTech/blob/master/src/main/resources/assets/gregtech/lang/zh_cn.lang#L2552) (something like "breast buffer liquid").
* Containing errors. [Darmstadtium is not tantalum](https://github.com/Timozer/GregTech/blob/master/src/main/resources/assets/gregtech/lang/zh_cn.lang#L1119). [You can't use a honey comb to comb your hair](https://github.com/Timozer/GregTech/blob/master/src/main/resources/assets/gregtech/lang/zh_cn.lang#L968). Several other items are translated to some unrelated things, etc.
* Inconsistent. Including that: [same item](https://github.com/Timozer/GregTech/blob/master/src/main/resources/assets/gregtech/lang/zh_cn.lang#L127) having [different translations](https://github.com/Timozer/GregTech/blob/master/src/main/resources/assets/gregtech/lang/zh_cn.lang#L984); [same translation](https://github.com/Timozer/GregTech/blob/master/src/main/resources/assets/gregtech/lang/zh_cn.lang#L1396) referring to [different items](https://github.com/Timozer/GregTech/blob/master/src/main/resources/assets/gregtech/lang/zh_cn.lang#L1397); translation doesn't make sense after substituting the placeholders; same sentence schema translated to different patterns; some entries are not using their habitual words. Well, the last two may not be that severe, but they're really annoying.
* Other minor issues.

Of course, when reviewing #430 , I do discovered some typos, and you had some better ideas on a few entries; but generally, I think that your translation still have a long way to go.

Archengius:
outdated

PR #433, opened 2018-11-07
Chinese translation

Antifluxfield:
A complete and up-to-date version inheriting habitual translation of GT5u.
Based on #430 , corrected & improved several entries.
Fixed some errors in en_us


Archengius:
outdated

PR #458, opened 2018-11-27
fix #454 use isAir instead of Blocks.AIR comparison

### LINKED ISSUE #454: Multiblock not forming with TorchMaster light inside
I use the "Feral Flare Latern " by TorchMaster to illuminate my base. The problem is this mod spawns lights in dark areas like inside of multiblocks. This leads to invalid structures cause inside the multiblocks is now an other block then air.... (took hours to find that reason btw)

Never tested it with Arcane Lamp by Thaumcraft in 1.12. but i used the Arcane Lamp in 1.7.10 with gregtech 5 unofficial and had never problems.

Versions:
gregtech-1.12.2-1.0.124
torchmaster_1.12.2-1.6.4.62

before
after

leagris:


src/main/java/gregtech/api/metatileentity/multiblock/MultiblockControllerBase.java
```diff
@@ -115,6 +116,10 @@ public final int getLightValue() {
         return blockWorldState -> ArrayUtils.contains(block, blockWorldState.getBlockState().getBlock());
     }
 
+    public static Predicate<BlockWorldState> isAirPredicate() {
+        return blockWorldState -> blockWorldState.getBlockState().getBlock().isAir(blockWorldState.getBlockState(),null,null);
```
Archengius: passing nulls to method excepting world-sensitive info is not how it works.

leagris: These parameters are advertised as nullable and not used to check a block is air
image

leagris:
I updated my commit with your requested changes:
I had to provide blockWorldState.getWorld() method and made use of it in my call of isAir()

PR #462, opened 2018-12-18
Fixed server-side localized item names and tooltips for MetaItems/Blocks/Tools

YuRaNnNzZZ:
This pull request addresses issues with other mods which handle item names server side, like Quark and its "Item Linking" feature.
Before fix:

After fix:

(you can safely ignore the I18n deprecation warnings, vanilla code uses it in many places)


Archengius:
No. GTCE will never use server-side localization as it works incorrect and will be removed pretty soon. It's on mods using server-side localization to implement it properly, not on other mods to do stupid things just to make their shit work without issues.

Archengius:
You can implement your own ITextComponent and register serializer for it if you really want to display item stack names in chat that way. On the other hand, simpler solution would be just broadcasting packet to all players and translating item stack name on client side, and then printing message.

Issue #468, opened 2018-12-27
[Suggestion] Only re-lookup machine recipes when items/fluids change.


omsfuk:
I took a look at your code and found that most of the code is elegant but there are also some questions. For example the recipe lookup uses a linear lookup method. When the recipe table is large, such as an assembler which contains 2000+ recipes and cause a lot of overhead (my test also proves this, Especially when there are items inside the assembly machine). So why not use binary search, just need to sort the recipe, the cost is only a little more loading overhead.

This is just my opinion or suggestion. please indicate if there is any error. 
Anyway, I hope this mod will be better.
Sorry for my poor english : )
as a GregTechCE fan

Archengius:
You cannot sort recipe by inputs, because inputs are not ItemStack but just Predicate<ItemStack> with any complex or simple logic behind it. RecipeMap has no knowledge of possible combinations and values of these predicates, so the only thing it can do for searching recipes is linear lookup with matching every recipe's input predicates.
Taking in account all machines cache their recipe for future usage, i don't really see the problem here. Looping through 2k elements is pretty fast too.
However, PRs and ideas how it can be optimized while keeping things as dynamic as they are now, are welcome.

omsfuk:
Thank you for your reply.
* But this still has problems. The cache only works when the recipe exists. I know once the machine is running, there will be no such problems.But what if the recipe doesn't exist as well as the machine is NOT Running?
If the machine is not running, it will try to check the inventory at regular intervals, thus causing the need to search the recipes frequently, and the cache will not hit.

* Due to the automation of Machine Hull, I need to keep a configuration circuit in the assembler。When I use LagGoggle to watch the TileEntiy which take more tick time, I found the assembler was the most.
* You can reproduce it by following the steps below.
	* Place an assembly machine
	* Place a configuration circuit in the assembler  with any number
	* Place a assembler with empty inventory
	* Place a chemical reactor
	* Open LagGoggle to get the results
	* Compree the result

(And I don't think this was a mistake by LagGoggle)
you will found the assember takes up a lot of tick time.
![2018-12-27_19 28 01](https://user-images.githubusercontent.com/26737684/50478998-30248e00-0a0f-11e9-9b4a-0ced698c9abf.png)
(Please ignore the Chinese : )

* If you agree with what I said above, let's start the next topic : )
So why not try to split the recipes , for each possible list of a separate recipe, ranther than "Predicate"?
[...]

PR #518, opened 2019-01-18
Temporarily fix for #468, need for more tests.

Takakura-Anri:
Current potential issues: 
- [ ] Insuffcient energy supply will cause not working
- [ ] Exist fluid tanks occasionally crash in Gui Interaction, can be fixed by re-placing the block


Takakura-Anri:
Need further optimization work on recipe finding algorithm to ultimately solve this issue

PR #570, opened 2019-02-16
ellipsoid generator height property

pyure:
* Allows user to specify height of generated ellipsoid
* Does NOT correct a minor radius bug (will PR that separately if/when this gets merged)
* If no height specified (properly) in json, we fall back to original logic
* Tested to confirm fallback generates original ellipsoids
* Optimized away one "continue" clause: it was mathematically impossible to reach it.

### DIFF
--- a/src/main/java/gregtech/api/worldgen/shape/EllipsoidGenerator.java
+++ b/src/main/java/gregtech/api/worldgen/shape/EllipsoidGenerator.java
@@ -1,5 +1,6 @@
 package gregtech.api.worldgen.shape;
 
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import crafttweaker.annotations.ZenRegister;
 import gregtech.api.worldgen.config.OreConfigUtils;
@@ -14,6 +15,8 @@ public class EllipsoidGenerator extends ShapeGenerator {
 
     private int radiusMin;
     private int radiusMax;
+    private int heightMin;
+    private int heightMax;
 
     public EllipsoidGenerator() {
     }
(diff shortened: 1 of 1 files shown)


src/main/java/gregtech/api/worldgen/shape/EllipsoidGenerator.java
```diff
@@ -27,31 +30,47 @@ public EllipsoidGenerator(int radiusMin, int radiusMax) {
...
-        this.radiusMin = data[0];
-        this.radiusMax = data[1];
+        int[] radiusData = OreConfigUtils.getIntRange(object.get("radius"));
+        this.radiusMin = radiusData[0];
+        this.radiusMax = radiusData[1];
+
+        JsonElement heightJson = object.get("height");
+        if (heightJson != null) {
+            int[] heightData = OreConfigUtils.getIntRange(heightJson);
+            if (heightData.length == 2) {
```
Archengius: always evaluates to true

src/main/java/gregtech/api/worldgen/shape/EllipsoidGenerator.java
```diff
@@ -27,31 +30,47 @@ public EllipsoidGenerator(int radiusMin, int radiusMax) {
...
-        int b = radiusMin >= radiusMax ? radiusMin / 2 : (gridRandom.nextInt(radiusMax - radiusMin) + radiusMin) / 2;
-        int c = radiusMin >= radiusMax ? radiusMin : (gridRandom.nextInt(radiusMax - radiusMin) + radiusMin);
+        int b;
+        if (heightMin == 0) {
+            // default if no height prop provided (also: originally coded behavior)
+            b = radiusMin >= radiusMax ? radiusMin / 2 : (gridRandom.nextInt(radiusMax - radiusMin) + radiusMin) / 2;
+        } else {
+            b = heightMin >= heightMax ? heightMin : (gridRandom.nextInt(heightMax - heightMin) + heightMin);
+        }
+        int c = a; // somewhat pointless, but establishes pattern for other (non-symmetrical) shapes.  Ellipsoids will always be circular as viewed from above.
```
Archengius: You understand that you're breaking original generator by forcing same size on both axis, right? Originally, it randomly chose radius by X, Y, Z axes separately, forming ellipsoid shape, and you made it sphere by forcing X to equal Z.
Lookup and compare what you got old way and the new one. Ellipsoid doesn't look like a sphere and is not forced to be circular when viewed from above. Only body forced to be circular when viewed from above is sphere.

Archengius:
dead

PR #585, opened 2019-02-22
Added translation of new materials and some corrections of Russian translation

WoozyMasta:


BlueCreeper160:
That's a good idea. And Why not for other languages? (i can help with spanish)

WoozyMasta:
@BlueCreeper160 what is the problem? Prepare the translation and send the PR if the quality will be acceptable your translation will add to the project.

PR #647, opened 2019-03-16
Fix incorrect PBF fuel values, fix NPE with PBF fuel slot

codewarrior0:
Fixes an issue where clicking a non-fuel item into the fuel slot results in a crash, and fixes an issue where tiny piles of dust count as full fuel items (and blocks count as single items).


Archengius:
Bad way of fixing issue.

PR #751, opened 2019-04-15
Add a Flood Gate

wishAI:


src/main/java/gregtech/common/metatileentities/electric/MetaTileEntityFloodGate.java
```diff
@@ -0,0 +1,306 @@
...
+            .widget(new FluidContainerSlotWidget(importItems, 0, 90, 17, false)
+                .setBackgroundTexture(GuiTextures.SLOT, GuiTextures.IN_SLOT_OVERLAY))
+            .widget(new ImageWidget(91, 36, 14, 15, GuiTextures.TANK_ICON))
+            .widget(new SlotWidget(exportItems, 0, 90, 54, true, false)
+                .setBackgroundTexture(GuiTextures.SLOT, GuiTextures.OUT_SLOT_OVERLAY))
+            .bindPlayerInventory(entityPlayer.inventory);
+        return builder.build(getHolder(), entityPlayer);
+    }
+
+    private void updateQueueState() {
```
Archengius: This function entirely won't work with unlimited radius.
Needs rewriting, most probably via grouping queue by chunks to make filling easier and checking chunk loaded/unloaded faster.

PR #781, opened 2019-04-28
Fixes crash caused by null value for side

Kurtchekov:
This function can crash the game if EnumFacing side is null.


Archengius:
because it doesn't expect parameter being null and in fact shouldn't

Kurtchekov:
Then why are you using @Nullable annotation? Shouldn't you remove that instead?

Archengius:
this method's parameter is not annotated by @Nullable, which means it's not nullable

Kurtchekov:
Well, I guess I'll have to keep using a modified version of this mod then, since in Gregblock modpack having redstone signal on top of the rock breaker added by Gregic Additions causes the game to crash and there's no way to override that method because it's final.

Archengius:
Play GregBock on your own peril, I don't give a shit about unmaintained packs which run 100 years old version of gtce.

Kurtchekov:
The version of Gregblock I'm playing currently runs GTCE 1.7.0.387 from Apr 10, 2019. If my math doesn't fail me, you've miscalculated by 99 years and 342 days, unsurprisingly, considering how much you assume things. There's only one more recent version of your wondrous mod, which I tried to update to, but it's incompatible with Gregic Additions, GTCE Bees and possibly more because some "programmer" thinks it's proper coding to change the signature of functions every few versions without warning, breaking mods that depend on it.
Seriously, I've come here, to try and help you fix this mess, and that's how you treat me? Might as well change the name of your mod to not include "Community Edition", unless you mean it as an insult to your own playerbase.

Archengius:
I warned most of add-on authors after making that change, GA is abandoned and nobody responded. Updates come with big time gap, add-ons had enough time to update (unless author abandoned them).
What you call "mess" was fixed in latest release long time ago, your fix is absolute shit because it masks problem by just making function not crash on meaningless value. Looks like the real "programmer" solution.

PR #791, opened 2019-05-01
Allow screwdriving the output side using the cover hitbox

codewarrior0:


Archengius:
no, it changes exactly side you're pointing at, to prevent accidental changing

codewarrior0:
I just had to screwdriver almost one hundred machines after the latest update. None of them had the output side accessible. Without this patch, I would have had to use a wrench to move the output side to where I can see it, use the screwdriver to toggle "allow input", and then wrench it again to get the output side back to where it was.

And even if it's accidentally changed, it can only ever be toggled on and off, and you even get a helpful chat message to tell you that. Just whack it again if you change it on accident!

Issue #529, opened 2019-01-20
[Suggestion] Only try to "extract" energy when machine need it


omsfuk:
After fix #468 , the laggest machine in GTCE become Transformer or BatterBuffer with a lot of "route node" (to be exactly, RoutePath)

In the following example, a transformer linked to 8 not working machine. Transformer takeup 18us, but machine just 0us or 1us
![2019-01-21_01 42 02](https://user-images.githubusercontent.com/26737684/51442983-e4b78400-1d1d-11e9-99c6-33af303d761f.png)

passive energy transmission is better than active energy transmission.
That means, "working" machine try to extract energy on their own initiative, other than Transformer or BatteryBuffer active push energy to machines.

Because in mose cases, most machine was not working, only a few are working. Only working machine need energy. In many cases, transformers don't need to work at all.

In my world, machine not working take up about 1 us, but a transformer take up about 20us, even no machine consume energy.

From my last failed pull request #519 , I know my changes may not suit your taste, so I will not try to submit my code as well as pull request anymore. I think it would be better for you to solve this problem.

Again, sorry for my bad english : )
Anyway, I just want GTCE to be better & TPS friendly

Archengius:
Pulling energy from network is not how things should work, and iterating 10 routes is not that heavy.
Nope.

omsfuk:
One moment please, but it is really a problem. my BatteryBuffer linked with more than 50 tranformer takes up about 200us. This is totally an unnecessary overhead. Passive energy transfer does work.
Thank you for your reply

Archengius:
Passive energy transfer causes even more bugs in current tricky design in e-net and introduces more difficulties.
Don't link battery buffer to 50 transformers, also.

omsfuk:
Oh, please forgive my ignorance, thank you :)

GregTecher:
I came up with the idea of expanding the machine's energy buffer and changing energy transmission interval, for example, 0.5s
For example, the IV transformer, which was originally transmitted per Tick (4A x 8192V) eu, 
is now changed to every 10Tick transmission (10 x 4A x 8192V) eu, which may cause some minor problems, 
but I think that these problems can be ignored compared to the benefits it brings.

PS: I have to think about this again, because half of my TPS is now being taken away by the transformer. 
I have about 200 transformers in the whole base. Even if each transformer only takes 4us, it will be a big one. 
The overhead, especially in the server. 

[...]

PR #815, opened 2019-05-09
added idle timeout to update method

krakel:
At first, your code looks very well from my job point of view! Good implementation.

**Now my hack:**

Inside the AbstractRecipeLogic.java class the recipe will be checked at the update() method.
The main parameter progressTime control the actions.

progressTime > 0: updateRecipeProgress() with recipe recalculation after maxProgressTime
progressTime == 0: trySearchNewRecipe() with recipe search every tick

The seaching for valid recipes was using expensive calculations with recipe.matches() and findRecipe().

The new parameter recipeIdleTime delay the recipe recalculation.
I added a maxIdleTime to the ConfigHolder with a default of 10 (0.5s).

A better choice should be to eleminate the recipe search from the update() method and adding this behaviour to the fluid/item input and output handler. I was not diving deep enough in your code to find good modifications.

**Important!**
I had no test environment to verify the effect.

### DIFF
--- a/src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
@@ -44,6 +44,7 @@ public abstract class AbstractRecipeLogic extends MTETrait implements IWorkable 
     protected NonNullList<ItemStack> itemOutputs;
     protected final Random random = new Random();
 
+    private int recipeIdleTime;
     private boolean isActive;
     private boolean workingEnabled = true;
     private boolean hasNotEnoughEnergy;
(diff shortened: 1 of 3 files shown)


Archengius:
Recipe lookup occurs only on inventory contents change. Manage to at least read all code before submitting PRs.

krakel:
Ok, I see you check all Inputs with checkRecipeInputsDirty(), but the hole trySearchNewRecipe() do'nt need happen on every tick. On a server with many machines the update() methode should be at short at possible.
I hope you find a better solution for the big idle time load.

Archengius:
it's not big, only thing trySearchNewRecipe is extremely lightweight.

omsfuk:
@krakel If you actually test it, you will find that the current performance bottleneck is not recipe at all, but energy transmission. I've improved it in my fork and it works very well, however, Archengius don't agree with me : )

john01dav:
@omsfuk I can't find your fork on your github repository list. I run a server and a fork with performance enhancements would be *very* useful, if we can make it compatible with the modpack version (we probably can, with 3 Java developers on the staff team). Is your fork available somewhere publicly?

Archengius:
Issue which this fork would badly fix is fixed already in latest release

omsfuk:
> @omsfuk I can't find your fork on your github repository list. I run a server and a fork with performance enhancements would be _very_ useful, if we can make it compatible with the modpack version (we probably can, with 3 Java developers on the staff team). Is your fork available somewhere publicly?

here, https://github.com/GregTecher/GregTech
As @Archengius says, the recipe check problem was already fixed in latest release.
But I tried to optimize the energy transfer, and this was reject by Archengius.
My fork may cause some problems, see https://github.com/GregTechCE/GregTech/issues/529#issuecomment-464846995
https://github.com/GregTechCE/GregTech/issues/529#issuecomment-466622597

Archengius:
it's just too late to change energy API that much at that point

PR #824, opened 2019-05-13
Added a Fisher

### LINKED ISSUE #499: [Utility Suggestion] Fisher
When placed above water gets items from the fishing loot table over time. Luck and speed increases over time, so at LV you'll probably be getting some junk. (I personally think they shouldn't require bait)

BxFS:
First time doing this, sorry if my code looks terrible and/or too copypasty.
Closes #499 if accepted.

### DIFF
--- a/src/main/java/gregtech/common/metatileentities/MetaTileEntities.java
+++ b/src/main/java/gregtech/common/metatileentities/MetaTileEntities.java
@@ -486,6 +486,11 @@ public class MetaTileEntities {
         TITANIUM_TANK = GregTechAPI.registerMetaTileEntity(815, new MetaTileEntityTank(gregtechId("titanium_tank"), Materials.Titanium, 48000));
         TUNGSTENSTEEL_TANK = GregTechAPI.registerMetaTileEntity(816, new MetaTileEntityTank(gregtechId("tungstensteel_tank"), Materials.TungstenSteel, 64000));
 
+        FISHER[0] = GregTechAPI.registerMetaTileEntity(820, new MetaTileEntityFisher(gregtechId("fisher.lv"), 1));
+        FISHER[1] = GregTechAPI.registerMetaTileEntity(821, new MetaTileEntityFisher(gregtechId("fisher.mv"), 2));
+        FISHER[2] = GregTechAPI.registerMetaTileEntity(822, new MetaTileEntityFisher(gregtechId("fisher.hv"), 3));
+        FISHER[3] = GregTechAPI.registerMetaTileEntity(823, new MetaTileEntityFisher(gregtechId("fisher.ev"), 4));
+        
         for (int i = 1; i < 5; i++) {
             String voltageName = GTValues.VN[i].toLowerCase();
             PUMP[i - 1] = new MetaTileEntityPump(gregtechId("pump." + voltageName), i);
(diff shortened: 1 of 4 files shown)


src/main/java/gregtech/common/metatileentities/electric/MetaTileEntityFisher.java
```diff
@@ -0,0 +1,135 @@
...
+        return builder.build(getHolder(), entityPlayer);
+    }
+
+    public void update() {
+        super.update();
+        if (!getWorld().isRemote && energyContainer.getEnergyStored() >= energyAmountPerFish && getTimer() % fishingTicks == 0L &&
+            importItems.getStackInSlot(0).getCount() > 0) {
+            WorldServer world = (WorldServer) this.getWorld();
+            BlockPos waterCheckPos = getPos().down();
+            if (world.getBlockState(waterCheckPos).getBlock().getRegistryName().toString().equals("minecraft:water")) {
```
Archengius: would be better to check amount of water blocks, single-puddle fisher is weird imo
Also don't use registry names for block equality, use
`blockState.getBlock() instanceof BlockLiquid && blockState.getMaterial() == Material.WATER` here

src/main/java/gregtech/common/metatileentities/electric/MetaTileEntityFisher.java
```diff
@@ -0,0 +1,135 @@
...
+                LootTable table = world.getLootTableManager().getLootTableFromLocation(new ResourceLocation("minecraft:gameplay/fishing"));
```
Archengius: Don't use string constants for resource names, there's LootTables containing resource locations of all vanilla loottables

src/main/java/gregtech/common/metatileentities/electric/MetaTileEntityFisher.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    @Override
+    protected IItemHandlerModifiable createImportItemHandler() {
+        return new ItemStackHandler(1) {
+            @Nonnull
+            @Override
+            public ItemStack insertItem(int slot, @Nonnull ItemStack stack, boolean simulate) {
+                if(stack.getItem().getRegistryName().toString().equals("minecraft:string")) {
```
Archengius: same here - do not use registry name for block checks, and using Ore Dictionary would be better here
`OreDictUnifier.getOreDictNames(itemStack).contains("string")`

src/main/java/gregtech/common/metatileentities/MetaTileEntities.java
```diff
@@ -495,6 +496,8 @@ public static void init() {
 
         TESLA_COIL = new MetaTileEntityTeslaCoil(gregtechId("tesla_coil"));
         GregTechAPI.registerMetaTileEntity(1001, TESLA_COIL);
+        FISHER = new MetaTileEntityFisher(gregtechId("fisher"));
+        GregTechAPI.registerMetaTileEntity(1050, FISHER);
```
Archengius: Multiple tiers would be better here, just like most of machines do

src/main/java/gregtech/common/metatileentities/electric/MetaTileEntityFisher.java
```diff
@@ -0,0 +1,135 @@
...
+import net.minecraftforge.items.IItemHandlerModifiable;
+import net.minecraftforge.items.ItemHandlerHelper;
+import net.minecraftforge.items.ItemStackHandler;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.List;
+
+
+public class MetaTileEntityFisher extends MetaTileEntity {
```
Archengius: As i said earlier, better make it tiered, that way you only need to extend TieredMetaTileEntity
and not mess with energy container yourself

src/main/java/gregtech/common/metatileentities/electric/MetaTileEntityFisher.java
```diff
@@ -0,0 +1,135 @@
...
+        };
+    }
+
+    @Override
+    protected IItemHandlerModifiable createExportItemHandler() {
+        return new ItemStackHandler(inventorySize);
+    }
+
+    public void renderMetaTileEntity(CCRenderState renderState, Matrix4 translation, IVertexOperation[] pipeline) {
+        Textures.SOLID_STEEL_CASING.render(renderState, translation, pipeline);
```
Archengius: for tiered machines you also don't need that

BxFS:
Well, forgot to leave a note that I've completed required changes. Please review when you have time.

PR #872, opened 2019-06-27
Improve some displayed Informations

Blood-Asp:
- Add decimal seperators for long numbers
- Add per Bucket Fuelvalue to Fuel JEI page


karikamiya:
Please adhere to the international standard ISO 80000-1:2009: numbers should be grouped in threes with spaces, not decimals or commas since both of those are reserved for the decimal place.

123,456.00 No
123.456,00 No
123 456.00 Yes
123 456,00 Yes

Archengius:
no, it looks shitty

Archengius:
and in java dot is used for decimal, just like everywhere else

Exaxxion:
I definitely welcome digit separators as some of the longer numbers become difficult to read. I find it a bit strange that you would revert display of byproduct percentages per macerator tier though. It seems like a useful feature: until I was told about it specifically I didn't realize they even do increase, as JEI only showed the MV byproduct chances.

Archengius, Java natively supports Locale formatting for internationalization purposes. This functionality adheres to standards like the one referenced by karikamiya. Dot as separator is a part of the language specification and affects the programmer, not the end user. 

Like yourself, I am in a region where commas and dot are used for thousands separation and decimal point, respectively, so I agree that subjectively the other formats seem strange. They are however correct in their respective locales. You would not be affected as your system Locale would produce the format you expect, and people in other regions would have the formatting they expect.

leagris:
Grouping and decimal sep are two differnt locale settings or preference.
In the *nix world or POSIX standard http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap07.html#tag_07_03_04

You can check what it has at your system in a Bash terminal:
**en_US.UTF-8** locale
```Bash
LC_ALL=en_US.UTF-8 locale -ck LC_NUMERIC
```
Output:
```
LC_NUMERIC
decimal_point="."
thousands_sep=","
grouping=3;3
numeric-decimal-point-wc=46
numeric-thousands-sep-wc=44
numeric-codeset="UTF-8"
```
**fr_FR.UTF-8** locale
```Bash
LC_ALL=fr_FR.UTF-8 locale -ck LC_NUMERIC
```
Output:
```
LC_NUMERIC
decimal_point=","
thousands_sep=" " # non-breaking short space
grouping=3
numeric-decimal-point-wc=44
numeric-thousands-sep-wc=8239
numeric-codeset="UTF-8"
```

**POSIX** | **C** locale
```Bash
LC_ALL=POSIX locale -ck LC_NUMERIC
```
Output:
```
LC_NUMERIC
decimal_point="."
thousands_sep=""
grouping=-1
numeric-decimal-point-wc=46
numeric-thousands-sep-wc=0
numeric-codeset="ANSI_X3.4-1968"
```

As you can see the grouping and thousands separators vary significantly between locales.
And not all of these characters are supported within even Minecraft locales.
Also lots of the floating point ⇔ string conversion libraries totally ignores the grouping and thousands separator rules or roll their own.

If you really need large number to remain meaningful, I suggest you change the unit every 10³ K M G T P Z and avoid growing the displayed numbers too much.
Alternatively offer a mean to display the entire number at unit 1 when hovering the mouse with all the digits, if this really matters.

Blood-Asp:
I did talk with Arch a few days back. Will update this PR to use 10³ unit steps.
Thinking about it, would it be correct to use Kilo, Million, Billion, Trillion or Kilo, Mega, Giga, Terra,..

Also the tooltip of the Macerator chance was reverted after Archs request. The mechanic is still up to change and he didn't really like the implementation.

Archengius:
How does the end result look like?

Blood-Asp:
image
image
image

LAGIdiot:
I am closing this PR as stale. I may later return to it and reopen it or reuse it in some form.

PR #904, opened 2019-07-11
Fix #902 

### LINKED ISSUE #902: Fluid filter cover works improperly
Steps to reproduce the problems:
1. Set up a centrifuge to convert air into nitrogen and oxygen.
2. Attach a fluid filter cover on one side, and set it in drain & whitelist mode.

And there are two problems:
1. When oxygen is in whitelist, it can't be extracted using any fluid pipes from other mods, like enderio, mekanism.
2. When nitrogen is in whitelist, it can be extracted using mekanism pipes, while enderio's can't.

I find the problems caused by some mistakes in the implementation of FluidHandlerFiltered's drain() method, and I'll submit a PR to fix it.

cabbby:
#902


Archengius:
Did you test it? Casting delegate to FluidHandlerProxy will throw exception when using filter on fluid pipes and isn't necessarily here

cabbby:
Did not notice that fluid filter can be placed on pipes and tanks. Now it should work.

Archengius:
You did it wrong again. Instead of special-casing FluidHandlerProxy, you should enumerate all tanks on all types of fluid handlers and try to drain their fluid

cabbby:
Edited

PR #913, opened 2019-07-14
Add OreByproduct page to JEI

Blood-Asp:
As Titel says. Image should be best to show result:

image

Additional ideas: 
- Make this page searchable by tiny/small dusts like in GT5/6.
- Add sifting to page.
- Add electrolyze/centrifuge to page for complete overview.


Archengius:
GUI looks quite cool, however it's not quite obvious from which processing step you get 4th component without hovering on it.
Also, if only purpose of these "fake" recipes is showing information generated automatically from ore type, you don't need to separate recipe implementation from IRecipeWrapper, and field in Recipe maps is not necessary at all. Overall, avoid using "fake" terminology.

HiggsBonbon:
I actually really like this, I often find myself getting lost in production chains with r/u trying to find what ores are most efficient for their primary products, and which I should save for later advanced processing. Make it an opt-in maybe?

PR #987, opened 2019-11-16
Pump, Rotor recipe tweaks

pyure:
* Pump now requires plates instead of tools
* Rotor now requires 1 tool instead of 3
* Reduced amount of soldering fluid required in assembler 
* Doubled pump extraction rate


Archengius:
Remove soldering fluid requirement completely, also remove screw from assembler recipe.

Agrax:
> 
> 
> Remove soldering fluid requirement completely, also remove screw from assembler recipe.

I agree on the screw. It's kind of obsolete. Assembler should make the recipe simpler, but i'd insist on the soldering material, as in the assembler it traditionally replaces all these screws and other fiddly bits and pieces. But hey! Im just a rando from the internet! Why would anyone listen to the humble me?

pyure:
PR changes done.

Agrax:
Looks bad :[

pyure:
Looks/feels good to me.

Agrax:
Well... One of us has bad taste

PR #1012, opened 2020-01-27
add metal_casing as autogenerate block

huneau:
Hello @Archengius,
I created a metal_casing autogenerate from Material and generate Texture.
In the future, I will need to have a quick metal_casing generator to create a lot of multiblock so I think it's the right place to add this. 
If you don't want it, let me know,  I will add it in my own project.

Edit: The idea is to create metal casing using new flag `GENERATE_METAL_CASING` and the important thing is that metal casing color is related to ingot color.
To create multiblock I need more casing and this feature can be reuse for anyone that wants to add new multiblock with the other colors than the 4 one we have :)
metal casing


Archengius:
do you really need separate texture for each material set?

src/main/java/gregtech/common/blocks/BlockMetalCasing.java
```diff
@@ -1,53 +1,83 @@
...
 import net.minecraft.entity.EntityLiving.SpawnPlacementType;
-import net.minecraft.util.IStringSerializable;
+import net.minecraft.util.EnumFacing;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.IBlockAccess;
+import net.minecraftforge.fml.relauncher.Side;
+import net.minecraftforge.fml.relauncher.SideOnly;
 
-public class BlockMetalCasing extends VariantBlock<BlockMetalCasing.MetalCasingType> {
+public final class BlockMetalCasing extends Block  implements ICubeRenderer, IIconRegister {
```
Archengius: It is straight wrong. Why do you need to implement ICubeRenderer from scratch? Why do you implement it within block glass?

src/main/java/gregtech/common/blocks/MetaBlocks.java
```diff
@@ -179,6 +178,11 @@ public static void init() {
                 blockFrame.setRegistryName("frame_" + material.toString());
                 FRAMES.put((SolidMaterial) material, blockFrame);
             }
+            if (material instanceof SolidMaterial && material.hasFlag(GENERATE_METAL_CASING)) {
```
Archengius: do you really need casings from gems?

src/main/java/gregtech/loaders/oreprocessing/MaterialRecipeHandler.java
```diff
@@ -362,6 +362,24 @@ public static void processFrame(OrePrefix framePrefix, SolidMaterial material) {
         }
     }
 
+    public static void processMetalCasing(OrePrefix prefix, SolidMaterial material) {
+        if (material.hasFlag(GENERATE_PLATE | GENERATE_FRAME | GENERATE_METAL_CASING)) {
```
Archengius: that's not how things work. reason of these if checks is protecting handling code from items registered by other mods, not preventing fucked up frames from gems from generating recipes.

src/main/resources/assets/gregtech/lang/en_us.lang
```diff
@@ -1702,17 +1703,6 @@ tile.boiler_casing.steel_firebox.name=Steel Firebox Casing
 tile.boiler_casing.titanium_firebox.name=Titanium Firebox Casing
 tile.boiler_casing.tungstensteel_firebox.name=Tungstensteel Firebox Casing
 
-# Metal casing blocks
```
Archengius: Do you understand that names here are just DIFFERENT from the names you implemented with autogenerated casings? Have you ever thought about retaining old names for currently registered casings?

src/main/resources/assets/gregtech/lang/ru_ru.lang
```diff
@@ -1712,6 +1701,8 @@ tile.machine_casing.steel_hull.name=Стальной корпус
 tile.machine_casing.steel_hull.tooltip=Для улучшения паровой машины
 tile.machine_casing.steel_bricks_hull.name=Кирпичный стальной корпус
 tile.machine_casing.steel_bricks_hull.tooltip=Для улучшения паровой машины
+tile.metal_casing.primitive_bricks.name=Примитивные кирпич
```
Archengius: Do you really know russian and chinese? Russian translation makes no sense, and i guess it's the same for chinese.

src/main/java/gregtech/api/unification/material/type/SolidMaterial.java
```diff
@@ -32,6 +32,7 @@
         public static final long GENERATE_LONG_ROD = createFlag(22);
         public static final long MORTAR_GRINDABLE = createFlag(24);
         public static final long GENERATE_FRAME = createFlag(45);
+        public static final long GENERATE_METAL_CASING = createFlag(65);
```
Archengius: that flag should include GENERATE_FRAME | GENERATE_PLATE automatically. See how it is implemented for other things like long sticks, which generate sticks when being added in solo.

src/main/java/gregtech/api/unification/material/MaterialIconType.java
```diff
@@ -85,6 +85,7 @@
     frameGt,
     frameSide,
     frameTop,
+    metalCasing,
```
Archengius: i'd advice using gtMetalCasing here, since such ore prefix can be used in other mods and will mean different thing.

src/main/java/gregtech/common/blocks/BlockMetalCasing.java
huneau: Yes, you right, I speed my day on that. The multiblock has a method called `public ICubeRenderer getBaseTexture(IMultiblockPart sourcePart)` And I'm not able to understand  how I can get an `ICubeRenderer ` from a `Block`( maybe I should use other than a `Block`). It's a copy of BlockFrame class because I need globally the same thing

src/main/java/gregtech/common/blocks/MetaBlocks.java
huneau: It's a copy from Frames, not a good Idea ?

src/main/resources/assets/gregtech/lang/en_us.lang
huneau: The original name doest exist anymore and it's autogenerated with the name of the metal

src/main/resources/assets/gregtech/lang/ru_ru.lang
huneau: I just move the translation It's not a creation

huneau:
> do you really need separate texture for each material set?

Actually, when I want a new metal_casing texture I spent time on paint to create a new one. It seems tedious for me but maybe is there a tool to make texture quickly than paint? 🤔

src/main/java/gregtech/api/unification/material/MaterialIconType.java
huneau: ok I change that

src/main/java/gregtech/loaders/oreprocessing/MaterialRecipeHandler.java
huneau: I'm not understanding your comment, what I'm supposed to do?

Agrax:
> 
> 
> > do you really need separate texture for each material set?
> 
> Actually, when I want a new metal_casing texture I spent time on paint to create a new one. It seems tedious for me but maybe is there a tool to make texture quickly than paint? thinking

I usually use paint in conjunction w/ Gimp or PhotoShop. Paint to make textures and PS/Gimp to add shading

src/main/resources/assets/gregtech/lang/en_us.lang
Archengius: And look worse than original ones, nobody asked you to get rid of them. Bring old names back for casings currently available in gtce.

src/main/java/gregtech/common/blocks/MetaBlocks.java
Archengius: No, bad idea. If you blindly copy stuff, are you sure you want to submit PRs?

huneau: it's changed to IngotMaterial

src/main/resources/assets/gregtech/lang/en_us.lang
huneau: name revert

src/main/java/gregtech/common/blocks/BlockMetalCasing.java
huneau: still that, If you can help me or give me a trace I can follow ...

Archengius:
Have you thought about backward compatibility in people saves btw? Currently, all casings will just disappear. This PR cannot be accepted in such condition.

PR #1026, opened 2020-02-11
Fix multiblock GUI on dedicated servers

Glease:
The old constructor reference will leave a client only class in the method signature in the resulting bytecode. This in turn leads to JVM attempts to load said client only class on every initialization of AdvancedTextWidget (and boom). Turning it into a lambda should solve the problem.


Glease:
This doesn't acutally work. Will send a valid one later.

PR #1028, opened 2020-02-11
Fix SideOnly related issues on dedicated servers

Glease:
The unchecked at ClientSideField is necessary, or you have to use unchecked at object instantiation site.

The two field in surface rock have type of Object which is safe. They have initializers so they will be referenced in constructor. Having SideOnly cause any constructor call to fail with NoSuchFieldError.

If you don't like my approach, feel free to close my PR and create your own solution.


Archengius:
These both are fixed locally already, will push changes tomorrow

Glease:
Thats fine then

Glease:
Just a kind reminder. Did you forget anything?

PR #1047, opened 2020-03-25
Fixed energy network ignoring sideness on connecting to machine

### LINKED ISSUE #1030: [Bug] Shutter Module not working properly
## Description
Shutter Module won't work properly in certain setup.

## To reproduce
1. Place these machines like the image below:
- HV Generator
- MV Machine (Input Hatch in the image)
- 4x MV Cables
- LV Machines with Shutter Module, preventing connection to cables
image1
2. Place HV Transformer (HV to MV).
3. Two of the LV Machines will explode.
image2

## Expected behavior
Won't explode due to Shutter Module.

## Version
gregtech-1.12.2-1.9.0.481.jar

LAGIdiot:
Energy network did not correctly provide information from which side is energy being pushed to machine, which prevented cover capability checks which in the end prevented Shutter Cover to work correctly.

Fixes #1030 

Tested all different sets of block accepting energy and all covers.

### DIFF
--- a/src/main/java/gregtech/common/pipelike/cable/tile/CableEnergyContainer.java
+++ b/src/main/java/gregtech/common/pipelike/cable/tile/CableEnergyContainer.java
@@ -85,7 +85,7 @@ public class CableEnergyContainer implements IEnergyContainer {
             if (tileEntity == null || tileEntityCable.getPipeBlock().getPipeTileEntity(tileEntity) != null) {
                 continue; //do not emit into other cable tile entities
             }
-            IEnergyContainer energyContainer = tileEntity.getCapability(GregtechCapabilities.CAPABILITY_ENERGY_CONTAINER, null);
+            IEnergyContainer energyContainer = tileEntity.getCapability(GregtechCapabilities.CAPABILITY_ENERGY_CONTAINER, facing.getOpposite());
             if (energyContainer == null) continue;
             amperesUsed += energyContainer.acceptEnergyFromNetwork(facing.getOpposite(), voltage, amperage - amperesUsed);
             if (amperesUsed == amperage)


src/main/java/gregtech/common/pipelike/cable/tile/CableEnergyContainer.java
```diff
@@ -85,7 +85,7 @@ private long dispatchEnergyToNode(BlockPos nodePos, int nodeBlockedConnections,
             if (tileEntity == null || tileEntityCable.getPipeBlock().getPipeTileEntity(tileEntity) != null) {
                 continue; //do not emit into other cable tile entities
             }
-            IEnergyContainer energyContainer = tileEntity.getCapability(GregtechCapabilities.CAPABILITY_ENERGY_CONTAINER, null);
+            IEnergyContainer energyContainer = tileEntity.getCapability(GregtechCapabilities.CAPABILITY_ENERGY_CONTAINER, facing.getOpposite());
```
pyure: LAGIdiot, can you explain to me what this does?  (What was wrong with null?)

LAGIdiot: Sure, when getCapability does not know (null supplied) which side of tile entity it should check it will just ignore all covers capabilities on it.

pyure: Ultimately what was the result to the end user?

LAGIdiot: Result was energy network did not check that it was connected to machine on Shutter Cover which should prevent all connection. And it blowed machine up as MV cable was connected to LV machine.

PR #1070, opened 2020-04-25
Improvement/update dependencies to last version

LAGIdiot:
All dependencies were updated to latest/last official version available
(last is really meant as LAST because most of them did not get update in years)

Forge updated to 14.23.5.2847

CraftTweaker updated to 4.1.9.6
CodeChickenLib updated to 3.2.3.358
Forestry updated to 5.8.2.387
ForgeMultipart updated to 2.6.2.83
JEI updated to 4.15.0.291
The One Probe updated to 1.4.28

Bump version

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -1,10 +1,10 @@
 minecraft.version=1.12.2
-forge.version=14.23.5.2838
-ccl.version=3.2.2.355
+forge.version=14.23.5.2847
+ccl.version=3.2.3.358
 chickenasm.version=1.0.2.9
-forestry.version=5.8.0.256
-jei.version=4.9.1.188
-multipart.version=2.4.2.58
+forestry.version=5.8.2.387
+jei.version=4.15.0.291
+multipart.version=2.6.2.83
 baubles.version=1.5.2
-crafttweaker.version=1.12-4.1.8.9
-top.version=1.4.23-16
+crafttweaker.version=1.12-4.1.9.6
+top.version=1.4.28-17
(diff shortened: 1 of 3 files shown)


huneau:
The One Probe updated to 1.4.28
this version seems buggy. 
on my side, I have issue with `GregtechTileCapabilities.CAPABILITY_CONTROLLABLE.cast(this);` (or other capabilities) when TOP ask info

LAGIdiot:
I have retest it again and I don't see any problems on GTCE side. There are no errors in logs, no exceptions what so ever. And all info provides for TOP works correctly.

Are you sure you are using it correctly in your code?

huneau:
unknown
The issue is probably not TOP but I see a lot of issue like that... and it's not particularly mine

LAGIdiot:
It seems that issue must be in how _topaddons:ic2_ or _gtadditions:screwdriver_provider_ works, maybe both or block you are looking at. But either way this is **not** GTCE problem. This version of TOP exists for more then year so it is kind of theirs problem that they did not adapt yet.

huneau:
I made some testing and the issue is not in TOP... its somewhere near `GregtechTileCapabilities.CAPABILITY_CONTROLLABLE.cast(this);` 
my side, I dont use that because it doesn't work I do `(T) this`
exactly the same but .......................

huneau:
the codechicken lib is a real pain, the repo doesnt work I have to add another one

huneau:
@pyure I tested the branch of @LAGIdiot I dont see issue, seems fine for me

PR #1072, opened 2020-04-29
Fix typo in Crafttweaker API

Ludi87:


Archengius:
breaks compatibility with scripts already using it. create new method with fixed name.

Issue #1068, opened 2020-04-22
JEI integer overflow with recipe


hjae78:
This occurs when the total required EU of a recipe exceeds int max. It can be fixed quite easily [here](https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java) by simply casting the result of the multiplication to long.

huneau:
![image](https://user-images.githubusercontent.com/9199240/80216693-bb7f5500-863e-11ea-81be-dec63cae790d.png)

LAGIdiot:
Fixed in #1076.

PR #1076, opened 2020-05-02
Add flexibility into API

huneau:
remove taking damage when cable loss is == 0
add a possibility to disable direct smelting
add a possibility to change crushed material
SimpleGeneratorMetaTileEntity has now a createWorkableHandle
INSULATION_MATERIALS is now public to add more insulation tier
fix getTierByVoltage stop throw ArrayOutOfBoundException
Total EU display is a Long

### DIFF
--- a/src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
+++ b/src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
@@ -108,7 +108,7 @@ public class GTRecipeWrapper implements IRecipeWrapper {
 
     @Override
     public void drawInfo(Minecraft minecraft, int recipeWidth, int recipeHeight, int mouseX, int mouseY) {
-        minecraft.fontRenderer.drawString(I18n.format("gregtech.recipe.total", Math.abs(recipe.getEUt()) * recipe.getDuration()), 0, 70, 0x111111);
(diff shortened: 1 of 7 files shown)


src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
```diff
@@ -108,7 +108,7 @@ public void addTooltip(int slotIndex, boolean input, Object ingredient, List<Str
 
     @Override
     public void drawInfo(Minecraft minecraft, int recipeWidth, int recipeHeight, int mouseX, int mouseY) {
-        minecraft.fontRenderer.drawString(I18n.format("gregtech.recipe.total", Math.abs(recipe.getEUt()) * recipe.getDuration()), 0, 70, 0x111111);
+        minecraft.fontRenderer.drawString(I18n.format("gregtech.recipe.total", Math.abs((long) recipe.getEUt()) * recipe.getDuration()), 0, 70, 0x111111);
```
pyure, line 111: What is this casting accomplishing?  Precision?

huneau: we have 1 integer multiply by 1 integer the result is an Integer, so 2^31 max and when I make UV recipe I reach that point. 
By casing the function into long I can multiply 1 Long to 1 Integer the result can be 2^63.

hjae78: See [this](https://github.com/GregTechCE/GregTech/issues/1068) issue.

PR #1090, opened 2020-05-29
Add pan and zoom for Multiblock Pattern category in JEI.

eutropius225:
These can be done with right click and shift right click respectively.

This would be particularly useful when using addon mods featuring large multiblocks, whose previews are otherwise mostly uninformative.


huneau:
Not directly related to that issue, but, should have a tooltip somewhere that explains how to use the Multiblock pattern viewer. (I discovers last that shift click can move up and down :O)

Issue #1092, opened 2020-06-01
[GTCE 1.9.1.505] Surface Rock error


whatsapp2:
Mod Version: 1.9.1.505
Forge version: 14.23.5.2854
Allocated Memory: 12 GB
List of mods: https://pastebin.com/QgPeD5du
Crash report: https://pastebin.com/RQDbwc4b
will this bug be fixed in the next version?

LAGIdiot:
Known workaround: Downgrade Forge to 14.23.5.2847
Fixing it is on list with high priority but it is quite complex problem so I can't guarantee it will be fixed soon.

Issue #1088, opened 2020-05-28
Invalid placement of Surface Rock crashes game


LAGIdiot:
[crash-2020-05-27_17.43.38-server.txt](https://github.com/GregTechCE/GregTech/files/4689932/crash-2020-05-27_17.43.38-server.txt)
Gonna add my crashreport on this as well. 
GregTechCE Version: 1.9.1.505
Forge Version: 14.23.5.2847

_Originally posted by @Freebi85 in https://github.com/GregTechCE/GregTech/issues/1086#issuecomment-634756906_

Issue #1086, opened 2020-05-25
Error: failed to load data for block entity gregtech:surface_rock


DMTfarc:
This error constantly spamming in the console every minute.
Server: Spongeforge 7.2.1
Plugins: FastAsyncWorldEdit(19.11.13), LuckPerms(5.0.39)
Mods: Baubles(152), Botania(110-363), CodeChicken Lib(3.2.3.358), **GregTech Community Edition(1.9.1.505)**, MrCrayfish's Furniture Mod(631), Pam's HarvestCraft(1122zg), Thaumcraft(6.1b26), The Twilight Forest(3.10.1013)
`[12:38:18] [Server thread/ERROR] [FML]: A TileEntity gregtech:surface_rock(gregtech.common.blocks.surfacerock.TileEntitySurfaceRock) has thrown an exception during loading, its state cannot be restored. Report this to the mod author
java.lang.IllegalAccessError: tried to access field net.minecraft.nbt.NBTTagList.field_74747_a from class gregtech.common.blocks.surfacerock.TileEntitySurfaceRock
        at gregtech.common.blocks.surfacerock.TileEntitySurfaceRock.func_145839_a(TileEntitySurfaceRock.java:68) ~[TileEntitySurfaceRock.class:?]
        at net.minecraft.tileentity.TileEntity.func_190200_a(TileEntity.java:120) [avj.class:?]
        at net.minecraft.world.chunk.storage.AnvilChunkLoader.loadEntities(AnvilChunkLoader.java:495) [aye.class:?]
        at net.minecraftforge.common.chunkio.ChunkIOProvider.syncCallback(ChunkIOProvider.java:101) [ChunkIOProvider.class:?]
        at net.minecraftforge.common.chunkio.ChunkIOExecutor.tick(ChunkIOExecutor.java:150) [ChunkIOExecutor.class:?]
        at net.minecraft.server.MinecraftServer.func_71190_q(MinecraftServer.java:728) [MinecraftServer.class:?]
        at net.minecraft.server.dedicated.DedicatedServer.func_71190_q(DedicatedServer.java:397) [nz.class:?]
        at net.minecraft.server.MinecraftServer.func_71217_p(MinecraftServer.java:668) [MinecraftServer.class:?]
        at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:526) [MinecraftServer.class:?]
        at java.lang.Thread.run(Unknown Source) [?:1.8.0_241]
[...]

PR #1093, opened 2020-06-03
Bugfix/surface rocks

LAGIdiot:
**Problem:**
1) Surface rocks crashing game when placed on invalid terrain (terrain got changed after validating if surface rock can be placed).
2) NBT data loading for surface rocks not working correctly on latest Forge (version 14.23.5.2854). Cluttering log and resetting surface rocks to aluminium.

**Why:**
1) Missing check if surface rock was placed correctly before accessing it. 
2) NBT data list which GTCE is working with in this case is going to be private.

**How solved:**
1) Added missing check.
2) Used iterator instead of accessing list.

**Outcome:**
Fixes: #1092 
Fixes: #1088 
Fixes: #1086

### DIFF
--- a/src/main/java/gregtech/api/worldgen/populator/SurfaceRockPopulator.java
+++ b/src/main/java/gregtech/api/worldgen/populator/SurfaceRockPopulator.java
@@ -2,7 +2,6 @@ package gregtech.api.worldgen.populator;
 
 import com.google.gson.JsonObject;
 import gregtech.api.unification.OreDictUnifier;
-import gregtech.api.unification.material.type.DustMaterial;
 import gregtech.api.unification.material.type.Material;
 import gregtech.api.unification.stack.UnificationEntry;
 import gregtech.api.worldgen.config.OreConfigUtils;
@@ -38,10 +37,7 @@ public class SurfaceRockPopulator implements VeinChunkPopulator {
 
     @Override
     public void loadFromConfig(JsonObject object) {
-        DustMaterial material = OreConfigUtils.getMaterialByName(object.get("material").getAsString());
-        if (!(material instanceof DustMaterial))
-            throw new IllegalArgumentException("Only dust materials are supported for surface rocks");
-        this.material = material;
+        this.material = OreConfigUtils.getMaterialByName(object.get("material").getAsString());
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/gregtech/api/worldgen/populator/SurfaceRockPopulator.java
```diff
@@ -38,10 +37,7 @@ public SurfaceRockPopulator(Material material) {
 
     @Override
     public void loadFromConfig(JsonObject object) {
-        DustMaterial material = OreConfigUtils.getMaterialByName(object.get("material").getAsString());
-        if (!(material instanceof DustMaterial))
-            throw new IllegalArgumentException("Only dust materials are supported for surface rocks");
-        this.material = material;
+        this.material = OreConfigUtils.getMaterialByName(object.get("material").getAsString());
```
pyure, line 40: Should we sanity-check object.get() and/or getAsString() prior to passing into getMaterialByName?

src/main/java/gregtech/api/worldgen/populator/SurfaceRockPopulator.java
```diff
@@ -68,16 +64,23 @@ public void initializeForVein(OreDepositDefinition definition) {
...
     private void setStoneBlock(World world, BlockPos blockPos, Collection<Material> undergroundMaterials) {
-        world.setBlockState(blockPos, MetaBlocks.SURFACE_ROCK_NEW.getDefaultState());
-        TileEntitySurfaceRock tileEntity = (TileEntitySurfaceRock) world.getTileEntity(blockPos);
-        tileEntity.setData(material, undergroundMaterials);
+        boolean surfaceRockPlaced = world.setBlockState(blockPos, MetaBlocks.SURFACE_ROCK_NEW.getDefaultState());
+        if(surfaceRockPlaced)
+        {
+            TileEntitySurfaceRock tileEntity = (TileEntitySurfaceRock) world.getTileEntity(blockPos);
+            if(tileEntity != null)
+                tileEntity.setData(material, undergroundMaterials);
```
pyure, line 72: For clarity can we refer to the class members with this.material?

src/main/java/gregtech/api/worldgen/populator/SurfaceRockPopulator.java
pyure: Fair, thanks.

pyure:
Which forge versions did you test against?

LAGIdiot:
> Which forge versions did you test against?

I tested against Forge versions: 14.23.5.2847 and 14.23.5.2854 I also tested for both with and without vanillafix. So we should be in clear.

PR #1104, opened 2020-06-10
Restore MV macerator byproduct slot

ALongStringOfNumbers:
This PR restores the byproduct slot for the MV macerator. In https://github.com/GregTechCE/GregTech/commit/b21b6ae036cb22ee6b3d45bf473b8ec518e10417 which is the commit that removed the MV macerator byproduct slot, The byproduct chance was also set to start at the MV tier, which is a little useless if you cannot get any byproducts at this tier. Therefore, this PR readjusts the slots so that you can get byproducts at this tier.


Shattered2909:
No... MV unlocks the Thermal Centrifuge, HV the Macerator byproduct... it makes sense and does not need changing at all.

ALongStringOfNumbers:
HV unlocks two byproduct slots, not just one. Before there was a build up of byproduct slots from 0 at LV to 1 at MV to 2 at HV. Restoring the 1 slot to MV also allows for byproducts not available through the Thermal Centrifuge.

Shattered2909:
The third slot is an absolute joke that might give Stone Dusts and thats it. The HV Macerator gate exists for a reason.

ALongStringOfNumbers:
Even if what you are saying is correct, and the Thermal Centrifuge is the intended route, then it does not make sense for the byproduct chance to start at the MV tier, as the byproduct chance is not needed for the Thermal Centrifuge, and not available for the Macerator at the MV tier.

Shattered2909:
What. That is just the way its calculated. feel free to change it with Crafttweaker.

ALongStringOfNumbers:
This would be beyond tedious to change with crafttweaker for every ore available in GTCE.

Shattered2909:
Pro tip: keep it as it is. GTCE does not need to be made easier than it is.

ALongStringOfNumbers:
That is fine as your personal opinion, but is not the subject of this pull request.

pyure:
Good conversation that should probably go to the discord for more input.

LAGIdiot:
Sign me up for conversation regarding this

Agrax:
@ALongStringOfNumbers have this been discussed w/ either pyure or LAGidiot *BEFORE* PRing a change? Usually all modifications to the mod should be discussed first in order to not be dissapointed when a maintainer rejects the PR because it doesn't sound good to him.

Either Shattered's input or mine don't mean much when it comes to the final decision, but one of the maintainers should always be informed in advance.

eutropius225:
When the change is a single character there's really not much lost in starting discussion through a pull request.

huneau:
Maybe the good idea is config file ?

Saereth:
config options file would be preferable here. The problem is, just as you want the slot available at MV, many packs are already designed around the principle of it not being available. And just as you said, changing every recipe with crafttweaker would be just as tedious for those packs as well if it were changed at this point. If this were a discussion for initial design I'd agree with you, but it's a little late in the game to make a blanket change like this imo. A config option, in theory could satisfy both side of the debate.

ALongStringOfNumbers:
Will open a new PR with the config option. My lack of familiarity with IntelliJ's git interface messed this one up.

PR #1108, opened 2020-06-21
Add minimum tier required to JEI information.

clienthax:
Required to prevent me from going insane while playing omnifactory.

image

### DIFF
--- a/src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
+++ b/src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
@@ -1,12 +1,16 @@
 package gregtech.integration.jei.recipe;
 
 import codechicken.lib.util.ItemNBTUtils;
+import gregtech.api.GTValues;
 import gregtech.api.recipes.CountableIngredient;
 import gregtech.api.recipes.Recipe;
 import gregtech.api.recipes.Recipe.ChanceEntry;
 import gregtech.api.recipes.RecipeMap;
 import gregtech.api.unification.OreDictUnifier;
+import gregtech.api.util.GTUtility;
+import gregtech.integration.jei.utils.JEIHelpers;
 import mezz.jei.api.ingredients.IIngredients;
+import mezz.jei.api.ingredients.VanillaTypes;
 import mezz.jei.api.recipe.IRecipeWrapper;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.resources.I18n;
@@ -48,7 +52,7 @@ public class GTRecipeWrapper implements IRecipeWrapper {
                 });
                 matchingInputs.add(ingredientValues);
             }
-            ingredients.setInputLists(ItemStack.class, matchingInputs);
+            ingredients.setInputLists(VanillaTypes.ITEM, matchingInputs);
         }
 
         if (!recipe.getFluidInputs().isEmpty()) {
@@ -63,7 +67,7 @@ public class GTRecipeWrapper implements IRecipeWrapper {
                     stack.amount = 1;
                 }
             });
-            ingredients.setInputs(FluidStack.class, recipeInputs);
+            ingredients.setInputs(VanillaTypes.FLUID, recipeInputs);
         }
 
         if (!recipe.getOutputs().isEmpty() || !recipe.getChancedOutputs().isEmpty()) {
@@ -77,13 +81,13 @@ public class GTRecipeWrapper implements IRecipeWrapper {
                 recipeOutputs.add(chancedStack);
             }
             recipeOutputs.sort(Comparator.comparing(stack -> ItemNBTUtils.getInteger(stack, "chance")));
-            ingredients.setOutputs(ItemStack.class, recipeOutputs);
+            ingredients.setOutputs(VanillaTypes.ITEM, recipeOutputs);
         }
 
         if (!recipe.getFluidOutputs().isEmpty()) {
             List<FluidStack> recipeOutputs = recipe.getFluidOutputs()
                 .stream().map(FluidStack::copy).collect(Collectors.toList());
-            ingredients.setOutputs(FluidStack.class, recipeOutputs);
+            ingredients.setOutputs(VanillaTypes.FLUID, recipeOutputs);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
```diff
@@ -111,11 +111,37 @@ public void drawInfo(Minecraft minecraft, int recipeWidth, int recipeHeight, int
...
+
+        int baseYPosition = 110;
         for (String propertyKey : recipe.getPropertyKeys()) {
             minecraft.fontRenderer.drawString(I18n.format("gregtech.recipe." + propertyKey,
                 recipe.<Object>getProperty(propertyKey)), 0, baseYPosition, 0x111111);
             baseYPosition += 10;
         }
     }
+
+    private String getMinTierForVoltage(long voltage) {
```
LAGIdiot, line 124: This function is redundant and wrong (using magic numbers and hardcoded strings is not good practice). Please use functionality from https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/api/util/GTUtility.java#L380 and https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/api/GTValues.java#L58 instead to get correct tier name.

src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
```diff
@@ -111,11 +111,37 @@ public void drawInfo(Minecraft minecraft, int recipeWidth, int recipeHeight, int
         minecraft.fontRenderer.drawString(I18n.format("gregtech.recipe.total", Math.abs((long) recipe.getEUt()) * recipe.getDuration()), 0, 70, 0x111111);
         minecraft.fontRenderer.drawString(I18n.format(recipe.getEUt() >= 0 ? "gregtech.recipe.eu" : "gregtech.recipe.eu_inverted", Math.abs(recipe.getEUt())), 0, 80, 0x111111);
         minecraft.fontRenderer.drawString(I18n.format("gregtech.recipe.duration", recipe.getDuration() / 20f), 0, 90, 0x111111);
-        int baseYPosition = 100;
+        minecraft.fontRenderer.drawString(I18n.format("gregtech.recipe.min_tier", getMinTierForVoltage(recipe.getEUt())), 0, 100, 0x111111);
```
LAGIdiot, line 114: Did you test it on all machines, are you sure that you are not clipping in any of them? Does this cover even Multiblocks?

src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
clienthax: It looks like GTUtility.getTierByVoltage is returning values off by 1, when referenced against VN
eg GTValues.VN[GTUtility.getTierByVoltage(recipe.getEUt())] for "Nicken Zinc Ferrite Dust" at 60 EU/t should be MV, but this returns LV.
Is this to be expected?

src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
clienthax: I am probably not familiar enough with gregtech to test this properly on anything with special logic, but it seems to be correct for the electrolyzer.

src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
clienthax: https://gist.github.com/clienthax/594dda020925b92e61742a80c70a6912

huneau:
Have you test on blast furnace ? Maybe overlap with heat required.
Maybe put it in parentheses after eu/t

src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
LAGIdiot: Interesting I did not expect that. It seems that function I recommend is not suited for this job. But that does not make much difference. Please create similar function in integration/jei/utils which will correctly calculate minimal tier using constants from GTValues.V and GTValues.VN and use that instead.

src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
LAGIdiot: Okay then I will specify it more. Please check these machines that they dos not show any misleading values or it does not cover machine GUI or is covered by it or anything that looks bad: Steam Turbine (representative of energy generators), Assembly machine (problematic GUI), Electric Blast Furnace (multiblock with more lines of informations in GUI).

clienthax: It looks like the assembler already clipped the existing text
image

This line doesn't show at all on steam turbines / diesel generator as they use a different wrapper.

It does however cause clipping on the blast furnace
image

src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
clienthax: Yeah that function in GTUtility is odd, I would say its either badly named or just not returning what its meant to.

src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
clienthax: @LAGIdiot Any ideas on how to go about sorting this?

huneau:
I repost my suggestion 
image

clienthax:
> I repost my suggestion
> image

Your suggestion doesn't address the other issues with overlapping.

src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
LAGIdiot: I was going to look at and properly test it tomorrow as I may be able to get some time for it.

But if you need opinion regarding this right now I can tell you strait of bat that Blast Furnace GUI must be fixed as it does not look good and it may interfere with recipes underneath. Also don't worry about Assembler fluid slot as it is not part of scope, but I can't see on provided picture that added text it is not implemented for that machine? 

And I think that min tier line should be moved under usage maybe even before. @pyure what is your take on that?

And last thing, please change as minimal parts of code as possible because we need to retest everything that is changed.

huneau:
> > I repost my suggestion
> > image
> 
> Your suggestion doesn't address the other issues with overlapping.

for this case 
image
its a substantive problem, the best things you can do is put a ducktape on it
look at here 
image

LAGIdiot:
Okay I test this PR and everything works as expected, no problems found.
So last thing to handle is where and how exactly to show that tier. Even @huneau suggestion seems like good idea.

Let's leave this open for discussion and I will make call based on feedback (if any) at Monday.

LAGIdiot:
Okay as no one had any more inputs on this. Let's role with @huneau suggestion (putting tier to parenthesis after EU/t) as it is most consistent with rest of JEI information and it save as pain regarding extending Blast Furnace UI.

clienthax:
@LAGIdiot Should be fine now
image

LAGIdiot:
Thanks for your hard work and I hope you will contribute again in future.
I have few pointers for you so you can do even better next time:
- Touch only code that you need (it will save time in testing)
- Use tools for automatic code formatting (it will be more readable)
- Don't use force push (it is more difficult to read changes and some automatic tools does not work properly with it)

Issue #1112, opened 2020-06-27
Add a Dimension Whitelist to Air Collectors in the Config


Shattered2909:
Just to make sure the Collectors do not work in space or other Dimensions.
Might wanna add a Tooltip to them as well, they currently have none.

PR #1113, opened 2020-06-27
Feature/air collector qol improvements

LAGIdiot:
**Problem:**
1) Air Collector works even in dimensions where is no air
2) Air Collector does not have any tooltip in JEI

**How solved:**
1) There is config entry _airCollectorDimensionBlacklist_ which allows users to add dimensions in which they don't want Air Collector to function. I choose to go with blacklist as it is more probable that added dimensions contains air so most of time no action from user will be required. Also for all current users nothing will change and all Air Collectors will work as expected.
2) Added standard set of tooltips regarding voltage in, energy capacity and fluid capacity. Plus description of what machines does and Air collection speed tooltip.

**Outcome:**
Fixes: #1112


Shattered2909:
Seems good :D

Issue #1077, opened 2020-05-05
Assembler recipes requiring liquids cover total EU 


Zeruel13:
Some recipes e.x. assembler recipes that require liquid in addition to items will cover the total EU required when viewing the recipe

![image](https://user-images.githubusercontent.com/63512252/81096751-e2f2df00-8ed4-11ea-80c8-7dbaecf53b0a.png)

ALongStringOfNumbers:
I am seeing similar on version 1.9.1.505. I went through and checked on all the single block machines, and the assembler was the only place where I noticed this taking place, due to the layout of the liquid slots in relation to the position of the item slots.

huneau:
Not an easy fix due to the auto-generated layout. This also cause me issue :/

LAGIdiot:
Closed by: #1118

PR #1118, opened 2020-06-30
Render JEI recipe info from the bottom instead of at a constant height.

eutropius225:
This would match [my override](https://github.com/eutropius225/MultiblockTweaker/blob/684cce935afc3c569ea47c3c1c2d456a3d86e141/src/main/java/eutros/multiblocktweaker/jei/CustomRecipeWrapper.java#L20-L33), which was necessary to support categories with arbitrary height, or with an arbitrary number of properties.

For most categories, this results in the recipe info being rendererd 10 units lower.

### DIFF
--- a/src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
+++ b/src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
@@ -108,14 +108,18 @@ public class GTRecipeWrapper implements IRecipeWrapper {
 
     @Override
     public void drawInfo(Minecraft minecraft, int recipeWidth, int recipeHeight, int mouseX, int mouseY) {
-        minecraft.fontRenderer.drawString(I18n.format("gregtech.recipe.total", Math.abs((long) recipe.getEUt()) * recipe.getDuration()), 0, 70, 0x111111);
-        minecraft.fontRenderer.drawString(I18n.format(recipe.getEUt() >= 0 ? "gregtech.recipe.eu" : "gregtech.recipe.eu_inverted", Math.abs(recipe.getEUt())), 0, 80, 0x111111);
-        minecraft.fontRenderer.drawString(I18n.format("gregtech.recipe.duration", recipe.getDuration() / 20f), 0, 90, 0x111111);
-        int baseYPosition = 100;
+        int yPosition = recipeHeight - getPropertyListHeight();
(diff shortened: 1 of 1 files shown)


LAGIdiot:
At first I was like what a great idea but when I thought about it little more I am not so sure any more. 

This would bring inconsistency in positioning key lines. Point is lines regarding **duration** and **energy consumption** may change position based on of how much tooltip is under them. Which is not UX friendly. Also in situation when tooltip spans across many lines duration and energy consumption may be rolled in recipe input grid making it unreadable. Which in current setup these lines will be always readable (with exception of broken UI on assembler) and other info may be lost (but there is way much space down here).

What you think @pyure ?

eutropius225:
I haven't seen any recipe maps that allow for a varying number of properties, though if you can show one, it may be possible to find a solution.

Here are some screenshots thet @pyure requested.

Old:
image
New (may even fix #1077?):
image

Old:
image
New:
image

Old:
image
New:
image

pyure:
My thoughts:
* Running from bottom is a UX improvement overall.  The pieces bouncing around isn't ideal (as LAGIdiot indicates), but its vastly superior to overlapping elements, which is criminal.
* Pseudo-related: we should consider clawing back some space by combining the Usage and Duration lines onto a single line.  That level of verbosity doesn't help the user much.

LAGIdiot:
As there is **not much** recipes which would use additional tooltip the _bouncing_ is not that problematic as I pictured. So we may proceed with this one.

Regarding #1077 it look like it will be fixed by this so I will link it in. 

I don't think that **Usage** and **Duration** should be on same line as Usage is already quite long on higher tiers.

And last note thank you for providing pictures but please use pictures only containing our textures and recipes (no texture packs and recipes from other mods as they may be confusing).

src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
```diff
@@ -112,15 +112,18 @@ public void addTooltip(int slotIndex, boolean input, Object ingredient, List<Str
...
         for (String propertyKey : recipe.getPropertyKeys()) {
             minecraft.fontRenderer.drawString(I18n.format("gregtech.recipe." + propertyKey,
-                recipe.<Object>getProperty(propertyKey)), 0, baseYPosition, 0x111111);
-            baseYPosition += 10;
+                recipe.<Object>getProperty(propertyKey)), 0, yPosition += 10, 0x111111);
         }
     }
+
+    private int getPropertyListHeight() {
+        return (recipe.getPropertyKeys().size() + 3) * 10;
```
LAGIdiot, line 126: Please move that magic number **10** to constant like _lineHeight_ and use it instead. This applies for whole file.

PR #1125, opened 2020-07-03
QoL for pump cover

huneau:
A bit of QoL for the pump cover, I need to avoid like 10000000000 click to select the flow rate.
Also added consistency on some button.
Render:
image
image

I also added consistency on the button
image

a bit weird, I perfectly center the text but it's not centered.
image

the `SimpleTextWidget` should be changed to avoid that default behavior to automatically center the text but its not the purpose of the PR
image

### DIFF
--- a/src/main/java/gregtech/common/covers/CoverPump.java
+++ b/src/main/java/gregtech/common/covers/CoverPump.java
@@ -81,6 +81,8 @@ public class CoverPump extends CoverBehavior implements CoverWithUI, ITickable, 
 
     public void setBucketMode(BucketMode bucketMode) {
         this.bucketMode = bucketMode;
+        if (this.bucketMode == BucketMode.BUCKET)
+            setTransferRate(transferRate / 1000 * 1000);
         coverHolder.markDirty();
     }
 
@@ -244,7 +246,6 @@ public class CoverPump extends CoverBehavior implements CoverWithUI, ITickable, 
         super.writeToNBT(tagCompound);
         tagCompound.setInteger("TransferRate", transferRate);
         tagCompound.setInteger("PumpMode", pumpMode.ordinal());
-        tagCompound.setInteger("BucketMode", bucketMode.ordinal());
         tagCompound.setBoolean("WorkingAllowed", isWorkingAllowed);
         tagCompound.setBoolean("AllowManualIO", allowManualImportExport);
         tagCompound.setTag("Filter", fluidFilter.serializeNBT());
@@ -255,9 +256,8 @@ public class CoverPump extends CoverBehavior implements CoverWithUI, ITickable, 
         super.readFromNBT(tagCompound);
         this.transferRate = tagCompound.getInteger("TransferRate");
         this.pumpMode = PumpMode.values()[tagCompound.getInteger("PumpMode")];
-        this.bucketMode = BucketMode.values()[tagCompound.getInteger("BucketMode")];
         //LEGACY SAVE FORMAT SUPPORT
-        if(tagCompound.hasKey("FluidFilter")) {
+        if (tagCompound.hasKey("FluidFilter")) {
             this.fluidFilter.deserializeNBT(tagCompound);
         } else {
             this.fluidFilter.deserializeNBT(tagCompound.getCompoundTag("Filter"));
--- a/src/main/java/gregtech/common/covers/CoverPump.java
+++ b/src/main/java/gregtech/common/covers/CoverPump.java
@@ -49,6 +49,7 @@ public class CoverPump extends CoverBehavior implements CoverWithUI, ITickable, 
     private CoverableFluidHandlerWrapper fluidHandlerWrapper;
     protected boolean isWorkingAllowed = true;
     protected final FluidFilterContainer fluidFilter;
+    protected BucketMode bucketMode;
 
     public CoverPump(ICoverable coverHolder, EnumFacing attachedSide, int tier, int mbPerTick) {
         super(coverHolder, attachedSide);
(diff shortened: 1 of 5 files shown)


LAGIdiot:
This improvement looks quite useful it should definitely save quite few clicks on higher tier, but does it save enough of them? 

Mentioned consistency on buttons is in regard to what exactly?

huneau:
> Mentioned consistency on buttons is in regard to what exactly?

I make the button 18*18 instead of 20*20

LAGIdiot:
Regarding first part of my comment. After this PR you can change throughput by 500mb on click which seems nice. But on **UV** pump you would still need to do **41944 clicks** and even on **IV** pump you would need **164** clicks which is quite a lot to go all the way. How about making border buttons go like 1% or (5% on shift) to do heavy lifting and inner buttons could be used for fine tuning? That may make it look silly on lower tiers (which can be mitigated by not showing these buttons) but could help on higher tiers. What is you take on this idea?

Regarding second part of my comment I am interested in why did you change them. Is it because all other covers has buttons 18x18? Or because all GTCE buttons are 18x18? Or ???

huneau:
This button can be switch to mB/s to B/s (and the +1 +10 and +100 take the /s config)
image

And yes, I noticed that all interface is base 18 not 20

LAGIdiot:
This GUI is not intuitive for me at all right now. As it still works in background with mB/s even when GUI is showing me B/s.

Lets imagine situation that I want it to pump 15 B/s. I place Electric Pump (HV) and open it's GUI. On first open it shows max throughput which is 20480 mB/s (default behavior). I click mB/s button because putting it in buckets is going to be quicker. Which changes number in GUI to 20 B/s then I click -1 with Shift. Now GUI is showing 15 B/s and I exit it.

Question is how much will it pumped and why? Please answer without looking at code or testing it.

huneau:
it will set up to 15482mb/t

LAGIdiot:
> it will pump 15480mb/s and I think you ask me why because you know the answer.

Yeah I know the answer. But will it be obvious to everyone? That we are calculating with mB/s on background and showing only part of it in GUI. I feel like this could bring lot's of confusion to players.

DeckerCHAN:
@huneau In case #1135 been merged prior to this, I will fix your PR.

DeckerCHAN:
BTW, my personal opinion is that if we have pump with 10000 mb/t capacity, do we really care about the difference between 100000mb/t and 100010mb/t? Therefore, let people adjust with ±10% and ±1% percent of maximum capacity should be good enough.  @pyure

pyure:
Logically I don't really disagree.  Two concerns:
1) someone MIGHT care for some weird modpack reason we're not considering
2) Adjusting by percent just FEELS weird.

src/main/java/gregtech/common/covers/CoverPump.java
```diff
@@ -66,14 +68,24 @@ protected void setTransferRate(int transferRate) {
     }
 
     protected void adjustTransferRate(int amount) {
-        setTransferRate(MathHelper.clamp(transferRate + amount, 1, maxFluidTransferRate));
+        if (this.bucketMode == BucketMode.BUCKET)
+            setTransferRate(MathHelper.clamp(transferRate + amount * 1000, 1, maxFluidTransferRate));
+        else if (this.bucketMode == BucketMode.MILLI_BUCKET)
+            setTransferRate(MathHelper.clamp(transferRate + amount, 1, maxFluidTransferRate));
```
LAGIdiot, line 74: I think we should use this one as it is clear enough but does not have that much boiler plate on it
```suggestion
        amount *= this.bucketMode == BucketMode.BUCKET ? 1000 : 1;
        setTransferRate(MathHelper.clamp(transferRate + amount, 1, maxFluidTransferRate));
```

src/main/java/gregtech/common/covers/filter/FluidFilterContainer.java
```diff
@@ -54,7 +54,7 @@ public void initUI(int y, Consumer<Widget> widgetGroup) {
         widgetGroup.accept(new LabelWidget(10, y, "cover.pump.fluid_filter.title"));
         widgetGroup.accept(new SlotWidget(filterInventory, 0, 10, y + 15)
             .setBackgroundTexture(GuiTextures.SLOT, GuiTextures.FILTER_SLOT_OVERLAY));
-        this.filterWrapper.initUI(y + 15 + 18 + 5, widgetGroup);
```
LAGIdiot, line 57: Why did you move these numbers to WidgetGroupFluidFilter.java?

LAGIdiot:
@huneau Do you agree with my suggestion regarding `adjustTransferRate` in `CoverPump`? And can you answer me regarding that change in `FluidFilterContainer`?

src/main/java/gregtech/common/covers/filter/FluidFilterContainer.java
huneau: Ì change to have this layout 
image

huneau:
Sorry @LAGIdiot was very busy. 
for your suggestion, why not, its your personal feeling.
And I answer for the change of fluid filter class

src/main/java/gregtech/common/covers/filter/FluidFilterContainer.java
LAGIdiot: I can see what it does, but I am interested in why did you change it. I think that previous design was better. As it would also allow us to use available space better. We could have 2 columns for modules (filter and #1141 ). With this change there is not much space left on right side and space which we get on bottom is too small for anything. Don't you think? @pyure Your take on this from UX perspective?

huneau: this doesnt fit anymore.
the new row I added make impossible to add the 3by3 square under the filter
image

LAGIdiot: Okay then, let's go with current design as it can be adjusted later when needed.

LAGIdiot:
Thanks for you hard work on this PR and for keeping with my request. I have just committed my suggestion regarding logic in `adjustTransferRate` and I will merge this.

huneau:
Thx LAGIdiot. I'm sorry to not handle your suggestion. My computer is off since 2 weeks, and I cannot dev for now.

PR #1127, opened 2020-07-04
fix: drained tank not stackable

DeckerCHAN:
Since freshly made tank has `null` NBT but drained tank as `{}` NBT tag, they are not stackable. This PR removed NBT data when tank has been drained.

### DIFF
--- a/src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityTank.java
+++ b/src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityTank.java
@@ -492,9 +492,7 @@ public class MetaTileEntityTank extends MetaTileEntity implements IFastRenderMet
             }
 
             private void removeTagWhenEmptied(boolean doDrain) {
-                if (doDrain && this.getFluid() == null &&
-                    this.container.hasTagCompound() &&
-                    this.container.getTagCompound().hasNoTags()) {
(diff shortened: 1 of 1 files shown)


src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityTank.java
```diff
@@ -477,6 +477,28 @@ public void getSubItems(CreativeTabs creativeTab, NonNullList<ItemStack> subItem
...
+            public FluidStack drain(int maxDrain, boolean doDrain) {
+                FluidStack drained = super.drain(maxDrain, doDrain);
+                this.removeTagWhenEmptied(doDrain);
+                return drained;
+            }
+
+            private void removeTagWhenEmptied(boolean doDrain) {
+                if (doDrain && this.getFluid() == null &&
+                    this.container.hasTagCompound() &&
+                    this.container.getTagCompound().hasNoTags()) {
```
LAGIdiot, line 497: There is no need to check for that TagCompound as it does not matter in this case. We are **stating** that when we drained fluid and there is none left that there is **no need** for that Tag.
```suggestion
                if (doDrain && this.getFluid() == null) {
```

DeckerCHAN:
Cheers @LAGIdiot!

Issue #1128, opened 2020-07-06
When breaking machines with a Thermal Foundation hammer only the one machine that was actually focused drops


Creysys:
**Version**
855d8a1

**Description**
When breaking machines with a Thermal Foundation hammer only the one machine that was actually focused drops

**Expected behavior**
All broken machines drop instead of just one

**Steps to reproduce**
Break at least two machines at the same time with a Thermal Foundation hammer

I am trying to create a pull request to fix this issue.

Creysys:
The same thing also happens with the Gregtech chests.
Whats interesting is that the contents of the blocks that disappear are dropped correctly however.

Creysys:
https://github.com/GregTechCE/GregTech/pull/1130

PR #1130, opened 2020-07-06
Fixed block drops for Thermal Foundation hammers

Creysys:
At breakBlock we still have access to the TileEntity so we can cache it there

breakBlock should always be called before harvestBlock

This is a "workaround" to Thermal Foundation hammers passing "null" as "te" in harvestBlock

### DIFF
--- a/src/main/java/gregtech/common/blocks/surfacerock/BlockSurfaceRockNew.java
+++ b/src/main/java/gregtech/common/blocks/surfacerock/BlockSurfaceRockNew.java
@@ -1,6 +1,7 @@
 package gregtech.common.blocks.surfacerock;
 
 import gregtech.api.items.toolitem.IScannableBlock;
+import gregtech.api.pipenet.tile.IPipeTile;
 import gregtech.api.unification.material.Materials;
 import gregtech.api.unification.material.type.Material;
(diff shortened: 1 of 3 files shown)


Creysys:
Closes #1128

LAGIdiot:
Is it possible to extend this PR to cover all blocks which suffers same problems from being broken by Thermal Foundation Hammer not just machines?

Creysys:
> Is it possible to extend this PR to cover all blocks which suffers same problems from being broken by Thermal Foundation Hammer not just machines?

I think i got them all now :)

src/main/java/gregtech/common/blocks/surfacerock/BlockSurfaceRockNew.java
```diff
@@ -1,6 +1,7 @@
 package gregtech.common.blocks.surfacerock;
 
 import gregtech.api.items.toolitem.IScannableBlock;
+import gregtech.api.pipenet.tile.IPipeTile;
```
LAGIdiot, line 4: I guess that this line get here by mistake as this class does not have nothing to do with pipenet.
```suggestion
```

LAGIdiot:
Thanks for your hard work regarding this PR.

PR #1133, opened 2020-07-09
Unify the zh_cn translations for fluid_extractor

RikaSugisawa:


src/main/resources/assets/gregtech/lang/zh_cn.lang
```diff
@@ -1979,8 +1979,8 @@ gregtech.machine.fluid_extractor.tooltip=从物品中提取流体
 
 gregtech.machine.fluid_extractor.lv.name=基础流体提取机
 gregtech.machine.fluid_extractor.mv.name=进阶流体提取机
-gregtech.machine.fluid_extractor.hv.name=快速流体灌装机
-gregtech.machine.fluid_extractor.ev.name=涡轮流体灌装机
+gregtech.machine.fluid_extractor.hv.name=快速流体提取机
```
LAGIdiot, line 1982: Should not this one be more like: 进阶流体提取机 II to be constant with other machines? I mean most HV and EV tiers of machines are using same name as MV tier with just Roman numerals at end.

RikaSugisawa: Updated. Thanks.

LAGIdiot:
Thanks for your hard work!

Issue #1129, opened 2020-07-06
[Suggestion] Pump Configuration for Only Pump In/Out Portion of Liquid


DeckerCHAN:
# Case1: 

**AS** a GTCE player
**GIVEN** A Wooden tank with pump cover
**AND** The tank has 4000mb of water
**AND** Pump cover has been set to **Export**
**AND** Pump cover has connected to another tank with enough empty space
**WHEN** I configure Pump covert
**AND** I can set the Portion of Liquid to 50%
**THEN** Pump will drain water from Wooden tank
**AND** Pump will stop drain when Wooden tank has 2000mb of water left

# Case2: 

**AS** a GTCE player
**GIVEN** A empty Wooden tank with pump cover
**AND** Pump cover has been set to **Import**
**AND** Pump cover has connected to another tank with enough water
**WHEN** I configure Pump covert
**AND** I can set the Portion of Liquid to 50%
**THEN** Pump will pump water from another to Wooden tank
**AND** Pump will stop pump when Wooden tank has 2000mb of water

DeckerCHAN:
https://github.com/GregTechCE/GregTech/issues/952

PR #1135, opened 2020-07-09
Pump fluid portion

DeckerCHAN:
Adding features for #1129


DeckerCHAN:
image
image
image
image

src/main/java/gregtech/api/util/FluidTankUtils.java
```diff
@@ -0,0 +1,34 @@
+package gregtech.api.util;
+
+import net.minecraftforge.fluids.FluidStack;
+import net.minecraftforge.fluids.capability.IFluidHandler;
+import net.minecraftforge.fluids.capability.IFluidTankProperties;
+
+import java.util.function.Predicate;
+
+public class FluidTankUtils {
```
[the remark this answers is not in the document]
DeckerCHAN: Renamed

LAGIdiot:
Do you need any help regarding requested changes?

DeckerCHAN:
@LAGIdiot Ah I just need some more tests.

LAGIdiot:
Do you need any help regarding requested changes?

src/main/java/gregtech/api/util/GTFluidUtils.java
```diff
@@ -9,13 +9,18 @@
 public class GTFluidUtils {
 
     public static int moveHandlerFluids(IFluidHandler sourceHandler, IFluidHandler destHandler, int transferLimit) {
+        return moveHandlerFluids(sourceHandler, destHandler, transferLimit, (fluid) -> true);
+    }
+
+
```
LAGIdiot, line 15: Please don't leave multiple empty lines if it is not required.

LAGIdiot:
I have updated some of my comments as of reaction to last push. And I am glad that it is still worked on and as mention before if you need help I am available.

DeckerCHAN:
> Did I do something wrong or is this PR missing some commits?

The issue was due to **previous** behaviors:  
When current liquid amount is under limit 
Then stop the transfer
When current liquid is NOT under limit
Then Ignore limit
And transfer liquid amount based on Pump capability

After the fix, the behavior is:
When there is a margin between current amount and amount limit
Then calculate final transfer amount by both pump capability and margin, pick the one which is smaller
Then do the transfer

src/main/java/gregtech/common/covers/CoverPump.java
```diff
@@ -105,37 +116,68 @@ protected int doTransferFluids(int transferLimit) {
...
+            if (currentFluid == null || currentFluid.amount == 0 || !fluidFilter.testFluidStack(currentFluid)) continue;
+
+            FluidStack currentFluidCopy = currentFluid.copy();
+            currentFluidCopy.amount = fluidLeftToTransfer;
+            FluidStack canExtractFluid = dest.drain(currentFluidCopy, false);
+            if (canExtractFluid == null || canExtractFluid.amount == 0) continue;
+            int canInsertAmount = me.fill(canExtractFluid, false);
+            if (canInsertAmount == 0) continue;
+            int liquidDeltaLimitFromPortion = Math.round((sourceTankProperties.getCapacity() * ((float) this.percentageInTank / 100))) - sourceTankProperties.getContents().amount;
+            int finalTransferAmount = Math.min(liquidDeltaLimitFromPortion, canInsertAmount);
```
LAGIdiot, line 145: Maximal liquid amount that can be transferred should be calculated from `transferLimit` before even attempting draining fluid as dry run because there may be no need for it as tank already has enough fluid.

src/main/java/gregtech/common/covers/CoverPump.java
```diff
@@ -105,37 +116,68 @@ protected int doTransferFluids(int transferLimit) {
...
+            int actualFilled =   me.fill(fluidExtracted, true);
+            fluidLeftToTransfer -= actualFilled;
+
+        }
+        return transferLimit - fluidLeftToTransfer;
+
+
+    }
+
+    protected int doExportFluid(IFluidHandler me, IFluidHandler dest, int transferLimit) {
```
LAGIdiot, line 158: Same comments as in `doImportFluid` applies.

LAGIdiot:
I am also having problem with importing fluid setup as on pictures below.



Pump on empty bronze tank is set to import 10% but nothing happens even when wooden tank from which it should import is full.

LAGIdiot:
There is also problem with spacing when 100% is used. Button for Manual Import/export should be moved on next row.

LAGIdiot:
I would also like better control over setting percentage. Players should be allow to set what ever percentage they need (want). This could be implemented by making buttons to change percentage by 1 and by 10 when clicked with Shift.

LAGIdiot:
Do you need any help regarding requested changes?

DeckerCHAN:
I am still figuring out fluid import for multi tanks.

LAGIdiot:
Do you need any help regarding requested changes?

src/main/java/gregtech/common/covers/CoverPump.java
```diff
@@ -105,37 +116,68 @@ protected int doTransferFluids(int transferLimit) {
...
-
-                    fluidLeftToTransfer -= fluidStack.amount;
-                    if (fluidLeftToTransfer == 0) break;
-                }
+        for (IFluidTankProperties sourceTankProperties : dest.getTankProperties()) {
+
+            FluidStack currentFluid = sourceTankProperties.getContents();
+            if (currentFluid == null || currentFluid.amount == 0 || !fluidFilter.testFluidStack(currentFluid)) continue;
+
+            FluidStack currentFluidCopy = currentFluid.copy();
```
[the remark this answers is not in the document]
DeckerCHAN: No more copy.

src/main/java/gregtech/common/covers/CoverPump.java
DeckerCHAN: Added `if (fluidLeftToTransfer == 0) break;` before drain

DeckerCHAN:
> 
> 
> I would also like better control over setting percentage. Players should be allow to set what ever percentage they need (want). This could be implemented by making buttons to change percentage by 1 and by 10 when clicked with Shift.

Fixed in https://github.com/GregTechCE/GregTech/pull/1135/commits/75bf4ab3da8c6a0a6ab9838abd6a79af44051733

LAGIdiot (changes requested):
Good work updating this PR I am glad that you did not abandon this.

But I am going to be honest I am disappointed with you work. I am not talking about your skills or speed. I am talking about ignoring about half of things I requested or pointed out. You did not use any of my offered help and instead produced this half baked solution. You closed some conversations as resolved when they clearly are not. And your code still does not follow required code style by this repository which I already requested to be fixed.

Now don't get me wrong I am glad that we have contributors. But I have to maintains some quality level and every review is quite time consuming for me as I even test implementation in game. I like doing code reviews but I hate wasting time on pointing same things again and again.

So if you feel like continuing this PR please do. I would be glad to land you help if you would needed. And I will be here for another review but only when everything is resolved.

LAGIdiot:
I am marking this as stale and in current state I don't feel like accepting this PR as per comments above. If you have anything to add to this either as discussion point or from code perspective please do or it will be rejected.

PR #1139, opened 2020-07-11
Add a Config option for the number of byproduct slots for the MV macerator

ALongStringOfNumbers:
#1104 except I did not break the commits. My apologies for the trouble, it was among my first time using the git integration in IntelliJ. 

As brought up in that PR, I have made this a config option. The config option is the number of byproduct slots, from 0 to 1. Zero byproduct slots will be the current behavior, and is the default option. One byproduct slot is the old functionality. This option requires a MC restart to take effect.


LAGIdiot:
This one does not sits right with current GTCE concept. So I will reject this one.

Additional Notes:
Issues regarding Macerator not providing enough byproducts were mitigate by #1145.

PR #1143, opened 2020-07-13
Clarify overclocking tooltip in GUI

ALongStringOfNumbers:
Currently the overclocking tooltips say "Enable Overclocking" when overclocking is enabled and "Disable Overclocking" when overclocking is disabled. This could get confusing for players thinking they have to click the GUI element to enable overclocking, when in fact overclocking is already enabled and vice-versa. This PR changes the tooltip to provide clearer information on the current state of the machine overclocking.


pyure:
This is one of my favorite UX tweaks in app development in general and gaming in particular.  Do you think its worth taking a small step further? "Overclocking is Disabled.  Click to Enable" ?

ALongStringOfNumbers:
The Click to Enable part might be useful, in case people do not realize that the GUI element is a button you can press.

pyure:
You're probably right; its not too long for my taste but I can see how it might be too long for some screens.
Critically, your original tweak addressed the main issue that the button text was literally backwards.  I recommend just rolling with that.

LAGIdiot:
Thanks for you hard work on fixing this tooltip. I would like you to use text proposed by @pyure . Regarding problem with length you can use `/n` it for putting second sentence on next line.

ALongStringOfNumbers:
Updated the tooltip to bring the new information onto a new line

gui2

PR #1145, opened 2020-07-14
Add support custom chance functions.

eutropius225:
A custom chance function can now be set, which is given the base chance, the increase per tier, and the byproduct increase tier, and returns the true chance with which the chance entry will be dropped.

This allows for reverting to old byproduct chances, for example, which is one of the main reasons Omnifactory is still running and outdated version.

The function is global, and can be defined in CraftTweaker at any time.
The benefit of it being global is that the localization value can just be changed by the modpack author to reflect the changed byproduct scalings.

### DIFF
--- a/src/main/java/gregtech/api/recipes/Recipe.java
+++ b/src/main/java/gregtech/api/recipes/Recipe.java
@@ -176,7 +176,7 @@ public class Recipe {
             chancedOutputsList = chancedOutputsList.subList(0, Math.max(0, maxChancedSlots));
         }
         for (ChanceEntry chancedOutput : chancedOutputsList) {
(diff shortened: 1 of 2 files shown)


LAGIdiot:
Can you please post small script with usage of this function.

eutropius225:
This script:
```zs
import mods.gregtech.recipe.RecipeMap;

RecipeMap.chanceFunction = function(chance as int,
                                    boostPerTier as int,
                                    tier as int) as int {
    return chance * pow(2, tier);
};
```
will replace the function such that it replicates the old behaviour.

Then adding
```lang
gregtech.recipe.chance=Chance: %s%%; doubled each tier
```
to a lang file will also change the tooltip text to reflect the changes.

LAGIdiot:
Thanks for you hard work on this and for providing us with CT script for usage,

PR #1148, opened 2020-07-16
Modify Implosion Compressor CraftTweaker integration for the ability …

ALongStringOfNumbers:
…to specify explosives types

This PR modifies the existing CraftTweaker support for the Implosion Compressor to add the ability to set the item designated as the "explosives" instead of hardcoding the explosives to only use TNT. This change was made in a fashion to keep compatibility with existing scripts, which will still default to using TNT if the old syntax is used.

An example script to test this is

` import mods.gregtech.recipe.RecipeMap;
import crafttweaker.item.IItemStack;

val implosion as RecipeMap = RecipeMap.getByName("implosion_compressor");

implosion.recipeBuilder()
	.inputs(<minecraft:cobblestone>)
	.outputs(<minecraft:stone>)
	.property("explosives", <gregtech:meta_item_1:32629> * 16)
	.duration(20).EUt(20).buildAndRegister();

implosion.recipeBuilder()
	.inputs(<minecraft:cobblestone>)
	.outputs(<minecraft:stone>)
	.property("explosives", 4)
	.duration(20).EUt(20).buildAndRegister(); `

This script results in the recipes

implecompress

As a note, the Implosion Compressor by default cuts the given amount of explosives in half, which is why the numbers are different from the script.

### DIFF
--- a/src/main/java/gregtech/api/recipes/RecipeBuilder.java
+++ b/src/main/java/gregtech/api/recipes/RecipeBuilder.java
@@ -81,6 +81,10 @@ public abstract class RecipeBuilder<R extends RecipeBuilder<R>> {
         return false;
     }
 
+    public boolean applyProperty(String key, ItemStack item) {
+        return false;
+    }
+
     public R inputs(ItemStack... inputs) {
         return inputs(Arrays.asList(inputs));
     }
--- a/src/main/java/gregtech/api/recipes/builders/ImplosionRecipeBuilder.java
+++ b/src/main/java/gregtech/api/recipes/builders/ImplosionRecipeBuilder.java
@@ -16,6 +16,7 @@ import stanhebben.zenscript.annotations.ZenMethod;
 public class ImplosionRecipeBuilder extends RecipeBuilder<ImplosionRecipeBuilder> {
 
     protected int explosivesAmount;
+    protected ItemStack explosivesType;
 
     public ImplosionRecipeBuilder() {
     }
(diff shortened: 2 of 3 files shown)


Agrax:
I mean... The TNT was usable in these recipes once. 
I, personally, prefer the GT5u way: 8 dynamites, 2 TNTs or 1 industrial TNT... The CT integration might not be the... optimal way. Is there any check if the material is truly an explosive?
I don't know coding that much, but i'd love to make sure that you can't use ie. cobble as an "explosive" material.

That's why the values were hardcoded. If you want any, uh. integration, i'd go the PBF way:
Assign an valueExplosive to each, uh... explosive material and require the recipe a set value of explosibility.
Allow ppl to somehow assign other mod's explosives a value... But that's just my 0,02$

Saereth:
personally I like this change more than hard coding, gives us the flexibility to use things like alumentum as a valid explosive material in packs or tiny tnt from ae, things of that nature

pyure:
I have to agree with Saereth.  Confirming that the input qualifies as an "explosive" isn't really a decent responsibility of the mod.  If I want to compress diamonds with the explosive power of a banana for my weird pack that should be my decision (problem).  It costs the mod nothing to allow the flexibility (and its not like its easy to screw it up as the pack dev)

LAGIdiot:
I am okay with this change as it brings flexibility regarding "explosive" use for compression.

Thanks for your hard work on this PR and for providing us with example of use CT script.

PR #1150, opened 2020-07-16
Remove redundant DISABLE_DECOMPOSITION flag

ALongStringOfNumbers:
In 13cde74 the ability for several materials to be broken down was removed. However, a redundant DISABLE_DECOMPOSITION flag was added to Polyphenylene Sulfide, which already had that flag. This PR simply removes the redundant flag.


LAGIdiot:
Thanks for spotting and fixing this issue.

PR #1156, opened 2020-07-21
Config value for plasma duration

programgames:
This pr add the ability to change plasma duration in turbine by tweaking values in config file.

### DIFF
--- a/src/main/java/gregtech/common/ConfigHolder.java
+++ b/src/main/java/gregtech/common/ConfigHolder.java
@@ -97,6 +97,27 @@ public class ConfigHolder {
     @Config.RequiresMcRestart
     public static int gasTurbineBonusOutput = 6144;
 
+    @Config.Comment("Sets the duration of Helium Plasma of Plasma turbine")
+    @Config.RequiresMcRestart
+    public static int heliumPlasmaDuration = 2560;
+
+    @Config.Comment("Sets the duration of Nitrogen Plasma of Plasma turbine")
+    @Config.RequiresMcRestart
+    public static int nitrogenPlasmaDuration = 4032;
+
+    @Config.Comment("Sets the duration of Oxygen Plasma of Plasma turbine")
+    @Config.RequiresMcRestart
+    public static int oxygenPlasmaDuration = 4096;
+
+    @Config.Comment("Sets the duration of Iron Plasma of Plasma turbine")
+    @Config.RequiresMcRestart
+    public static int ironPlasmaDuration = 103219;
+
+    @Config.Comment("Sets the duration of Nickel Plasma of Plasma turbine")
+    @Config.RequiresMcRestart
+    public static int nickelPlasmaDuration = 106905;
+
+
     public static class VanillaRecipes {
 
         @Config.Comment("Whether to nerf paper crafting recipe. Default is true.")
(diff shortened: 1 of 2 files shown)


src/main/java/gregtech/common/ConfigHolder.java
```diff
@@ -97,6 +97,27 @@
     @Config.RequiresMcRestart
     public static int gasTurbineBonusOutput = 6144;
```
LAGIdiot, line 99: We should probably create whole new section for Plasma Turbine in _MachineSpecificConfiguration_ which we can put these in.

Also we could move `plasmaTurbineBonusOutput` to that section which would introduce compatibility issue but we can mitigate it by mentioning it in change log. What you think @pyure ?

src/main/java/gregtech/common/ConfigHolder.java
```diff
@@ -97,6 +97,27 @@
...
 
+    @Config.Comment("Sets the duration of Helium Plasma of Plasma turbine")
```
LAGIdiot, line 100: Please mention default value in comment for all newly added records.

src/main/java/gregtech/common/ConfigHolder.java
```diff
@@ -97,6 +97,27 @@
...
+
+    @Config.Comment("Sets the duration of Iron Plasma of Plasma turbine")
+    @Config.RequiresMcRestart
+    public static int ironPlasmaDuration = 103219;
+
+    @Config.Comment("Sets the duration of Nickel Plasma of Plasma turbine")
+    @Config.RequiresMcRestart
+    public static int nickelPlasmaDuration = 106905;
+
+
```
LAGIdiot, line 120: Please don't leave more then one empty line when not required,

src/main/java/gregtech/common/ConfigHolder.java
pyure: Sections are good.  Personally I'd leave the old value there and override it with the one one, to preserve compatibility.  Probably not crucial though since this doesn't have the capacity to blow up bases.

pyure: (I'd edit the old comment to indicate that the field is deprecated / overriden)

LAGIdiot: Sounds good we can do that. Let's add new class `ConfigHelper.java` which will handle checking for both values and if one is not set to default it will be use. Deprecation of used old value can be indicated in `@Config.Comment` with info regarding new place for value. And we can remove old one later.

huneau:
I dont understand the purpose of that PR.
you can change it using CT ? 
as a mod dev dependency, this will break my code because I change recipes and I need to have the exact value. if someone changes that value I cannot overwrite in my mod.

LAGIdiot:
@huneau Thank you for voicing you concerns but I don't see any problem here. Values supplied through config will be used for initialization of given fuels and can be updated later in mod loading process like via CT or overridden by your mod (which I am not fond of). 

Only potential problem witch I see with this is situation when these values are overridden later and some will come to report bug that we are not loading data from config.

LAGIdiot:
Do you need any help regarding requested changes?

LAGIdiot:
I am closing this issue as stale. If you plan to continue working on this PR feel free to reopen.

PR #1185, opened 2020-08-08
Cleanup the Config file

ALongStringOfNumbers:
This PR cleans up the config file by adding specified defaults to all entries, and unifying the style of the existing specified defaults. In addition, some comments were adjusted for grammatical correctness.


LAGIdiot:
Thanks for your hard work.

PR #1208, opened 2020-08-23
Fix JEI ghost items/fluids.

detav:


detav:
Sure, screenshots are from basic item filters but this fix is valid for anything that uses the PhantomSlotWidget/PhantomFuidWidget classes (like fluid filters and the work table).

**What was the bug:**
The phantom slots from JEI were not aligned with the slots from GT. This was just a visual rendering bug.


**How was fixed:**
Removed the shift caused by calculating the phantom rectangle position twice.


**Further information:**
Seems like toScreenCoords was created for this and is now not being used by anything.

LAGIdiot:
Thank you for providing description and screenshots for this error. It seems that it was behaving even worse for you then me so thanks for fixing it. Regarding `toScreenCoords` thank you for informing us, I will take a look at it in next planned clean up.

PR #1213, opened 2020-08-24
Update zh_cn.lang

### LINKED ISSUE #1206: [BUG] Chinese translation mistake about fluid extractor
The advanced fluid extractor (HV and EV version) has been translated to "进阶流体灌装机" II and III, which actually means fluid canner. It should be "进阶流体提取机" II and III. The translation of LV and MV fluid extractor is right.
https://github.com/GregTechCE/GregTech/blob/a3e6fdca2893c19e97d23f5d2f682830e5fb57b2/src/main/resources/assets/gregtech/lang/zh_cn.lang#L1964-L1974

LovelyCatHyt:
Fixed #1206 ：Change “进阶流体灌装机” into “进阶流体提取机” for fluid extractor (hv and ev).


LAGIdiot:
Thanks for your hard work.

PR #1214, opened 2020-08-24
Scanner tooltip clarification

DoctorWeirdGuy:
The tooltip said to rightclick to analyze surface rocks, but that never analyzed them. Instead you have to hold right mouse to analyze surface rocks. It was pretty confusing.


LAGIdiot:
Thanks for your hard work.

PR #1230, opened 2020-09-02
chore: Added button which allowing hopper extract items from machine …

DeckerCHAN:
**What:**
Since some items won't be consumed during machine processing, automation to re-use the machine with another recipe can be hard to achieve.

**How solved:**

Introducing a button to machine GUI that allowing others (such as hopper) to extract items from import slots.

**Outcome:**

Now users can switch button to allowing others (such as hopper) to extract items from import slots.


LAGIdiot:
Thank you for taking your time and creating this PR.

On first look it seems like good idea. But after few moments of thinking about it I am not sure this can even work properly. Can you please describe or show real use case of this. As I was not able to use it even once successfully.

Things I tried: 
I have Assembling Machine with Hopper underneath and extracting from input is enabled. Now I want to make Weighted Pressure Plate and get circuit extracted. So I put in 2x Gold Plate and 1x Integrated Circuit. Result of crafting is nothing because everything was extracted by hopper. Another try via automation (I use hopper to put things in) and another unsuccessful craft, everything got extracted. Now I filter hopper to pull only circuit. And another unsuccessful craft, circuit got pulled out too soon, plates stayed in machine.

I have  Assembling Machine, auto output and extract from input enabled. Now I want to make Weighted Pressure Plate and get circuit extracted. So I put in 2x Gold Plate and 1x Integrated Circuit. Craft successful but circuit did not get extracted.

DeckerCHAN:
The motivation of this PR was due to GTCE with RFTools Control. In that mod, it allows you write a program to interact with a specific block on a specific slot but the issue is that program can not retrieve Integrated Circuit from input slots. It turns out that you have to make a machine with fixed config which kills the versatility. 

Yes I agreed with your point and admit using hopper is problematic but the workaround is using cover to whitelist Integrated Circuit.

Or I have another solution is that when the crafting is finished, it auto-output ingredient with result together.

LAGIdiot:
I am afraid that using cover is with filter will produce same result as Hopper with filter. And that is Circuit being pulled out before crafting is initiated.

And extracting Circuit on finished crafting does not make sense to me either as there may be second crafting batch which uses same circuit (for example I used this with AE2 where one patter sometimes had more items in output then which one machine crafting cycle produced). Also you would have to NBT which circuit was used latest for extraction logic as there may already be another for next crafting batch.

LAGIdiot:
I am marking this as stale and in current state I don't feel like accepting this PR. If you have anything to add to this either as discussion point or from code perspective please do or it will be rejected.

trishume:
Just registering that I figured out a way to use this feature and then was sad that there appeared to be no way of extracting integrated circuits. I have a number of tricks I can use to emit a redstone signal when all the expected results from a craft are done, which would have made it safe to pull everything out of the input slots, since I can also make the next set of inputs not arrive until the previous recipe is done. I had a glorious challenge plan to have one array of each machine type which could make any recipe for that machine, but without this feature I can only do every recipe that can be made with a given integrated circuit or lack thereof in the machines.

PR #1237, opened 2020-09-09
Add all energy tiers

hjae78:
**What:**
This PR adds API support for voltages between UV and MAX while not changing or adding any additional hulls, hatches etc.
See [this](https://gtnh.miraheze.org/wiki/Electricity) for reference.

**How solved:**
Expanded the voltage array in GTValues and resolved all conflicting MetaTileEntity ID's and recipe errors.

**Outcome:**
Added API support for voltages between UV and MAX.

This will cause compatibility issues for addons as they loop over GTValues.V.


huneau:
aaaaaaaaaaaaaaaaaaahhhhhhhhh Nice !

I noticed you say GTValues.V[9] maybe you can let GTValues.MAX ?

hjae78:
I thought about it but it would just be strange to have GTValues.MAX referencing wrong things in other arrays. Like if GTValues.MAX is 9 then GTValues.VN[GTValues.MAX] = "UHV" which just makes no sense.

LAGIdiot:
Thank you for taking your time and creating this PR.

**Decision:**
I am going to decline this PR as it does not bring anything significant to GTCE with all problems related to it. After quite long inspection and consideration this is not worth it addition to GTCE and that is final decision. Reasoning follows...

**Reasoning:**
If I understand correctly you are doing this to provide more possibilities to Gregicality (and possibly other addons) in regards of voltage tiers. Even if it is noble cause the price which would have GTCE and all addons pay is way too high.
We are talking about complete incompatibility with previous version. Most likely all addons will need to be fixed as they may generate blocks based on voltages - and breaking everyone is not something that can be taken lightly. Same as GTCE all addons will require "hacks" (maybe only just hard to read, strange code) to fix their MAX voltage problems in arrays (not because of tight MetaTileEntity ID ranges) but because change of voltage may blow someone base - which is game stopper. Just look how many additional lines were needed for GTCE. Even if you fix usage of magic numbers in code it will still be hard to read. Which means possibility that some place which requires adjustments for new voltages was missed - and that is end.

hjae78:
I'm disappointed by this decision. I understand your reasoning and respect your decision. However it is important to not that there is no incompatibility with previous versions. The ugly code's purpose is precisely to avoid incompatibility, to preserve MetaTileEntity ID's etc. As far as anything blowing up that should also not be the case. There would be no change to any existing blocks in worlds meaning that their associated voltage would not change. Machines only blow up when fed incorrect voltage and if no block's voltage changes then no functioning world should run into issues.

Trust me that I would do as much extensive testing as you would require for this to be accepted. I would also gladly fix any issues that arise for addons. I already help to maintain Gregicality and the difference Gregicality and SoG are not significant. 

We completely rely on your API for voltage, any solution on our end would be much, much more difficult and far more ugly. I urge you to reconsider this.

LAGIdiot:
Okay I am going to be more concrete regarding incompatibility I have mentioned. If we build GTCE with this implementation all addons and in extension all CT scripts which either relays on generation of blocks based on voltage or work with it in some not streamlined manner. Will either crash on start as they will have problem with MetaTileEntity IDs clash (which is exactly what you were fixing with ugly code for GTCE). In this case we break mod which may not get updated ever again or in long time which may block updating of modpacks. Or they will regenerate their blocks which means that previously MAX tier machine will became UHV and it may got supplied with power from real GTCE MAX voltage provider. Which is one of possible places for explosion and in this case we will betray trust of our players. Same goes for cases when some is using something like `GTValues.V[9]` for MAX voltage which will no longer work and possible causing explosion. Example: https://github.com/GregTechCE/GregTech/blob/a53edbe7d1cf42fd101464f50c580a3d73349015/src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityLargeBoiler.java#L277 (not explosive one but hard to find or at least you miss it in this PR which proves my point).

Now can you truly go around and fix all addons which rely on GTCE API or check all CT scripts? Even abandoned or in private repositories? I don't think so and that is why I don't want to take this risk. I know we can mention this in Changelog but who truly reads them?

I understand your complete dependency on GTCE API. And I am opened to adding new features to it or extending current ones but in this case we are talking about changing core part of GTCE which could have unfathomable consequences which even I as maintainer may not foresee. But I will reopen this case and mark it for possible discussion so someone may come up with other reasons why to do this or how to mitigate or even remove problems which I mentioned.

hjae78:
After consideration and looking at some of GTCE's code I think that I can achieve what I want without this PR. You are right that with this PR comes a massive undertaking of pretty much looking at every single use of the voltage array in the GTCE code and every single add-on. And you are right that even if a lot of add-ons are public repos not all of them are. In the end this PR will definitely result in a bug somehow and ruin someone's experience and that is not acceptable.

It is a shame that not all these voltages were implemented from the beginning but the rest of the code is incredibly flexible and I believe I can achieve everything I want without this. Thank you for your consideration.

LAGIdiot:
Thank you for understanding my point of view. And good luck implementing support for it on your side.

PR #1240, opened 2020-09-15
Update Rotor Holder spinning status after world reload.

### LINKED ISSUE #1200: [BUG] Rotor holder does not damage player on interaction after reload
**Describe the bug**
Normally, interacting with an active rotor holder will damage the player. This state is not properly saved or initialized on reload, and if the rotor is already at full speed, the rotor holder can be opened and accessed without taking damage.

In MetaTileEntityRotorHolder.java, isRotorLooping is only updated in incrementSpeed, which will not be called if the rotor is already at full speed. On reload, the variable will reset, and will not be updated when loading the NBT data to set the speed.
Trivial solution would be to initialize `this.isRotorLooping = currentRotorSpeed > 0` at the end of readFromNBT.

**Versions**
Forge: 14.23.5.284
GTCE: 1.9.3.531
_Modpack:_ Omnifactory v1.2.2, manually modified to update GTCE and addons
_Addons:_ Shadows of Greg 1.12.2-2.13.0; CEU 1.0.4.3

**Setup**
Tested on existing singleplayer world, likely not single/multi or world specific

**Steps To Reproduce**
1) Set up a Large Steam Turbine, add a rotor
2) Feed it enough steam, let the rotor speed up to maximum
3) Save and quit, reload.
4) Interact with the rotor holder. You can see its inventory without taking damage. Visually, it will not be spinning, either.

**Expected behavior**
Interacting with a spinning rotor holder should damage the player

ALongStringOfNumbers:
**What:**
As mentioned in #1200, the Rotor Holders will not damage the player after a world reload in some cases. In addition, when the multiblock was just starting, I never saw the rotor start spinning.

**How solved:**
As mentioned in the issue, adds an update to the read from NBT method which will manually set the fact that the rotor is spinning. Also adds an update method that I believe is missed which should fix Rotors not spinning after the Multiblock is formed.

**Outcome:**
Rotor Holders damage players after world reload.
Closes #1200

### DIFF
--- a/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityRotorHolder.java
+++ b/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityRotorHolder.java
@@ -218,6 +218,7 @@ public class MetaTileEntityRotorHolder extends MetaTileEntityMultiblockPart impl
         super.receiveInitialSyncData(buf);
         this.isRotorLooping = buf.readBoolean();
         this.rotorColor = buf.readInt();
+        getHolder().scheduleChunkForRenderUpdate();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


LAGIdiot (approved):
Thanks for your hard work. PR accepted.

Issue #1252, opened 2020-09-27
Mod Priorities config option does not work


hacatu:
I have many mods installed but just gtce 1.10.1.557, chisel 0.2.1.35, and embers rekindled 1.13hf2 should be needed to replicate.  I want gtce items to be used most of the time when there are multiple items with the same oredict, so I set
```
    S:modPriorities <
        minecraft
        biomesoplenty
        gregtech
        embers
     >
```
However, if I look at the ebf output for gtce aluminium dust, it is embers aluminium.  The alloy smelter recipe for electrum gives embers electrum.  The fluid solidifier recipe for silver blocks gives chisel silver.  Etc.

Did I set the option wrong or does it not work?  It's such a shame trying to get ore dict nonsense perfectly right because unidict and jaopca don't work, many mods don't have mod priority settings, and those settings don't always work.

hacatu:
Turns out there's another option a bit farther down: `B:useCustomModPriorities=true`.  This option is false by default.  Oops.  Maybe these options should be moved next to each other or the `useCustomModPriorities` should just be removed since the presence of entries in the list already indicates which behavior should be used.

ALongStringOfNumbers:
They could be moved into the same category, to group them together and make them more easily seen together.

LAGIdiot:
This is little bit unfortunate design but touching this in significant way would caused a regression and possibly making someone's game having unexpected behavior. When config is generated anew they should be next to each other. I think we could add comment to `modPriorities` that `useCustomModPriorities` has to be set to true to work.

LAGIdiot:
Well it seems that default config handler which we are using is ordering config entries by name with which we can't do anything (because of reasons mentioned above) so that additional comment has to suffice.

PR #1254, opened 2020-09-27
Feature/extend comment for config option modPriorities

LAGIdiot:
**What:**
Implements solution for #1252 

**Outcome:**
Added additional comment to modPriorities config option to inform of need to set useCustomModPriorities to true for it to work
Closes: #1252 

**Additional info:**
GTCE-mod-priorities-additional-comment


ALongStringOfNumbers:
Why not group them into a category, so that they will always be displayed next to each other?

LAGIdiot:
Because it will invalidate existing config options and created new ones with default setting. Even if we mention this in Changelog not everyone will register it and check their settings before running game after update. Which in end means unexpected behavior and possible problems for players.

PR #1258, opened 2020-09-28
Add NanoSaber configuration options

### LINKED ISSUE #1242: [Suggestion] Nano Saber Damage Config Option
At the moment, the Nano Saber is really expensive for what it does and some Modpacks mess with the Stats of Mobs, also Mods like Abyssalcraft have harder Mobs than Vanilla, so it would be nice to be able to tweak the damage of the Nano Saber in the Config. Maybe also a config option to allow it to ignore Armor? Maybe a number between 0 and 100 to ignore % of the Armor.

ALongStringOfNumbers:
**What:**
This PR adds configuration options for the NanoSaber, so it would be possible to buff it, as mentioned in #1242. Does not implement the armor penetration due to the fact that there is no clean way to implement it, without going outside the scope of the mod.

**How solved:**
Adds a config category for the NanoSaber configuration options, which contains entries for the NanoSaber's base damage, additional attack damage when powered, EU/t consumption, and if Zombies should be able to spawn with the NanoSabers.

**Outcome:**
Closes #1242. Adds configuration options for the NanoSaber's base damage, additional damage, EU/t consumption, and if Zombies should be able to spawn with it.

None expected

### DIFF
--- a/src/main/java/gregtech/common/ConfigHolder.java
+++ b/src/main/java/gregtech/common/ConfigHolder.java
@@ -86,6 +86,9 @@ public class ConfigHolder {
     @Config.Comment("Category that contains configs for machines with specific behavior")
     public static MachineSpecificConfiguration machineSpecific = new MachineSpecificConfiguration();
 
+    @Config.Comment("Category that contains configs for the NanoSaber")
+    public static NanoSaberConfiguration nanoSaberConfiguration = new NanoSaberConfiguration();
+
     @Config.Comment("Sets the bonus EU output of Steam Turbines. Default: 6144")
     @Config.RequiresMcRestart
     public static int steamTurbineBonusOutput = 6144;
@@ -128,5 +131,26 @@ public class ConfigHolder {
     public static class MachineSpecificConfiguration {
         @Config.Comment("Array of blacklisted dimension IDs in which Air Collector does not work.")
         public int[] airCollectorDimensionBlacklist = new int[]{};
+    }
+
+    public static class NanoSaberConfiguration {
+
+        @Config.RangeDouble(min = 0, max = 100)
+        @Config.Comment("The additional damage added when the NanoSaber is powered. Default: 20f")
+        @Config.RequiresMcRestart
+        public double nanoSaberDamageBoost = 20;
+
+        @Config.RangeDouble(min = 0, max = 100)
+        @Config.Comment("The base damage of the NanoSaber. Default: 5f")
+        @Config.RequiresMcRestart
+        public double nanoSaberBaseDamage = 5;
+
+        @Config.Comment("Should Zombies spawn with NanoSabers? Default: true")
+        public boolean zombieSpawnWithSabers = true;
+
+        @Config.RangeInt(min = 1, max = 512)
+        @Config.Comment("The EU/t consumption of the NanoSaber. Default: 64")
+        @Config.RequiresMcRestart
+        public int energyConsumption = 64;
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/gregtech/common/ConfigHolder.java
```diff
@@ -129,4 +132,25 @@
         @Config.Comment("Array of blacklisted dimension IDs in which Air Collector does not work.")
         public int[] airCollectorDimensionBlacklist = new int[]{};
     }
+
+    public static class NanoSaberConfiguration {
+
+        @Config.RangeDouble(min = 0, max = 100)
+        @Config.Comment("The additional damage added when the NanoSaber is powered. Default: 20f")
```
LAGIdiot, line 139: I think that players should be able to copy and paste default value to config and it should work so this value should be 20.0

src/main/java/gregtech/common/ConfigHolder.java
```diff
@@ -129,4 +132,25 @@
...
+        @Config.RequiresMcRestart
+        public double nanoSaberDamageBoost = 20;
+
+        @Config.RangeDouble(min = 0, max = 100)
+        @Config.Comment("The base damage of the NanoSaber. Default: 5f")
+        @Config.RequiresMcRestart
+        public double nanoSaberBaseDamage = 5;
+
+        @Config.Comment("Should Zombies spawn with NanoSabers? Default: true")
```
LAGIdiot, line 148: I think this needs to mention in which case zombie can spawn with it.

src/main/java/gregtech/common/items/MetaItem2.java
```diff
@@ -69,7 +70,7 @@ public void registerSubItems() {
         POWER_UNIT_HV = addItem(575, "power_unit.hv") .addComponents(ElectricStats.createElectricItem(1600000L, GTValues.HV)).setMaxStackSize(8);
         JACKHAMMER_BASE = addItem(576, "jackhammer_base").addComponents(ElectricStats.createElectricItem(1600000L, GTValues.HV)).setMaxStackSize(4);
 
-        NANO_SABER = addItem(577, "nano_saber").addComponents(ElectricStats.createElectricItem(4000000L, GTValues.HV)).addComponents(new NanoSaberBehavior(5.0f, 20.0f, 64)).setMaxStackSize(1);
+        NANO_SABER = addItem(577, "nano_saber").addComponents(ElectricStats.createElectricItem(4000000L, GTValues.HV)).addComponents(new NanoSaberBehavior((float)ConfigHolder.nanoSaberConfiguration.nanoSaberBaseDamage, (float)ConfigHolder.nanoSaberConfiguration.nanoSaberDamageBoost, ConfigHolder.nanoSaberConfiguration.energyConsumption)).setMaxStackSize(1);
```
LAGIdiot, line 73: Using it here with all that casting does not feel right to me. I think that `NanoSaberBehavior` should have parameter-less constructor and all getting values from config should be done in it.  Also insides of `NanoSaberBehavior` should be changed to use double instead of floats to prevent all that casting.

LAGIdiot (changes requested):
Thank you for your work on this PR.
There are few areas which needs adjustments.

src/main/java/gregtech/common/ConfigHolder.java
ALongStringOfNumbers: This has been done.

src/main/java/gregtech/common/ConfigHolder.java
ALongStringOfNumbers: More clarification has been added.

src/main/java/gregtech/common/items/MetaItem2.java
ALongStringOfNumbers: Moved to a parameter-less constructor, with values populated based on the config option. Values additionally moved to doubles to prevent casting.

LAGIdiot (approved):
Thank you for addressing my requests.

Issue #1265, opened 2020-10-09
[Suggestion] manually setting covers on output side to import mode should not need to screwdrive that side before to allow input from the output side


PrototypeTrousers:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when my fancy automation setup doesn't work because i forgot to click on a specific side with a screwdrive.

**Describe the solution you'd like**
For example, placing an conveyor importing on the same side as the output side of a machine wont actually "just work", even tought its explicit function is to import.

**Additional context**
Would save some durability from the screwdrive and make automation a bit more "smooth".

PR #1266, opened 2020-10-09
Covers on import mode take precedence over tile entity "disallow input from output side"

PrototypeTrousers:
This PR allows covers explicitly set to import mode to work on the auto output side of a machine, without needing to right click the face (that might already be covered) with a screwdriver.

suggested fix for Issue #1265

This PR covers GTCE covers only. trying to insert from the auto-output side externally will still need to be allowed by using a screwdriver on it.


LAGIdiot (changes requested):
I don't think this is correct implementation for this. It should be handled in override of `placeCoverOnSide` which should just set `allowInputFromOutputSide` probably through some new interface so implementation of covers don't have to be checked.

PrototypeTrousers:
So placing a cover ( arms, conveyors and pumps) on the output side should:
- Set the tileentity allowInputFromOutputSide to true on placement.
- Print that change so the player knows it affects that TE permanently, even if the cover gets removed or change the output side.
When the output side gets changed to a already placed cover:
- Redo the steps above.

i think the java course i took was a little too basic lol

this works as above but i try to see how a interface would work

PrototypeTrousers:
now it does this
Set allowInputFromOutputSide to true on cover placement, if the placed side is the current output of the machine.
Set allowInputFromOutputSide to true on changing the output side with a side already placed with a cover

Checks for capability of covers so doesnt trigger for covers that do not interact, at all with items and fluids (shutter, controller)

its a cover feature, so if a player wants externally inserting things to the output side (for example crafting with Applied Energistics Interfaces) will need to use the screwdriver on the output side.

src/main/java/gregtech/common/covers/CoverConveyor.java
```diff
@@ -350,7 +350,7 @@ public GroupItemInfo(Object filterSlot, Set<ItemStackKey> itemStackTypes, int to
 
     @Override
     public boolean canAttach() {
-        return coverHolder.getCapability(CapabilityItemHandler.ITEM_HANDLER_CAPABILITY, attachedSide) != null;
+        return coverHolder.getCapability(CapabilityItemHandler.ITEM_HANDLER_CAPABILITY, null) != null;
```
LAGIdiot, line 353: Why is this required? This does not seems right at all.

src/main/java/gregtech/api/metatileentity/SimpleMachineMetaTileEntity.java
```diff
@@ -90,12 +91,30 @@ public boolean onWrenchClick(EntityPlayer playerIn, EnumHand hand, EnumFacing fa
             }
             if(!getWorld().isRemote) {
                 setOutputFacing(facing);
+                if (!isAllowInputFromOutputSide()) {
+                    if (getCoverCapability(CapabilityItemHandler.ITEM_HANDLER_CAPABILITY, facing) != null ||
+                        (getCoverCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, facing) != null)) {
+                        setAllowInputFromOutputSide(true);
+                    }
+                }
```
LAGIdiot, line 99: I don't think this should be included. Only adding cover to output side should trigger allowing input from output side.

src/main/java/gregtech/api/metatileentity/SimpleMachineMetaTileEntity.java
```diff
@@ -90,12 +91,30 @@ public boolean onWrenchClick(EntityPlayer playerIn, EnumHand hand, EnumFacing fa
...
         return super.onWrenchClick(playerIn, hand, facing, hitResult);
     }
 
+    @Override
+    public boolean placeCoverOnSide(EnumFacing side, ItemStack itemStack, CoverDefinition coverDefinition) {
+        boolean coverPlaced = super.placeCoverOnSide(side,itemStack,coverDefinition);
+        if (getOutputFacing() == side && !isAllowInputFromOutputSide()) {
+            if (getCoverCapability(CapabilityItemHandler.ITEM_HANDLER_CAPABILITY,side) != null ||
+                (getCoverCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY,side) != null)) {
+                setAllowInputFromOutputSide(true);
```
LAGIdiot, line 112: Calling `getCoverCapability` here is overkill as it does quite lot of stuff underneath which in this case is not needed or even wanted. It would be simpler just to `getCoverAtSide` and check if for existence of Interface or even call function on `CoverBehavior` which will return info if this cover should `setAllowInputFromOutputSide`.

LAGIdiot (changes requested):
Good work on updating this PR. I have added some comments which needs to be resolved. Also  please properly format your code before committing.

src/main/java/gregtech/common/covers/CoverConveyor.java
PrototypeTrousers: `canAttach` is called before placing, so if `allowInputFromOutputSide` is still set on the output side, machines that only have fluid or items output slots, will expose only one of the capabilities even though the machine may accept both.
If we check the side capability, you wont be able to place a conveyor on the output side of fluid extractor.
Now that i explained it, i think its not really needed.

src/main/java/gregtech/api/metatileentity/SimpleMachineMetaTileEntity.java
```diff
@@ -96,6 +97,18 @@ public boolean onWrenchClick(EntityPlayer playerIn, EnumHand hand, EnumFacing fa
         return super.onWrenchClick(playerIn, hand, facing, hitResult);
     }
 
+    @Override
+    public boolean placeCoverOnSide(EnumFacing side, ItemStack itemStack, CoverDefinition coverDefinition) {
+        boolean coverPlaced = super.placeCoverOnSide(side,itemStack,coverDefinition);
+        if (getOutputFacing() == side && !isAllowInputFromOutputSide()) {
+            if (getCoverAtSide(side).getCapability(CapabilityItemHandler.ITEM_HANDLER_CAPABILITY,null) != null ||
+                (getCoverAtSide(side).getCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, null) != null)) {
```
LAGIdiot, line 105: I still think this is overkill and I would rather see it as check for existence of Interface or even call function on CoverBehavior which will return info if this cover should setAllowInputFromOutputSide.

Or is there something I am not seeing and you have to do it this way?

LAGIdiot (changes requested):
Thank you for updating this PR there is still one part I am not satisfy with.

src/main/java/gregtech/api/metatileentity/SimpleMachineMetaTileEntity.java
PrototypeTrousers: still learning how java actually works, i guess. i was under the impression i needed to do it like that because only that class has access to return the actual output side.
Anyway figured out how to move that function to the cover behavior API, that had a onAttached hook, so the code would only run if the cover was actually placed. Since only conveyors/arms/pumps can move things by themselves, it doesnt make sense other covers messing with the "allowInputFromOutputSide"
Thanks for the patience.

LAGIdiot: Sure no problem about taking time to learn or work on it as long you properly communicate.
I see what you did and kind of understand why you did it but it is not correct. Fundamentally Covers were design in way that they don't need to know on what they are placed on, that is handled by `ICoverable`. By casting `coverHolder` to `SimpleMachineMetaTileEntity` you made coupling which goes against design (anti-pattern). So this has to be done other way around as Machines are aware of `CoverBehavior` and other high level Cover classes (not Covers by themselves). Which means that easiest and best solution is in `CoverBehavior` create function something like `canOverrideAllowInputFromOutputSide` which will return `false`. All Covers which support this behavior will override this function and return `true`. And then in `placeCoverOnSide` in `SimpleMachineMetaTileEntity` you will check it and set `allowInputFromOutputSide`.

LAGIdiot (approved):

PR #1267, opened 2020-10-09
fix: issue preventing battery from fully charged.

DeckerCHAN:
**What:**
Batteries can not be fully charged on high tier battery charger/buffer. This PR fixed issue.

**How solved:**
By allowing last charge operation going thought.

**Outcome:**
Now battery can be fully charged.

Shouldn't have any compatibility issue.

### DIFF
--- a/src/main/java/gregtech/api/capability/impl/EnergyContainerBatteryBuffer.java
+++ b/src/main/java/gregtech/api/capability/impl/EnergyContainerBatteryBuffer.java
@@ -43,8 +43,8 @@ public class EnergyContainerBatteryBuffer extends MTETrait implements IEnergyCon
                 ItemStack batteryStack = inventory.getStackInSlot(i);
                 IElectricItem electricItem = getBatteryContainer(batteryStack);
                 if (electricItem == null) continue;
-                if (chargeItemWithVoltageExact(electricItem, voltage, getTier(), true)) {
-                    chargeItemWithVoltageExact(electricItem, voltage, getTier(), false);
+                if (chargeItemWithVoltage(electricItem, voltage, getTier(), true)) {
+                    chargeItemWithVoltage(electricItem, voltage, getTier(), false);
                     inventory.setStackInSlot(i, batteryStack);
                     this.batterySlotsUsedThisTick.set(i);
                     if (--amperage == 0) break;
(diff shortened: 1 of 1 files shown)


src/main/java/gregtech/api/capability/impl/EnergyContainerBatteryBuffer.java
```diff
@@ -58,8 +58,9 @@ public long acceptEnergyFromNetwork(EnumFacing side, long voltage, long amperage
         return amperageUsed;
     }
 
-    private static boolean chargeItemWithVoltageExact(IElectricItem electricItem, long voltage, int tier, boolean simulate) {
-        return electricItem.charge(voltage, tier, false, simulate) == voltage;
+    private static boolean chargeItemWithVoltage(IElectricItem electricItem, long voltage, int tier, boolean simulate) {
+        long charged = electricItem.charge(voltage, tier, false, simulate);
+        return charged > 0 && charged <= voltage;
```
LAGIdiot, line 63: Is that `&& charged <= voltage` really necessary?

LAGIdiot (commented):
Thank you for creating this PR.

Can you please provide information regarding your testing process and if possible picture of fully recharged battery which was not possible to fully charge before.

ALongStringOfNumbers:
Here is my reproduction case. Image one is a general setup, where you can see an LV steam Turbine feeding through a 16x tin cable into a 16 slot LV battery buffer, which feeds through another 16x tin cable into an LV lathe. I started with all batteries in the battery buffer fully charged and then disconnected the steam turbine via rotating the output. I then let the Lathe run for a bit before stopping it. I then reconnected the Steam Turbine to let the batteries recharge. That is where I found image 2, a battery had gotten close to its maximum cap, but would not charge further.

Image 1



Image 2

DeckerCHAN:
image
I have done a test on my branch and confirm the batteries can be fully charged.

LAGIdiot:
From my testing session it seems it works as expected. Now only my question regarding code remains.

src/main/java/gregtech/api/capability/impl/EnergyContainerBatteryBuffer.java
LAGIdiot: I am suggesting this change as there is no need to check other end as `charge` can't return it anyway
```suggestion
        return charged > 0;
```

PR #1275, opened 2020-10-16
Fix #1273 Fuel Consumption for Turbines is as it was before fe429f7

### LINKED ISSUE #1273: [BUG] Revert Fuel Consumption Multiplication for Turbines
Making the Fuel usage rate for the Turbines a 256x of LV while is linear with the voltages removes the benefits of using the mechanics of the turbines.  It makes it impossible for the turbine to ever become more efficient then the normal tiers.  I propose reverting to the 64x rate so that the turbines can achieve 2x to 4x efficiency bonus to provide a benefit to crafting and maintaining it.

fclinton:
**What:**
Implement change to resolve #1273

**How solved:**
Reverted lines that affected fuel consumption in fe429f7

### DIFF
--- a/src/main/java/gregtech/common/metatileentities/multi/electric/generator/LargeTurbineWorkableHandler.java
+++ b/src/main/java/gregtech/common/metatileentities/multi/electric/generator/LargeTurbineWorkableHandler.java
@@ -26,8 +26,8 @@ public class LargeTurbineWorkableHandler extends FuelRecipeLogic {
     private MetaTileEntityLargeTurbine largeTurbine;
     private int rotorCycleLength = CYCLE_LENGTH;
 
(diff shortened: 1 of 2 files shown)


LAGIdiot:
Thank you for creating this PR. But I don't think this is correct way how to solve this issue. Turbines should provide more immersive game play then just shoveling fuel in and getting energy out. And if we are going to touch it should be upgraded all that way.

fclinton:
Seems like a moot point as generators are exactly that, the multi-block is more complex as the RPM has to be managed and the turbine blade has to be maintained.

Perhaps, while we are making the turbine more immersive, lets make generators output CO2 that you have to somehow dispose of?

At the very least this makes turbines a OK option versus the others that have better production with less maintenance until something more immersive is made.

LAGIdiot:
Changing anything is not that easy. As there is lot of implications from compatibility reasons, current saves, addons and modpack usages. I understand that changing this could help to anyone who is going to use it newly but I have to think about current users. 

So "just revert that before better solution is found" is not good enough for me. So I am going to decline this PR and I will leave conversation regarding turbines open for more suggestions until we can build something "full".

Anyway thank you for your hard work on this PR and your suggestions.

PR #1279, opened 2020-10-26
ru_ru.lang fixes.

Bombm:
Fixed Hydro-cracked and Steam-cracked liquids not being localized. Fixed uraninite and pitchblende having same names. Fixed some other small localization issues.


LAGIdiot:
Thanks for your hard work. I will accept this in good faith that it is correct as I am not able to confirm it.

PR #1282, opened 2020-10-28
Display all bucket fluid recipes for the Fluid Canner Machine

### LINKED ISSUE #1220: [Suggestion] Show Fluid Canner bucket/cell filling recipes in JEI
Some people do not know that you can fill buckets or cells with fluids in the fluid canner machine, and think that the fluid canner machine only will fill batteries. This is because there is no JEI information about filling these items in the fluid canner

Add JEI information demonstrating that the Fluid Canner can fill buckets and cells

fmarceau:
What:
As mentioned in #1220, the Fluid Canner Machine only shows the recipes related to batteries. With the proper configuration setting, we would like to show all cell-capable and bucket-capable recipes.

How solved:
Iterates through all the bucket fluids from the FluidRegistry and add the proper recipe

Outcome:
Recipes described in this PR will become available if configured accordingly.
Closes #1220

### DIFF
--- a/src/main/java/gregtech/common/ConfigHolder.java
+++ b/src/main/java/gregtech/common/ConfigHolder.java
@@ -107,7 +107,7 @@ public class ConfigHolder {
 
     @Config.Comment("Show all the cell and bucket recipes for the Fluid Canner Machine. Default false")
     @Config.RequiresMcRestart
-    public static boolean showAllBucketAndCellRecipesForFluidCannerMachine = true;
+    public static boolean showAllBucketAndCellRecipesForFluidCannerMachine = false;
 
     public static class VanillaRecipes {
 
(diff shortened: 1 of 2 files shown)


LAGIdiot:
Thank you for creating this PR. I must say this pick was quite unfortunate for first contribution.

**Fluid Canner** is not using `RecipeMap` for these recipes but it is working by scanning input when no recipe from `RecipeMap` matches and then extracting/filling it if it has `FLUID_HANDLER_ITEM_CAPABILITY`. And that is reason why it is not registered in JEI.

So this has to be handled in different way than by adding recipes to `RecipeMap` for every combination of _fluid container_, _fluid_, _Fill/Empty action_ and _all available fluid volumes (if container allows it)_ as it would be too expensive. I think that we should handle this by adding **JEI info tab** for **Fluid Canner** on which will be that information stated. @ALongStringOfNumbers what you think as author of related issue?

fmarceau:
No worries @LAGIdiot, that was a fun learning experience nonetheless. I will make the changes to use the JEI Info Tab instead.

PR #1288, opened 2020-10-30
Added a JEI Info Tab for the Fluid Canner Machines for Buckets and Fluid Cells

### LINKED ISSUE #1220: [Suggestion] Show Fluid Canner bucket/cell filling recipes in JEI

fmarceau:
**What:**
This PR adds the features that got described in #1220.

**How solved:**
The code was modified to add a JEI Info Tab to the Fluid Canner Machines which informs the player that they can use Buckets and Fluid Cells just as well.

**Outcome:**
Closes #1220


src/main/resources/assets/gregtech/lang/en_us.lang
```diff
@@ -2958,4 +2959,4 @@ gregtech.command.util.hand.fluid=Fluid Info: %s %d / %d mb; Can Fill: %s; Can Dr
 gregtech.command.util.hand.material_meta_item=Ore Prefix: %s; Material: %s
 gregtech.command.util.hand.tool_stats=Tool Stats Class: %s
 gregtech.command.util.hand.meta_item=Meta Item Name: %s
-gregtech.command.util.hand.not_a_player=This command is only usable by a player.
+gregtech.command.util.hand.not_a_player=This command is only usable by a player.
```
LAGIdiot, line 2962: Please be mindful regarding file endings so there are no necessary changes.

src/main/resources/assets/gregtech/lang/en_us.lang
```diff
@@ -2019,6 +2019,7 @@ gregtech.machine.fermenter.hv.name=Advanced Fermenter II
 gregtech.machine.fermenter.ev.name=Advanced Fermenter III
 
 gregtech.machine.fluid_canner.tooltip=Puts fluids into and out of fluid containers
+gregtech.machine.fluid_canner.jei_description=Don't you know that you could use the Fluid Canner Machine to fill Buckets and Fluid Cells?
```
LAGIdiot, line 2022: I think this should be more dry not personal info (I am not sure how to describe this). Also it should inform about possibility to fill and **empty** any **fluid containers** and then maybe mention Buckets and Fluid Cells as example.

@pyure, @ALongStringOfNumbers What you think regarding this?

LAGIdiot (changes requested):
Thanks for your hard work on this PR. I found two things which needs to be fixed (one requires little bit of discussion)

src/main/resources/assets/gregtech/lang/en_us.lang
fmarceau: @LAGIdiot, @pyure, @ALongStringOfNumbers  I've updated the PR with a proposal which takes @LAGIdiot 's suggestions into account... tell me how you feel about it.

src/main/resources/assets/gregtech/lang/en_us.lang
fmarceau: 100% agree, this is a deadly sin :) I was nervous with that first PR I guess... I fixed the problem.

src/main/resources/assets/gregtech/lang/en_us.lang
pyure: Agreed with LAGIdiot (mostly because of consistency rather than the tone itself).  Revised version looks good.

src/main/java/gregtech/integration/jei/GTJeiPlugin.java
```diff
@@ -172,5 +173,8 @@ public void register(IModRegistry registry) {
         }
 
         registry.addRecipeCatalyst(MetaTileEntities.WORKBENCH.getStackForm(), VanillaRecipeCategoryUid.CRAFTING);
+
+        Arrays.stream(MetaTileEntities.FLUID_CANNER).forEach(machine ->
+            registry.addIngredientInfo(machine.getStackForm(), VanillaTypes.ITEM, "gregtech.machine.fluid_canner.jei_description"));
```
LAGIdiot, line 178: Please change this to standard foreach so it is consistent with rest for registrations (for example: https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/integration/jei/GTJeiPlugin.java#L165)

LAGIdiot (changes requested):
Thank you for updating this PR. I have added one more comment for section which I missed previously.

src/main/resources/assets/gregtech/lang/en_us.lang
LAGIdiot: Revise version accepted

LAGIdiot (approved):
Thanks for your hard work on this PR. Approved and accepted.

Issue #1277, opened 2020-10-18
[Suggestion] Some way to remove fluid from input hatches


ALongStringOfNumbers:
**Is your feature request related to a problem? Please describe.**
Sometimes players accidentally put a batch of fluid into a multiblock Input Hatch, and don't realize there is no way to retrieve said fluid.

**Describe the solution you'd like**
The ability to somehow remove fluid from multiblock input hatches. The only possible way that I can think of doing it off the top of my head is shift + right click with a bucket/container (as regular right click just brings up the gui).

huneau:
@LAGIdiot you can still my hand pump could be a cool feature for default GTCE https://github.com/Gregicality/gregicality/blob/1ed741711f7512140024c99b25d0eb08134929b7/src/main/java/gregicadditions/item/behavior/HandPumpBehavior.java#L23

LAGIdiot:
@ALongStringOfNumbers  thank you for creating this suggestion I can understand frustration for loosing fluids to this as it happened to me before too. I think we could allow extraction of these fluids via GUI. I don't want to make this accessible by fluid automation and I don't feel like we need any special item for it. And then it will work same way Item Input where you can too extract what you put in.

@huneau thank you for suggesting it but I am not going to use it as I don't feel there is need for it.

fmarceau:
@LAGIdiot should it be possible to extract the liquid with a Bucket or a EIO Tank the same way we do for GT Machines? It would be a more flexible solution than simply voiding it and as long as you have a EIO Tank it would achieve the same thing.

LAGIdiot:
@fmarceau this issue is about providing way how to extracted back so it does not have to voided. As mentioned in my previous comment solution that is acceptable for me is do it through GUI by passing empty fluid container to it. As I don't want to allow any fluid automation for it.

fmarceau:
@LAGIdiot that's what I meant by using the example of a GT Machine and that you can use a fluid container such as a EIO Tank to extract the fluid through the GUI. I am still quite newb with modded minecraft programming but I will see if I can pull it out.

PR #1295, opened 2020-11-01
Allow for filling and emptying Fluid Hatches through the GUI

fmarceau:
**What:**
Based on #1277, this PR allows for filling and emptying Fluid Hatches.

**How solved:**
The TankWidget which Fluid Hatches are relying on already has that feature, I only had to enable it.

**Outcome:**
Closes #1227 

**Additional info:**
Please note that it is not allowed to fill an Output Fluid Hatch with a fluid container, only input hatches are supported to that regard.


LAGIdiot:
I had quite different idea about how this should be implemented. I was thinking about using `FluidContainerSlotWidget` to accept empty container and put fluids to it as it does for Export Hatch.

Even if your implementation is interesting I would rather not change feel for this block. And also it would not be obvious for players that `TankWidget` is clickable.

fmarceau:
> I had quite different idea about how this should be implemented. I was thinking about using `FluidContainerSlotWidget` to accept empty container and put fluids to it as it does for Export Hatch.
> 
> Even if your implementation is interesting I would rather not change feel for this block. And also it would not be obvious for players that `TankWidget` is clickable.

Ok, I will see what I can do to adapt it.

LAGIdiot:
Are you still looking at this? Do you need some help with it?

LAGIdiot:
Closing this as stale

Issue #1268, opened 2020-10-11
[BUG] Machine controllers display too high of redstone level


ALongStringOfNumbers:
**Describe the bug**
The machine controller minimum redstone strength slide bar goes from 1 to 16, despite vanilla redstone going from 0 to 15

**Versions**
Forge: 
GTCE: 1.10.1.557
_Modpack:_ None
_Addons:_ None

**Setup**
Playing Solo/Multiplayer/Both: Solo, only tested solo
_New world generated_ if applicable: yes

**Steps To Reproduce**
1) Place a machine and then add a machine controller as a cover
2) Acess the machine controller gui and see that the slider goes from 1 to 16

**Expected behavior**
I am unsure if the machine controller should have a signal strength of 0 as an option, as that does not make sense. So starting at 1 is fine for the machine controller, but it should be capped at 15 like the maximum vanilla redstone strength.

**Screenshots**
![level](https://user-images.githubusercontent.com/31759736/95667460-61eee200-0b1b-11eb-9547-68ad9f729b91.PNG)

LAGIdiot:
**Version:**
Forge: 14.23.5.2847
GTCE: 1.10.2 (dev build on f49c3f6427eb9b02a4797c31aa16bbfa875a7d2f)

**Enviroment:**
GTCE dev

**Outcome:**
Bug confirmed

**Implementation notes:**
Set maximum of `SliderWidge` to 15.
Set value loaded from NBT to 15 if it is above it.

PrototypeTrousers:
boolean shouldAllowWorking = getRedstoneSignalInput() < minRedstoneStrength;
if(isInverted) shouldAllowWorking = !shouldAllowWorking;

shouldnt this be <= ?
as it is setting the maximum of the slider to 15 will make the machine controller either on or off permanently.

ALongStringOfNumbers:
Hm, having to change the < to a <= will change the current behavior of the machine controllers, by decreasing required redstone levels by one. However, not changing the comparison will lead to maximum redstone level of 15 being ignored. This might have been the reason that the machine controllers originally went to a level of 16.

Issue #1161, opened 2020-07-27
Machines can ignore existing machine controller settings


ALongStringOfNumbers:
GTCE version: 1.9.2.515

Machines can ignore existing redstone control via a machine controller when another machine controller is added to the machine.

To Reproduce:
- Have a machine, powered, and forced into a not working state via a machine controller (This was done via a lever next to machine controller in testing, and a default setting machine controller was used)
- Add items to the input of the machine and see that it does not work.
- Add another machine controller on a different face than the one with a current machine controller, and do not provide it with a redstone signal
- Observe that the machine will turn on, and ignore the redstone control of the first machine controller, even though that controller was supposed to be forcing it off
- Remove the second machine controller, and observe that the machine is still working, even though there is only 1 machine controller left, which should be forcing the machine into a not working state
- This situation will continue until you toggle the first redstone signal (which was a lever) on and back off again

LAGIdiot:
**Version:**
Forge: 14.23.5.2847
GTCE: 1.9.3 (dev build on 5780bfd7aabc56c7fa4039ce29ee726fa099fb08)

**Enviroment:**
GTCE dev

**Outcome:**
Bug confirmed

PR #1299, opened 2020-11-06
[Bugfix] : Machine controllers wrong display/config ignored

galyfray:
**What:**
Fixed Machine controller slide bar not showing correct bounds
Fixed Machine controller setting being ignored if multiple controller control the same machine.

**How solved:**
Set maximum of `SliderWidge` to 15 and minimum to 0.
Set value loaded from NBT to 15 if it is above it.
Added two methods `shouldAllowWorking()` which is just moved code from `updateRedstoneStatus` and `isOtherAllowingWork()` for checking if other controllers will allow the machine to work.

**Outcome:**
Fixes : #1161 
Fixes : #1268 

**Additional info:**
Machine controller ui

Possible compatibility issue with older setup if the controller was set to redstone level 16, briefly tested with other level no compatibility issue found.

### DIFF
--- a/src/main/java/gregtech/common/covers/CoverMachineController.java
+++ b/src/main/java/gregtech/common/covers/CoverMachineController.java
@@ -182,17 +182,30 @@ public class CoverMachineController extends CoverBehavior implements CoverWithUI
     private void resetCurrentControllable() {
         IControllable controllable = getControllable();
         if(controllable != null) {
-            controllable.setWorkingEnabled(true);
+            controllable.setWorkingEnabled(isOtherAllowingWork());
         }
     }
 
     private void updateRedstoneStatus() {
-        boolean shouldAllowWorking = getRedstoneSignalInput() < minRedstoneStrength;
-        if(isInverted) shouldAllowWorking = !shouldAllowWorking;
         IControllable controllable = getControllable();
         if(controllable != null) {
-            controllable.setWorkingEnabled(shouldAllowWorking);
+            controllable.setWorkingEnabled(shouldAllowWorking() && isOtherAllowingWork());
         }
+    }
+
+    private boolean shouldAllowWorking(){
+        boolean shouldAllowWorking = getRedstoneSignalInput() < minRedstoneStrength;
+        return  isInverted != shouldAllowWorking ;// equivalent to isInverted ? !shouldAllowWorking: shouldAllowWorking
+    }
+
+    private boolean isOtherAllowingWork(){
+        boolean otherAllow = true;
+        for (EnumFacing side : EnumFacing.values()){
+            if(side != attachedSide && coverHolder.getCoverAtSide(side) instanceof  CoverMachineController ){
+                otherAllow = otherAllow && ((CoverMachineController)coverHolder.getCoverAtSide(side)).shouldAllowWorking();
+            }
+        }
+        return otherAllow;
     }
 
     @Override
--- a/src/main/java/gregtech/common/covers/CoverMachineController.java
+++ b/src/main/java/gregtech/common/covers/CoverMachineController.java
@@ -223,7 +223,7 @@ public class CoverMachineController extends CoverBehavior implements CoverWithUI
     public void readFromNBT(NBTTagCompound tagCompound) {
         super.readFromNBT(tagCompound);
         this.minRedstoneStrength = tagCompound.getInteger("MinRedstoneStrength");
-        if (minRedstoneStrength > 15) minRedstoneStrength = 15;
+        if (minRedstoneStrength > 15) this.minRedstoneStrength = 15;
         this.isInverted = tagCompound.getBoolean("Inverted");
         this.controllerMode = ControllerMode.values()[tagCompound.getInteger("ControllerMode")];
     }
--- a/src/main/java/gregtech/common/covers/CoverMachineController.java
+++ b/src/main/java/gregtech/common/covers/CoverMachineController.java
@@ -14,7 +14,6 @@ import gregtech.api.gui.GuiTextures;
 import gregtech.api.gui.ModularUI;
 import gregtech.api.gui.widgets.*;
 import gregtech.api.render.Textures;
-import gregtech.api.util.GTLog;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.EntityPlayerMP;
 import net.minecraft.item.ItemStack;
@@ -34,7 +33,7 @@ public class CoverMachineController extends CoverBehavior implements CoverWithUI
 
     public CoverMachineController(ICoverable coverHolder, EnumFacing attachedSide) {
         super(coverHolder, attachedSide);
-        this.minRedstoneStrength = 0;
+        this.minRedstoneStrength = 1;
         this.isInverted = false;
         this.controllerMode = ControllerMode.MACHINE;
     }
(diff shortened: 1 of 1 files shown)


LAGIdiot:
Updated PR massage to correctly trigger linking of issues

src/main/java/gregtech/common/covers/CoverMachineController.java
```diff
@@ -182,17 +183,30 @@ private IControllable getControllable() {
...
+
+    private boolean isOtherAllowingWork(){
+        boolean otherAllow = true;
+        for (EnumFacing side : EnumFacing.values()){
+            if(side != attachedSide && coverHolder.getCoverAtSide(side) instanceof  CoverMachineController ){
+                otherAllow = otherAllow && ((CoverMachineController)coverHolder.getCoverAtSide(side)).shouldAllowWorking();
+            }
         }
+        return otherAllow;
     }
```
LAGIdiot, line 210: You need also to check in which mode does controller work as they may not be controlling same thing. For example one is controlling machine and other one Pump Cover.

src/main/java/gregtech/common/covers/CoverMachineController.java
```diff
@@ -182,17 +183,30 @@ private IControllable getControllable() {
...
         IControllable controllable = getControllable();
         if(controllable != null) {
-            controllable.setWorkingEnabled(shouldAllowWorking);
+            controllable.setWorkingEnabled(shouldAllowWorking() && isOtherAllowingWork());
+        }
+    }
+
+    private boolean shouldAllowWorking(){
+        boolean shouldAllowWorking = getRedstoneSignalInput() < minRedstoneStrength;
+        return  isInverted != shouldAllowWorking ;// equivalent to isInverted ? !shouldAllowWorking: shouldAllowWorking
```
LAGIdiot, line 199: This constructs is way too hard to digest. When there must be comment to explain boolean logic than that logic is over-complicated and may lead to errors. I know IDE suggested but I won't allow it.

```suggestion
        //noinspection SimplifiableConditionalExpression
        return isInverted ? !shouldAllowWorking : shouldAllowWorking;
```

src/main/java/gregtech/common/covers/CoverMachineController.java
```diff
@@ -182,17 +183,30 @@ private IControllable getControllable() {
...
+    }
+
+    private boolean isOtherAllowingWork(){
```
LAGIdiot, line 202: Naming of this function does not sound right to me, maybe better naming would be `doesOtherAllowWorking`

src/main/java/gregtech/common/covers/CoverMachineController.java
```diff
@@ -182,17 +183,30 @@ private IControllable getControllable() {
...
-        if(isInverted) shouldAllowWorking = !shouldAllowWorking;
         IControllable controllable = getControllable();
         if(controllable != null) {
-            controllable.setWorkingEnabled(shouldAllowWorking);
+            controllable.setWorkingEnabled(shouldAllowWorking() && isOtherAllowingWork());
+        }
+    }
+
+    private boolean shouldAllowWorking(){
+        boolean shouldAllowWorking = getRedstoneSignalInput() < minRedstoneStrength;
```
LAGIdiot, line 198: This does not work correctly if both `minRedstoneStrength` and `getRedstoneSignalInput()` are set to 15. But I am not sure how to handle this yet. I may debate this on Discord.

LAGIdiot (changes requested):
Thanks for creating this PR. I have found some areas that requires some adjustments so please see to them.

src/main/java/gregtech/common/covers/CoverMachineController.java
galyfray: if both are set to 15 then the controller don't work if we use `<=` instead of `<` then the machine will always work even when directly putting a lever on it.

src/main/java/gregtech/common/covers/CoverMachineController.java
galyfray: yep wasn't nice for me too I will replace

galyfray:
changes done thank you for your feedback !

src/main/java/gregtech/common/covers/CoverMachineController.java
```diff
@@ -209,6 +223,7 @@ public void readFromNBT(NBTTagCompound tagCompound) {
         this.minRedstoneStrength = tagCompound.getInteger("MinRedstoneStrength");
         this.isInverted = tagCompound.getBoolean("Inverted");
         this.controllerMode = ControllerMode.values()[tagCompound.getInteger("ControllerMode")];
+        if (minRedstoneStrength > 15) minRedstoneStrength = 15;
```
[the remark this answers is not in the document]
LAGIdiot: This one does not seems resolved to me and I still require this change.

src/main/java/gregtech/common/covers/CoverMachineController.java
LAGIdiot: That would be correct statement but in this case it would lower requirements for Redstone signal by one level. Let's say from 8 to 7 which may break someone builds. And we always try to maintain high level of backward compatibility (not always possible). So in this case we need to add some special logic which will handle 15 case (and properly documented it - yes we have many hacks without documentation but I won't all any new ones missing documentation during my reign) and let's everything else work as before.

src/main/java/gregtech/common/covers/CoverMachineController.java
```diff
@@ -115,7 +115,7 @@ public ModularUI createUI(EntityPlayer player) {
         updateDisplayInventory();
         return ModularUI.defaultBuilder()
             .label(10, 5, "cover.machine_controller.name")
-            .widget(new SliderWidget("cover.machine_controller.redstone", 10, 20, 156, 20, 1.0f, 16.0f,
+            .widget(new SliderWidget("cover.machine_controller.redstone", 10, 20, 156, 20, 1f, 15.0f,
```
LAGIdiot, line 118: There is unnecessary change regarding 1.0f -> 1f it will not have any effect on code but for consistency sake and for making PRs as small as possible (so it's more readable).

LAGIdiot (changes requested):
Thanks for your hard work on this PR. I have added some comments to specific section that are not up to speed yet. And I have one overall comment, some sections of code seems not properly formatted so please run IDE formatter on all files that you changed before committing them.

I may be requiring quite a lot from you but I need to maintain code quality, readability and maintainability.

src/main/java/gregtech/common/covers/CoverMachineController.java
galyfray: as you can see here :
 https://github.com/GregTechCE/GregTech/blob/59cf924550f97635e1f576bf68309adcb8bface9/src/main/java/gregtech/common/covers/CoverMachineController.java#L190
my PR should not make any breaking change by reducing redstone requirement.
I understand that maintaining overall code quality is important don't worry I'll did any required changes until it become perfect.

src/main/java/gregtech/common/covers/CoverMachineController.java
galyfray: as you can see here :
 https://github.com/GregTechCE/GregTech/blob/59cf924550f97635e1f576bf68309adcb8bface9/src/main/java/gregtech/common/covers/CoverMachineController.java#L190
my PR should not make any breaking change by reducing redstone requirement.
I understand that maintaining overall code quality is important don't worry I'll did any required changes until it become perfect.

LAGIdiot: Yeah you are right I had that logic inverted in my head because this concept prevents working by default not allows it. My bad. This is correct implementation.

src/main/java/gregtech/common/covers/CoverMachineController.java
```diff
@@ -207,6 +223,7 @@ public void writeToNBT(NBTTagCompound tagCompound) {
     public void readFromNBT(NBTTagCompound tagCompound) {
         super.readFromNBT(tagCompound);
         this.minRedstoneStrength = tagCompound.getInteger("MinRedstoneStrength");
+        if (minRedstoneStrength > 15) minRedstoneStrength = 15;
```
LAGIdiot, line 226: Missing this...
```suggestion
        if (minRedstoneStrength > 15) this.minRedstoneStrength = 15;
```

LAGIdiot (approved):
Thanks for your hard work on this PR. I have added two comments and will merge it with small fix,

Issue #1289, opened 2020-10-31
[BUG] Crash-to-desktop on loading world with Gtech pumps exporting into ImTech pipes.


CelestialPhoenix:
**Describe the bug**
When using a Gregtech pump cover to export fluids into Immersive Technology pipes the game crashes to desktop when re-opening that world.

This specifically happens with the pump cover exporting into the ImTech pipe. 
Using the machine's auto-export works perfectly fine, and no issues with other mod machines when testing the full pack.
Pumping into other mod pipes does not cause a crash.

**Versions**
minecraft            | 1.12.2       | minecraft.jar              
mcp                  | 9.42         | minecraft.jar     
FML                  | 8.0.99.99    | forge-1.12.2-14.23.5.2854-universal.jar 
 forge                | 14.23.5.2854 | forge-1.12.2-14.23.5.2854-universal.jar     
 codechickenlib       | 3.2.3.358    | CodeChickenLib-1.12.2-3.2.3.358-universal.jar 
 jei                  | 4.16.1.302   | jei_1.12.2-4.16.1.302.jar       
 gregtech             | 1.10.3.569   | gregtech-1.12.2-1.10.3.569.jar   
 immersiveengineering | 0.12-98      | ImmersiveEngineering-0.12-98.jar   
(issue isolated down to the above list, full pack list [available here](https://www.curseforge.com/minecraft/modpacks/blood-sweat-and-gears))

**Setup**
Singleplayer

**Steps To Reproduce**
Place a Gtech pump cover on the side of a tank/machine, and set it to export.
Connect a ImEng Fluid Pipe to the pump, and add some fluid to the system.
Close and reload the game, then log back into that world.

**Other notes**
Lost access to my Mojang login before I could verify the issue on latest Gtech update. Changelog does not mention pumps- so its likely to effect the latest version too.
I asked on the Gtech discord if anyone could verify it as was asked to drop it here.

PrototypeTrousers:
related to #1263

LAGIdiot:
Thanks for reporting this issue. This does look like it is related to #1263 and investigation will continue there. I will leave this one open so w can retest solution for this too and for case that we would need additional info.

Issue #1263, opened 2020-10-05
[BUG] Crash with Covers when TileEntities lose their capabilities


PrototypeTrousers:
**Describe the bug**
 Placing two Gregtech tanks ( not quantum tanks ) with pumps facing each other in import mode will crash the game on world load.

**Versions**
Mods:
CodeChickenLib-1.12.2-3.2.3.358-universal
gregtech-1.12.2-1.10.2.566
jei_1.12.2-4.16.1.301
MouseTweaks-2.10-mc1.12.2

**Setup**
Playing Solo
New world generated

**Steps To Reproduce**
1)Place 2 tanks adjacent to each other
2)Place pumps on the tanks, facing each other
3)set them to import
4)Save and quit
5)try to load the world

**Expected behavior**
Placing 2 tanks touching will form a multiblock
Placing one pump will visually show the multiblock not forming anymore.
Placing the second one will work as usual.
Setting the pumps to import
Save and quit
Restart world. Crash.

Crash Report
https://pastebin.com/t2ERFTDJ

**Screenshots**
tanks placed down
![1](https://user-images.githubusercontent.com/11739607/95140036-431bd480-0744-11eb-8404-b494c036b1cd.jpg)
pump installed on right tank
![2](https://user-images.githubusercontent.com/11739607/95140037-43b46b00-0744-11eb-9202-6b365acc602d.jpg)
placing water will fill both tanks
![3](https://user-images.githubusercontent.com/11739607/95140039-444d0180-0744-11eb-9974-bc7112a32d9d.jpg)
saving and reloading will show the multiblock formed again
![4](https://user-images.githubusercontent.com/11739607/95140038-444d0180-0744-11eb-91de-4ffd8db60b07.jpg)
installed pump in the left tank, set as export
![5](https://user-images.githubusercontent.com/11739607/95140040-444d0180-0744-11eb-807f-0b625e735e43.jpg)
reload world ( if theres still water on both during the save it will crash on load too
![6](https://user-images.githubusercontent.com/11739607/95140041-44e59800-0744-11eb-8462-d45de8be1a98.jpg)
setting both pump to import will crash the world on load

PrototypeTrousers:
broke between gregtech-1.12.2-1.8.8.440 and gregtech-1.12.2-1.8.9.457 so i guess since multiblock tanks is a thing

LAGIdiot:
**Version:**
GTCE: 1.10.2

**Outcome:**
[...]

PR #1303, opened 2020-11-08
Fix Null pointer exception on covers

PrototypeTrousers:
**What:**
This PR aims to solve issues #1289 and #1263

Issue is due to missing capability expected by the covers installed on multiblock tanks and machines.
On multiblock tanks the issue comes from the fact that on tick 0 the multiblock tank doesnt have a valid tank, so it cant have fluid capability, leading to a crash when the pump cover tries to get the tank original capability and gets NULL from it.
On machines the same crash can occur with pumps AND conveyors on specific cases by taking away a expected capability due to how OutputSide can change these capabilities.

**How solved:**
Initializing the multiblock tank FluidInventory, just like QuantumTanks get them initialized.
Covers that may need a specific machine capability may change the outputside to allowInputFromOutputside, making sure no capability get cut away due to how that side works.

Also added extra checks to multiblock tanks as to retain their configurations to behave as expected when having blocked sides blocked.

The NULL crash can be avoided by just checking for defaultValue of null on both pumps and conveyors, but this PR aims to solve the cause of that null value being passed improperly. Not having the proper capabilities exposed at the expected time may lead to pipes/cables disconnecting.

**Outcome:**
World works on reload.
NPE doesnt occur.

**Additional info:**
image
Tanks on the left are separated by shutter covers. tanks on the bottom separated by pump covers.
will retain their shapes and inventories on reload, without crashing

This PR will not fix worlds that are already crashing due to OutputSide not exposing the capability expected by the cover installed.

any comments are appreciated.


src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityTank.java
```diff
@@ -695,6 +702,9 @@ public NBTTagCompound writeToNBT(NBTTagCompound data) {
     @Override
     public void readFromNBT(NBTTagCompound data) {
         super.readFromNBT(data);
+        if (data.hasKey("blockedSides")) {
+            this.blockedSides = data.getInteger("blockedSides");
+        }
```
LAGIdiot, line 707: I think this could be handled by more simple solution. Because `blockedSides` is information that can be found just from this block alone and it's computation is not that expensive so I would rather calculate it on load then save it to NBT

LAGIdiot (changes requested):
Thanks for your hard work on this PR. But I think you are trying to mix two things together here which in this case does not blend together that well.

I would like you to split this one to two different PRs. One which will handle multi tank formation to which I added one comment. And second handling loss of capabilities. For which I have this:
If machine does not provide compatibility on some side we have to respect that and it should cover even loss of given compatibility. Let's take for example Brewery and Logistic Pipes. As Brewery does not have output slot it does not have `ITEM_HANDLER_CAPABILITY` on its output side. Which we respect by not allowing placing Conveyor Cover on it and others should too. By clicking it with screwdriver we will allow input which means `ITEM_HANDLER_CAPABILITY` will be available which will in end allow placing of Conveyor cover on it.

So to handle this correctly we need to do 2 things:
1) Correctly handle `defaultvalue` being null in all `getCapability` for covers which means proper checks before constructing these delegates and throwing argument null exception if someone constructs them with null.
2) Inform Cover GUI that cover is missing "required" capability which prevents him from working so player can have some clue for why it does not work - this one may be quite tricky

PrototypeTrousers:
Currently we have three indicators, we can add more and make even more verbose or show only when the cover wont work.
Screenshot 2020-11-16 011754
Screenshot 2020-11-16 011816
Screenshot 2020-11-16 011840
I know i can do better on the indicators but have to know if they are needed in the first place.
widget is synced from the server, so shows current status dynamically.

LAGIdiot:
Honestly I was not expecting you to implement it right away as I did not even finished requirements for it. This will be part of public API so we can't rush it out because reworking it later would be pain. Also I want this to be usable on all GUIs that GTCE (and addons) have so it needs to be quite versatile and easy to use. With as little boiler plaiting on use side as possible. So let me get some analysis on this matter. Suggesting are welcome and this may even change to discussion for quite while.

**State types we want to report:**
Every state should have it's own icon which should be distinct and recognizable on first sight.
- Working (everything is going as expected - recipe is being work on, pump is pumping, ...)
- Idle (could work if there was something to work on - no recipe selected, nothing to pump, ...)
- Warning (situation is happening which may lead to problem or is preventing work - recipe is using more energy then machine is getting, recipe can't be processed because outputs are full, ...)
- Problem (something is not working as expected - recipe stopping and progress being reset because of missing energy, steam can't exhaust, missing capability, ...)
_Maybe warning could even be split to two soft and hard_

**Situations we want to report:**
This has to be most forgiving so GTCE or addons can add new ones when need arise _or remove them without breaking anything_.
We should use `GTControlledRegistry` for this and have publicly available list of know types, so addons could use them and even add it's own.
List to be build when needed.

Regarding use side there should be just one function which will return widget on specified position, and one that will take Situation and feed it to widget, no situation state caching - all this needs to be in widget itself.

PrototypeTrousers:
I struggled a lot with the registry setup so could really use some tips here.
Basically functions the same as before, but the widget gets from the registry the additional info.

LAGIdiot:
First of all good work! You are really putting up good fight with my requests.

Regarding registry setup. I would do it in same way as we have `StoneTypes`. Two classes one which contains our Situations registrations and one which is Situation. Situation should handle creation and whole registration process if possible in one go. Constructor should get id, name (just short name not full one which you will need for lang files), and state type as enum value (which will be in widget translated to textures). There is no need for holding counter it will be up to registrant not to register with same ID.

Also passing only id to widget seems bit un-intuitive (for synchronization purposes it is required). Regarding used icons and current texts I would like @pyure to say something as he is more UX person them me. There are also some formatting errors but I believe that you will fix them before posting it for verification.

LAGIdiot:
Also please update name and description of this issue

src/main/java/gregtech/api/Situation.java
```diff
@@ -0,0 +1,41 @@
+package gregtech.api;
+
+import gregtech.api.util.GTControlledRegistry;
+
+public class Situation {
+
+    public static GTControlledRegistry<String, Situation> SITUATION_REGISTRY = new GTControlledRegistry<>(Short.MAX_VALUE);
+    private final String situationName;
+    public Enum situationType;
+    public String situationLocaleName;
```
LAGIdiot, line 10: For situation name we should recommend using snake_case and provided name directly use in lang files. This will need some prefixing but it should save us need for another param

src/main/java/gregtech/api/gui/widgets/SituationWidget.java
```diff
@@ -0,0 +1,100 @@
...
+    private final IntSupplier currentSituationId;
+    protected String tooltipHoverString;
+    protected long hoverStartTime = -1L;
+    protected boolean isMouseHovered;
+    protected int currentError;
+
+    protected TextureArea area;
+    private boolean isVisible = true;
+
+    public SituationWidget(int xPosition, int yPosition, int width, int height, IntSupplier getSituationId) {
```
LAGIdiot, line 32: Also there is still lingering my previous comment regarding use `IntSuplier` here

LAGIdiot (changes requested):
I have to say great job on this PR and keeping with my request. There is another bunch of them.

src/main/java/gregtech/api/gui/widgets/SituationWidget.java
```diff
@@ -0,0 +1,100 @@
...
+        } else if (!isHovered && wasHovered) {
+            this.isMouseHovered = false;
+            this.hoverStartTime = 0L;
+        } else if (isHovered) {
+            long timeSinceHover = System.currentTimeMillis() - hoverStartTime;
+            setTooltipHoverString();
+            if (timeSinceHover > 1000L && tooltipHoverString != null) {
+                drawHoveringText(ItemStack.EMPTY, Collections.singletonList(tooltipHoverString), 300, mouseX, mouseY);
+            }
+        }
```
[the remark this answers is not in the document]
PrototypeTrousers: most of the code was based on the cycle button, yes. looks like we can move it up to the widget class since we have 2 different widgets using the same code

src/main/java/gregtech/api/gui/widgets/SituationWidget.java
PrototypeTrousers: i couldn't find any other way to sync the widget properly while its open without using it here.
setting it to int will only display the changes when the GUI is reopened.

src/main/java/gregtech/api/gui/widgets/SituationWidget.java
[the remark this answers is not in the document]
LAGIdiot: Good work extracting it out

src/main/java/gregtech/api/gui/widgets/SituationWidget.java
LAGIdiot: For synchronization you will still need that int id but you can get it from Situation it self

src/main/java/gregtech/api/cover/CoverBehavior.java
```diff
@@ -209,6 +214,20 @@ public boolean isRemote() {
...
+
+    public void setSituation(Situation situation) {
+        this.situation = situation;
+    }
+
+    public void update(){
+        if (!isWorkingAllowed) {
+            setSituation(DISABLED_BY_CONTROLLER);
+        }
+    }
```
LAGIdiot, line 229: I am not sure about this. All covers have to handle `isWorkingAllowed` on their own so there is probably no need for this.

src/main/java/gregtech/api/gui/widgets/SituationWidget.java
```diff
@@ -0,0 +1,94 @@
...
+    public void setTooltipHoverString() {
+        this.tooltipHoverString = I18n.format(this.currentSituation.situationLocaleName);
+    }
+
+    public SituationWidget setImage() {
+        SituationTypes iconTextures = this.currentSituation.situationTypes;
+        switch (iconTextures) {
+            case IDLE:
+                this.area = GuiTextures.STATUS_IDLING;
+                return this;
```
LAGIdiot, line 44: I think you should return only once at end

src/main/java/gregtech/api/situation/Situation.java
```diff
@@ -0,0 +1,38 @@
+package gregtech.api.situation;
+
+import gregtech.api.util.GTControlledRegistry;
+
+import static gregtech.api.GTValues.MODID;
+
+public class Situation {
+
+    public static GTControlledRegistry<String, Situation> SITUATION_REGISTRY = new GTControlledRegistry<>(Short.MAX_VALUE);
```
LAGIdiot, line 9: This needs to be final so anyone can mess with it.

src/main/java/gregtech/api/situation/Situation.java
```diff
@@ -0,0 +1,38 @@
...
+    public final String situationName;
+    public final SituationTypes situationTypes;
+    public final String situationLocaleName;
+    public int id;
```
LAGIdiot, line 13: This one should be final too

src/main/java/gregtech/api/situation/Situation.java
```diff
@@ -0,0 +1,38 @@
...
+    
+    public Situation(int id, String situationName, SituationTypes situationTypes) {
+        this.id = id;
+        this.situationName = situationName;
+        this.situationTypes = situationTypes;
+        this.situationLocaleName = MODID + ".situation." + this.situationName;
```
LAGIdiot, line 19: I don't think that `MODID` should be used here as it will be inconsistent with other our lang entries also it will always be our no matter who adds that Situation

src/main/java/gregtech/api/situation/Situation.java
```diff
@@ -0,0 +1,38 @@
...
+        SITUATION_REGISTRY.register(id, situationName, situation);
+    }
+
+    public static Situation getSituationFromId(int id) {
+        return SITUATION_REGISTRY.getObjectById(id);
+    }
+
+    public static int getIdFromSituation(String situationName) {
+        return SITUATION_REGISTRY.getIdByObjectName(situationName);
+    }
```
LAGIdiot, line 33: I don't think you need this anymore

src/main/java/gregtech/api/situation/Situation.java
```diff
@@ -0,0 +1,38 @@
...
+        this.id = id;
+        this.situationName = situationName;
+        this.situationTypes = situationTypes;
+        this.situationLocaleName = MODID + ".situation." + this.situationName;
+        registerSituation(id,situationName,this);
+    }
+
+    public static void registerSituation(int id, String situationName, Situation situation){
+        SITUATION_REGISTRY.register(id, situationName, situation);
+    }
```
LAGIdiot, line 25: This one should not be public and static as registration should be available only from constructor

src/main/java/gregtech/api/situation/Situation.java
```diff
@@ -0,0 +1,38 @@
...
+    public static Situation getSituationFromId(int id) {
+        return SITUATION_REGISTRY.getObjectById(id);
+    }
+
+    public static int getIdFromSituation(String situationName) {
+        return SITUATION_REGISTRY.getIdByObjectName(situationName);
+    }
+
+    public static String getLocaleName(Situation situation){
+        return situation.situationLocaleName;
```
LAGIdiot, line 36: This one should be just standard getter

LAGIdiot (commented):
I am looking that you are doing some good progress on this. I have added some comments you should take look at.

src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
```diff
@@ -140,25 +153,26 @@ protected void updateRecipeProgress() {
...
+                    this.previousRecipe = this.currentRecipe;
                 }
             }
         }
-        if (currentRecipe != null && setupAndConsumeRecipeInputs(currentRecipe)) {
-            setupRecipe(currentRecipe);
+        if (this.currentRecipe != null && setupAndConsumeRecipeInputs(this.currentRecipe)) {
+            setupRecipe(this.currentRecipe);
+        } else {
+            metaTileEntity.setSituation(OUTPUT_INVENTORY_FULL);
```
ALongStringOfNumbers, line 175: As a note, the `trySearchNewRecipe` method is overridden in `MetaTileEntityMultiFurnace`, so it would not benefit from this output full situation being set here. The situation being set should also be added to `MetaTileEntityMultiFurnace`.

LAGIdiot:
This looks good most of things I mentioned seems to be resolved. We may need to extend coverage of this to cases like one mentioned by @ALongStringOfNumbers.

I would like to see this in next minor version update. But for that we need extensive testing given it's size and integration. Which may be little problematic given my broken dev environment. But I will get to it.

Anyway thank you for you hard work.

PrototypeTrousers:
Single blocks to do:
Steam
- [ ] Boilers
- [ ] Forge Hammer texture overlapping
- [ ] Turbines
Electric
- [ ] Pump
- [ ] Item collector
- [ ] Diesel generator
- [ ] Gas turbines
Multiblocks to do : everything
- [ ] Query multiblock parts situation along with controller situation
- [ ] Fit situation type and descriptions to multiblock screen

LAGIdiot:
I was finally able to take a look at this. Let's start with thank you for not abandoning this in my "absence".

I quickly looked through code and it looks solid, for now I will not do review on it as it needs more work. I think that there was some bad merging done as there are issues with lang file. As some PR touching it which are in master are reverted here (for example numeric formatting and TOP integrations).

I see that you created list of machines which are missing this mechanism which is nice. We may also implement this for Battery buffer, Rock breaker and maybe some other on top of your list.

From quick testing I have stumble onto bug with fisher using missing lang entry when water check fails.

PR #1307, opened 2020-11-10
Fixed a missing command warning and a mistake in zh_cn.lang

ChromaPIE:
**What:**
Same as the title.

**How solved:**
Editting the three .lang files.

**Outcome:**
Fixed zh_cn Plant Ball localizaiton, added warning sentence for `/gt util hand` with empty mainhand


LAGIdiot (approved):
Thanks for your hard work. I have confirmed entries in English. And the rest I will accept in good faith that they are correct as I am not able to confirm it.

PR #1310, opened 2020-11-12
Integrate the Chemical Bath into the ore byproduct page. Remove extra…

### LINKED ISSUE #1032: [Bug&Question] OreByproduct Page
## Description
- Chemical Bath should be added for recipe catalyst.
- These ores are missing tooltip for processing with Chemical Bath. (It seems only 4th byproduct slot have description for Chemical Bath.)
  - Silver
  - Gold
  - Platinum
  - Cobalt
  - Copper
  - Osmium
  - Sheldonite
  - Tetrahedrite
  - Cobaltite
  - Nickel
  - Zinc
- Redstone and Cinnabar have 4th byproduct slot, but they say "Byproduct not obtainable". What does this actually mean? It's confusing.
- Redstone has 5th byproduct slot with no description. (And Pyrite cannot be obtained by processing Redstone ore.)
image

## Version
gregtech-1.12.2-1.9.0.481.jar

ALongStringOfNumbers:
… byproduct additions from some materials.

**What:**
This PR aims for better integration of the Chemical Bath for the Ore Byproduct page, solving several issues brought up in #1032.
For further clarification of the ore byproduct page, some materials with extraneous ore byproduct additions were removed.

**How solved:**
Adds the Chemical Bath as a catalyst to the ore byproduct page.
Adds the Chemical Bath tooltip to byproducts in all positions (if applicable) instead of just to the 4th position. This is done simply because the byproducts for the Chemical Bath are always the last byproduct in the list.
Remove duplicate byproduct additions for redstone and cinnabar, as they were causing in correct information to appear on the ore byproduct page, as they extra byproducts were not actually available from any machine.

**Outcome:**
Better integrates the Chemical Bath into the Ore Byproduct page, and removes some extraneous, non-applicable byproducts. Closes #1032.

**Additional info:**
An example of the updated tooltips for the chemical bath. This specific one comes from silver ore, which has sulfur as its second, and last byproduct

cbath

### DIFF
--- a/src/main/java/gregtech/api/unification/material/Materials.java
+++ b/src/main/java/gregtech/api/unification/material/Materials.java
@@ -712,8 +712,6 @@ public class Materials {
         Sodalite.addOreByProducts(Lazurite, Lapis);
         Spodumene.addOreByProducts(Aluminium, Lithium);
         Ruby.addOreByProducts(Chrome, GarnetRed);
-        Redstone.addOreByProducts(Cinnabar, Pyrite);
-        Cinnabar.addOreByProducts(Ruby);
         Phosphor.addOreByProducts(Apatite, Phosphate);
         Iridium.addOreByProducts(Platinum, Osmium);
         Pyrope.addOreByProducts(GarnetRed, Magnesium);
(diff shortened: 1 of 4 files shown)


LAGIdiot (approved):

I have checked that removed by products were really unobtainable and probably added by mistake so there will be no incompatibility.

PR #1311, opened 2020-11-15
properly initialize metaTileEntityTank fluid inventory

PrototypeTrousers:
**What:**
Suggested Fix for #1289 and #1263
Separated from PR #1303
those issues happen due to metaTileEntityTank fluid inventory not being properly initialised on load.

**How solved:**
calling this.fluidInventory = getActualFluidTank();
instead of initializeInventory();

instead of initializeInventory() is a metaTileEntity method that initialize fluidinventory as
`this.fluidInventory = new FluidHandlerProxy(importFluids, exportFluids);`
but we have neither `importFluids` nor `exportFluids` on meta tile entities tanks, leading
to `getTankProperties().length` being 0, so when the tank is asked by the cover for its capability, null is returned.
Since the pump need the capability to be non null, it crashes

**Outcome:**
World loading normally

### DIFF
--- a/src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityTank.java
+++ b/src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityTank.java
@@ -98,7 +98,7 @@ public class MetaTileEntityTank extends MetaTileEntity implements IFastRenderMet
                 MetaTileEntityTank.this.onFluidChangedInternal();
             }
         };
-        initializeInventory();
+        this.fluidInventory = getActualFluidTank();
     }
 
     @Override


LAGIdiot:
Thank you for your hard work. But this is just partial copy of functionality from #1303 right? As that one have even proper tank forming on load when cover was placed on it.

PrototypeTrousers:
Yes, this fixes ONLY the crash, and as its world breaking i thought this part could go first.
there still multiblock issues even with the other PR.
like client-side desync and fluid voiding.
if you rather wait some more, close this PR and i will merge this one with the tank changes when its ready. Thanks :)

PrototypeTrousers:
closing in favor of #1313

PR #1313, opened 2020-11-16
Multi block tank fixes

### LINKED ISSUE #1289: [BUG] Crash-to-desktop on loading world with Gtech pumps exporting into ImTech pipes.

PrototypeTrousers:
**What:**
This changes
 a) MetaTileEntityTank fluidInventory initialization, preventing a NPE with pump cover on world load.
 b) Check for MetaTileEntityTank blockedsides on load, before trying to assemble the multiblock.
 c) Changed how MetaTileEntityTank removal from multiblock works to try to fix fluid voiding

**How solved:**
As previously discussed:
a) instead of initializeInventory(), a metaTileEntity method that initialize fluidinventory as
this.fluidInventory = new FluidHandlerProxy(importFluids, exportFluids);
but we have neither importFluids nor exportFluids on meta tile entities tanks, leading
to getTankProperties().length being 0, so when the tank is asked by the cover for its capability, null is returned.
Since the pump need the capability to be non null, it crashes.

b) call for a `recheckBlocked `sides on the `onLoad `method, so `blockedSides` is not always 0 (no sides blocked) in `canTanksConnect()`

c) `removeTankFromMultiblock` calls for `getFluidStoredInTank` to get the amount of fluid based on a ratio of amount of fluid in the multiblock and the actual MetaTileEntityTank that contains the fluid.
i have split the method for this so we could get the fluid in the actual tank to determine how much should go to the removed tank.

**Outcome:**
a) fixes #1289 and #1263, that comes from the tank being asked for its capability before the actual valid fluid inventory exists
b) multiblocks keep their shape even after world reloads
c) this changes how tanks look after they are broken/multiblock gets invalidated, but retain the full amount of liquid, but does not look right

**Additional info:**
c)Screenshot 2020-11-16 015051
cutting horizontally with shutters will split the tank in halves, each with half of the original tank, even if the fluid looked like it was on the bottom originally:
Screenshot 2020-11-16 015123
Breaking one of the tanks will make the tanks show their real inventory. Keeps the total amount of fluid.
Screenshot 2020-11-16 015159

I know c) still needs some work but the actual issue here is getting over my head.

### DIFF
--- a/src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityTank.java
+++ b/src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityTank.java
@@ -98,7 +98,7 @@ public class MetaTileEntityTank extends MetaTileEntity implements IFastRenderMet
                 MetaTileEntityTank.this.onFluidChangedInternal();
             }
         };
-        initializeInventory();
+        this.fluidInventory = getActualFluidTank();
     }
 
     @Override
@@ -152,6 +152,7 @@ public class MetaTileEntityTank extends MetaTileEntity implements IFastRenderMet
     protected void onCoverPlacementUpdate() {
         super.onCoverPlacementUpdate();
         this.needsCoversUpdate = true;
+        setTankController(null);
     }
 
     @Override
--- a/src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityTank.java
+++ b/src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityTank.java
@@ -103,8 +103,8 @@ public class MetaTileEntityTank extends MetaTileEntity implements IFastRenderMet
 
     @Override
     public void onLoad() {
-        this.recheckBlockedSides();
         super.onLoad();
+        this.recheckBlockedSides();
     }
 
     @Override
--- a/src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityTank.java
+++ b/src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityTank.java
@@ -102,6 +102,12 @@ public class MetaTileEntityTank extends MetaTileEntity implements IFastRenderMet
     }
 
     @Override
+    public void onLoad() {
+        this.recheckBlockedSides();
+        super.onLoad();
+    }
+
+    @Override
     public void update() {
         this.fluidInventory = getActualFluidTank();
         super.update();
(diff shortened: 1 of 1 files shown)


src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityTank.java
```diff
@@ -98,7 +98,13 @@ protected void onFluidChanged(FluidStack newFluidStack, FluidStack oldFluidStack
...
             }
         };
-        initializeInventory();
+        this.fluidInventory = getActualFluidTank();
+    }
+
+    @Override
+    public void onLoad() {
+        this.recheckBlockedSides();
+        super.onLoad();
```
LAGIdiot, line 107: This should be other way around so that base is loaded first

LAGIdiot (changes requested):
Thanks for your hard work, A and B looks great there is just one small thing as pointed bellow.
Regarding C, that one is quite hard to comprehend problem and I was too once solving it but I was able only to fix it partially so I think it should be left for another PR as it will be quite time consuming and it would also require quite extensive testing before merging it.

LAGIdiot (approved):
Thanks for your hard work, approved and accepted

PR #1317, opened 2020-11-20
Ensure saw-broken ice blocks don't create water.

### LINKED ISSUE #1314: [BUG] Ice drops and also places water when broken with saw
**Describe the bug**
Title

**Versions**
Forge: 14.23.5.2854

GTCE: 1.10.5.582
CodeChickenLib: 3.2.3.358

**Setup**
New instance with only GTCE and CodeChickenLib, tested on new singleplayer world

**Steps To Reproduce**
1) Break ice with saw in survival mode

**Expected behavior**
The ice to drop and no water appearing when broken with a saw

eutropius225:
**What:**
This PR ensures that ice blocks don't melt into water when broken by a saw, but harvest like a silk-touch tool would, leaving no water behind.

**How solved:**
The presented solution is to use vanilla means to listen to block updates in the world, reverting the change if it appears to be from an ice block trying to turn into water. 
Essentially, when the block is to be harvested, this event listener is added to the world, and reverts the first attempt to set the block at that position to flowing water on that tick.

**Outcome:**
Ice blocks broken by saws no longer create water like they might if broken by other tools.
Fixes #1314

Other mods that use similar means to their ends may experience unexpected behaviours when an ice block is broken by a saw but wouldn't convert into water: other world listeners added after the ice listener may be skipped when the ice listener is removed from the list on a later tick.

### DIFF
--- a/src/main/java/gregtech/common/tools/ToolSaw.java
+++ b/src/main/java/gregtech/common/tools/ToolSaw.java
@@ -3,13 +3,20 @@ package gregtech.common.tools;
 import net.minecraft.block.material.Material;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.enchantment.Enchantment;
+import net.minecraft.entity.Entity;
 import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.init.Blocks;
 import net.minecraft.init.Items;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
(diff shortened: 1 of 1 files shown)


OneLemonyBoi (approved):
10/10

LAGIdiot (approved):
Good work on this PR. Approach seems reasonable and it does what it should. So approved and accepted.

PR #1319, opened 2020-11-23
Fix Drained Fluid Cells not stacking due to empty NBT.

ALongStringOfNumbers:
**What:**
Similar to #1127, this PR makes it so that fluid cells that are drained entirely have their empty NBT removed, so that they can stack with fluid cells with no NBT tag that have been created.

**How solved:**
Adds checks in two places to see if the fluid cells are drained all the way when they are drained, and if they are removes their tags. It was required to add this check in two places, as there are two implementations of the fluid cells, one that allows partial filling and one that does not allow partial filling.

**Outcome:**
Allows fully drained fluid cells to stack newly created fluid cells. Works for all types of fluid cells.

None expected.


LAGIdiot (approved):

PR #1326, opened 2020-11-25
Fix Fluid Canner and Electric Furnace recipe finder not working.

### LINKED ISSUE #1325: [BUG] Fluid Canner stopped operating
**Describe the bug**
After updating to 1.10.6.591, it was noticed that the fluid canner has stopped filling and draining cells and buckets

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.6.591
_Modpack:_ None
_Addons:_ None

**Setup**
Playing Solo
_New world generated_ Yes

**Steps To Reproduce**
1) Place down a powered Fluid Canner and fill the fluid input with water
2) Place both cells and buckets in the item input and notice that neither are filled with water
3) Place down another powered Fluid Canner and place a water bucket in it and see that the water bucket is not emptied

**Expected behavior**
For the Fluid Canner to correctly perform filling/draining recipes.

ALongStringOfNumbers:
**What:**
Fixes the Fluid Canner and Electric Furnace recipe maps not working, as they did not override the correct findRecipe method after the addition of MatchingMode

**How solved:**
Added the MatchingMode parameter so that the correct findRecipe method was being overwritten.

**Outcome:**
Fixed Fluid Canner and Electric Furnace not functioning. Closes #1325.

**Additional info:**
These were the two machines that I could find that did not work after the update, after some brief machine testing. 

Possibly, if some addon was using the findRecipe method of one of the changed recipe maps. I can change this if it determined that it needs to be changed.

### DIFF
--- a/src/main/java/gregtech/api/recipes/machines/RecipeMapFluidCanner.java
+++ b/src/main/java/gregtech/api/recipes/machines/RecipeMapFluidCanner.java
@@ -1,6 +1,7 @@
 package gregtech.api.recipes.machines;
 
 import gregtech.api.recipes.CountableIngredient;
+import gregtech.api.recipes.MatchingMode;
 import gregtech.api.recipes.Recipe;
 import gregtech.api.recipes.RecipeMap;
 import gregtech.api.recipes.builders.SimpleRecipeBuilder;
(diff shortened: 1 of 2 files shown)


Exaxxion:
Can confirm these changes restored the functionality of the fluid canning machine.

canning_fixed

LAGIdiot:
Thanks for finding and fixing this issue. Hotfix candidate

PR #1327, opened 2020-11-27
Fix a mistake in pr#1307

ChromaPIE:
What:
Same as the title, fixed a mistake making the chief objective of pr#1307 meaningless. Silly me...
And also with another fix fixing a funny typo in zh_cn.lang, which made the Rubber Chip displayed as _Rubber is Pitiable_ in Chinese.

How solved:
Editting the three .lang files.

Outcome:
Fix  mistake in pr#1307


LAGIdiot (approved):
Thanks for fixing that mistake.

PR #1331, opened 2020-12-01
Changed shouldCheckWeakPower to work around bug

solidDoWant:
**What:**
A vanilla bug prevents getWeakPower from being called. This results in machines and covers being unable to set redstone outputs.

From line 3546 of World::getRedstonePower:
image

As the above image shows, the ternary statement is reversed.

**How solved:**
I've changed the two implementations of shouldCheckWeakPower to return 'false' instead of 'true', along with an inline explanation of why that's the case.

**Outcome:**
Fixes: Machines and covers now have the ability to output redstone.

No machines or covers currently exist that use this. The change should only affect addons.


LAGIdiot:
Thank you for spotting this issue and fixing it.

LAGIdiot (approved):

PR #1333, opened 2020-12-03
Fix Multismelter Voiding items.

### LINKED ISSUE #1332: [BUG] Multismelter can void items due to coil bonus
**Describe the bug**
The Multismelter can void items due to the coil bonus and having only a single output slot open.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.7.594
_Modpack:_ None
_Addons:_ SoG 2.15

**Setup**
Playing Solo
_New world generated_ Yes

**Steps To Reproduce**
1) Create a multismelter. I used IV input and output buses to get a large enough output slot. Additionally, I used fusion coils since they have a very large item smelting bonus to easily reproduce this issue.
2) Using Sand (or any other smeltable item), smelt it in the mutlismelter to fill the output bus with glass.
3) Add in two stacks of sand to the input bus of the multismelter.
4) Remove one stack of glass from the item output bus of the multismelter. 
5) See that the multismelter consumes both of the stack of sand in the input bus (due to the coil bonus) but only 1 stack gets put in the output bus, due to only having space for 1 stack.

**Expected behavior**
For the Multismelter not to function if there is not enough room in the output bus for all items smelted, including those added by the coil bonus

**Additional Information**
Attached GIF to demonstrate the issue

ALongStringOfNumbers:
**What:**
This PR fixes the Multismelter voiding items due to the coil bonus of the multismelter.

**How solved:**
An additional method is added to check the number of empty output slots in the multismelter. If there are no free output slots, then nothing is added to the output list. If there is room in the output slots, then the number of engaged items has to be less than the number of item room available in the output bus for new outputs to be added to the output list.

**Outcome:**
Fixes multismelter voiding items due to the coil bonus. Closes #1332.

None to be expected.

Edit: One thing I could use feedback on is the harcoded check in the for loop of number of free slots * 64. I am not sure if this would work in cases where the stack size of the output is < 64. It did work for a test recipe I added via crafttweaker for pumpkin pie -> egg, so that the output item had a max stack size of 16, but I am not sure if I covered all the possible cases.

### DIFF
--- a/src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java
+++ b/src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java
@@ -136,6 +136,11 @@ public class MetaTileEntityMultiFurnace extends RecipeMapMultiblockController {
             ArrayList<CountableIngredient> recipeInputs = new ArrayList<>();
             ArrayList<ItemStack> recipeOutputs = new ArrayList<>();
             for (int index = 0; index < inputs.getSlots(); index++) {
+                //Check first if there is room in the output bus
+                if(findFreeOutputSlots(this.getOutputInventory()) == 0) {
+                    break;
+                }
+
                 ItemStack stackInSlot = inputs.getStackInSlot(index);
                 if (stackInSlot.isEmpty())
                     continue;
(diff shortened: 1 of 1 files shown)


LAGIdiot (approved):
Thanks for spotting and fixing this issue.

LAGIdiot:
Closed in favor of #1337

Issue #1215, opened 2020-08-26
[Suggestion] Recipe System Rework


eutropius225:
**Is your feature request related to a problem? Please describe.**
The [Recipe](https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/api/recipes/Recipe.java) class and the entire recipe system revolves around items, fluids and EU being the only requirements for recipes. Unfortunately, this is insufficient, even within GTCE, the [EBF checks for coil temperature](https://github.com/GregTechCE/GregTech/blob/a3e6fdca2893c19e97d23f5d2f682830e5fb57b2/src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityElectricBlastFurnace.java#L68), which is not implemented very nicely, and is prone to bugs, such as #1157.

**Describe the solution you'd like**
A rework or overhaul of the recipe system, to an extent would be ideal, where certain checks and actions, such as [input dirtiness](https://github.com/GregTechCE/GregTech/blob/a3e6fdca2893c19e97d23f5d2f682830e5fb57b2/src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java#L183) could be delegated to the recipe itself.
- The new system should abstract away certain parts of recipe handling:
  - Parts of the code operating on recipes shouldn't primarily rely on inspecting the inputs and outputs of the recipe.
  - Parts of the code operating on recipes shouldn't assume that only items, fluids and energy are possible ingredients and requirements of recipes.
- It should be possible to add new ingredients (such as Mekanism gases) or predicates (such as blast furnace coil temperature) for recipes to handle.
- It should be possible to omit any/all of the basic ingredients (item/fluid/energy) where only other ingredient types are operated on.

LAGIdiot:
First of all thank you for bringing this up I really like this idea.

**Status:**
Accepted

**Reasons:**
This will open GTCE many possibilities for integration with other mods and even allow as to make machines which would not be possible before. Also this will give us edge against other mods and make us more irresistible to modpack authors.

**Additional notes:**
Of course this will be compatibility nightmare and it has to be designed to break nothing of existing recipes and systems on our side. All changes to API has to be documented. And extended testing period should be allocated to this.

fmarceau:
[...]

PR #1337, opened 2020-12-05
Fix Multiblocks voiding partial multi item outputs when output is mostly Full

### LINKED ISSUE #1332: [BUG] Multismelter can void items due to coil bonus

### LINKED ISSUE #1334: [BUG] Output voiding for multi item output recipes is possible with GTCE mutliblocks
**Describe the bug**
It is possible to void part of an output from a recipe that has multiple item outputs in several GTCE multiblocks. The voiding occurs when having only 1 free slot in the output bus. The recipe will run once, fill the empty slot in the output bus with the first output from the recipe and void the other outputs from the recipe. Then the multiblock will stop running due to not having room for both item outputs. This has been tested with the blast furnace and the Implosion compressor.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.7.594
_Modpack:_ None
_Addons:_ SoG 2.15

**Setup**
Playing Solo
_New world generated_ Yes

**Steps To Reproduce**
1) Construct a Electric Blast Furnace, and use an LV item output to quickly reproduce the issue
2) Using the Ruby Dust recipe (1 ruby dust -> 5 aluminum nuggets + 4 tiny pile of dark ash), run the EBF until there are 2 slots of the LV item output occupied by the aluminum dust and 1 slot occupied by the dark ash. Make sure that all the stacks of these items are full stacks.
3) Place in 2 more ruby dust (2 recipes)
4) Notice that the recipe will be run once, and that 5 aluminum nuggets will be added to the open slot in the output bus, but the dark ash will be voided. And also notice that the EBF will not process the second recipe, because the output bus does not have room.

This can be repeated in the Implosion Compressor, and was repeated using the topaz dust to topaz gem recipe.

**Expected behavior**
For the mutliblock not to run if there is not room in the output bus for all the outputs of the recipe.

ALongStringOfNumbers:
**What:**
This PR fixes multiblocks voiding parts of multi item outputs when the output bus only has room for the first output of the recipe.

**How solved:**
This PR adds logic and methods checking if the output of recipes can merge into available slots in the output bus to prevent the issue of output voiding due to checking each item individually, which missed out on possible output inventory changes, causing some portions of the output to be voided.

**Outcome:**
Fixes Multiblocks voiding partial recipe outputs when output bus is mostly full. Closes #1334. Closes #1332. Supersedes #1333.

The old method was redirected to the new method, so there should be minimal chance for compatibility issues.

### DIFF
--- a/src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
@@ -2,6 +2,7 @@ package gregtech.api.capability.impl;
 
 import gregtech.api.capability.IEnergyContainer;
 import gregtech.api.capability.IMultipleTankHandler;
+import gregtech.api.metatileentity.MetaTileEntity;
 import gregtech.api.metatileentity.multiblock.RecipeMapMultiblockController;
 import gregtech.api.recipes.Recipe;
 import net.minecraftforge.items.IItemHandlerModifiable;
@@ -54,10 +55,38 @@ public class MultiblockRecipeLogic extends AbstractRecipeLogic {
     protected boolean setupAndConsumeRecipeInputs(Recipe recipe) {
         RecipeMapMultiblockController controller = (RecipeMapMultiblockController) metaTileEntity;
         if (controller.checkRecipe(recipe, false) &&
-            super.setupAndConsumeRecipeInputs(recipe)) {
+            multiBlockSetupAndConsumeRecipeInputs(recipe)) {
             controller.checkRecipe(recipe, true);
             return true;
         } else return false;
+    }
+
+    //Logic Mostly copied from AbstractRecipeLogic, but with some additional checking for Multiblock output spacing to prevent voiding
+    protected boolean multiBlockSetupAndConsumeRecipeInputs(Recipe recipe) {
+        int[] resultOverclock = calculateOverclock(recipe.getEUt(), getMaxVoltage(), recipe.getDuration());
+        int totalEUt = resultOverclock[0] * resultOverclock[1];
+        IItemHandlerModifiable importInventory = getInputInventory();
+        IItemHandlerModifiable exportInventory = getOutputInventory();
+        IMultipleTankHandler importFluids = getInputTank();
+        IMultipleTankHandler exportFluids = getOutputTank();
+        return (totalEUt >= 0 ? getEnergyStored() >= (totalEUt > getEnergyCapacity() / 2 ? resultOverclock[0] : totalEUt) :
+            (getEnergyStored() - resultOverclock[0] <= getEnergyCapacity())) &&
+            MetaTileEntity.addItemsToItemHandler(exportInventory, true, recipe.getAllItemOutputs(exportInventory.getSlots())) &&
+            recipe.getAllItemOutputs(exportInventory.getSlots()).size() <= getFreeSlots(exportInventory) &&
+            MetaTileEntity.addFluidsToFluidHandler(exportFluids, true, recipe.getFluidOutputs()) &&
+            recipe.matches(true, importInventory, importFluids);
+    }
+
+    //Determines the number of free slots in the paseed inventory
+    protected int getFreeSlots(IItemHandlerModifiable inventory) {
+        int emptySlots = 0;
+        for(int index = 0; index < inventory.getSlots(); index++) {
+            if(inventory.getStackInSlot(index).isEmpty()) {
+                emptySlots++;
+            }
+        }
+
+        return emptySlots;
     }
 
     @Override
--- a/src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
@@ -5,7 +5,12 @@ import gregtech.api.capability.IMultipleTankHandler;
 import gregtech.api.metatileentity.MetaTileEntity;
 import gregtech.api.metatileentity.multiblock.RecipeMapMultiblockController;
 import gregtech.api.recipes.Recipe;
+import net.minecraft.item.ItemStack;
 import net.minecraftforge.items.IItemHandlerModifiable;
+
+import java.util.List;
+
+import static gregtech.api.util.InventoryUtils.simulateItemStackMerge;
 
 public class MultiblockRecipeLogic extends AbstractRecipeLogic {
 
@@ -69,24 +74,13 @@ public class MultiblockRecipeLogic extends AbstractRecipeLogic {
         IItemHandlerModifiable exportInventory = getOutputInventory();
         IMultipleTankHandler importFluids = getInputTank();
         IMultipleTankHandler exportFluids = getOutputTank();
+        List<ItemStack> itemOutputs = recipe.getAllItemOutputs(exportInventory.getSlots());
         return (totalEUt >= 0 ? getEnergyStored() >= (totalEUt > getEnergyCapacity() / 2 ? resultOverclock[0] : totalEUt) :
             (getEnergyStored() - resultOverclock[0] <= getEnergyCapacity())) &&
-            MetaTileEntity.addItemsToItemHandler(exportInventory, true, recipe.getAllItemOutputs(exportInventory.getSlots())) &&
-            recipe.getAllItemOutputs(exportInventory.getSlots()).size() <= getFreeSlots(exportInventory) &&
+            MetaTileEntity.addItemsToItemHandler(exportInventory, true, itemOutputs) &&
+            simulateItemStackMerge(itemOutputs, exportInventory) &&
             MetaTileEntity.addFluidsToFluidHandler(exportFluids, true, recipe.getFluidOutputs()) &&
             recipe.matches(true, importInventory, importFluids);
-    }
-
-    //Determines the number of free slots in the paseed inventory
-    protected int getFreeSlots(IItemHandlerModifiable inventory) {
-        int emptySlots = 0;
-        for(int index = 0; index < inventory.getSlots(); index++) {
-            if(inventory.getStackInSlot(index).isEmpty()) {
-                emptySlots++;
-            }
-        }
-
-        return emptySlots;
     }
 
     @Override
--- a/src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
@@ -2,15 +2,9 @@ package gregtech.api.capability.impl;
 
 import gregtech.api.capability.IEnergyContainer;
 import gregtech.api.capability.IMultipleTankHandler;
-import gregtech.api.metatileentity.MetaTileEntity;
 import gregtech.api.metatileentity.multiblock.RecipeMapMultiblockController;
 import gregtech.api.recipes.Recipe;
-import net.minecraft.item.ItemStack;
 import net.minecraftforge.items.IItemHandlerModifiable;
-
-import java.util.List;
-
-import static gregtech.api.util.InventoryUtils.simulateItemStackMerge;
 
 public class MultiblockRecipeLogic extends AbstractRecipeLogic {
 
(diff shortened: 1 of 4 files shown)


src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
```diff
@@ -54,12 +55,40 @@ protected IMultipleTankHandler getOutputTank() {
...
+        int[] resultOverclock = calculateOverclock(recipe.getEUt(), getMaxVoltage(), recipe.getDuration());
+        int totalEUt = resultOverclock[0] * resultOverclock[1];
+        IItemHandlerModifiable importInventory = getInputInventory();
+        IItemHandlerModifiable exportInventory = getOutputInventory();
+        IMultipleTankHandler importFluids = getInputTank();
+        IMultipleTankHandler exportFluids = getOutputTank();
+        return (totalEUt >= 0 ? getEnergyStored() >= (totalEUt > getEnergyCapacity() / 2 ? resultOverclock[0] : totalEUt) :
+            (getEnergyStored() - resultOverclock[0] <= getEnergyCapacity())) &&
+            MetaTileEntity.addItemsToItemHandler(exportInventory, true, recipe.getAllItemOutputs(exportInventory.getSlots())) &&
+            recipe.getAllItemOutputs(exportInventory.getSlots()).size() <= getFreeSlots(exportInventory) &&
```
LAGIdiot, line 75: These two same calls to `recipe.getAllItemOutputs` should be consolidated to one before return and values from it used here.

src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
```diff
@@ -54,12 +55,40 @@ protected IMultipleTankHandler getOutputTank() {
...
+            MetaTileEntity.addFluidsToFluidHandler(exportFluids, true, recipe.getFluidOutputs()) &&
+            recipe.matches(true, importInventory, importFluids);
+    }
+
+    //Determines the number of free slots in the paseed inventory
+    protected int getFreeSlots(IItemHandlerModifiable inventory) {
```
LAGIdiot, line 81: This name `getFreeSlots` does not sound correct to me as it does not return free slots but just number of them.
Maybe it should go even to utils as it is not something specific for this and anyone could use it then.

src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
```diff
@@ -54,12 +55,40 @@ protected IMultipleTankHandler getOutputTank() {
...
+        IMultipleTankHandler exportFluids = getOutputTank();
+        return (totalEUt >= 0 ? getEnergyStored() >= (totalEUt > getEnergyCapacity() / 2 ? resultOverclock[0] : totalEUt) :
+            (getEnergyStored() - resultOverclock[0] <= getEnergyCapacity())) &&
+            MetaTileEntity.addItemsToItemHandler(exportInventory, true, recipe.getAllItemOutputs(exportInventory.getSlots())) &&
+            recipe.getAllItemOutputs(exportInventory.getSlots()).size() <= getFreeSlots(exportInventory) &&
+            MetaTileEntity.addFluidsToFluidHandler(exportFluids, true, recipe.getFluidOutputs()) &&
+            recipe.matches(true, importInventory, importFluids);
+    }
+
+    //Determines the number of free slots in the paseed inventory
```
LAGIdiot, line 80: Typo in `paseed`. But this comment won't be needed when that functions is that short with really descriptive name

src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
LAGIdiot, line 75: Also there is more elaborate logic required as you need to know number of slots you can put items in to and their capacity. 

For example when I build Implosion compressor. With LV Output bus (4 slots) and I let process 4x Amethyst Dust + 2x TNT => 3x Amethyst + 2x Tiny Pile of Dark Ashes it will only process 22x and then stop even if there is still capacity in output for another 10 runs.

LAGIdiot (changes requested):
Thank you for creating this PR. There few things that needs to be sorted out.

src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
ALongStringOfNumbers: I have adjusted the name of the function. Let me know if it should be adjusted further.

src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
ALongStringOfNumbers: There has been a refactor to catch this case. Now the recipe output will attempt to merge into the existing stacks in the output bus before attempting to add the recipe output items to empty slots in the output bus. The above issue no longer exists, and the voiding issue is also solved.

src/main/java/gregtech/api/util/InventoryUtils.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        // Deep copy the recipe output itemstacks
+        List<ItemStack> itemStackList = new ArrayList<>(items.size());
+        for(ItemStack i : items) {
+            itemStackList.add(i.copy());
+        }
+
+        // Sort by the number of items in each stack so we merge smallest stacks first.
+        itemStackList.sort((i,j) -> Integer.compare(i.getCount(), j.getCount()));
```
LAGIdiot, line 52: I think we should use comparator here so it can be more readable.

src/main/java/gregtech/api/util/InventoryUtils.java
```diff
@@ -0,0 +1,93 @@
...
+
+        // Deep copy the contents of the output bus
+        int numSlots = inventory.getSlots();
+        List<ItemStack> inventoryStacks = new ArrayList<>(numSlots);
+        for(int slotIndex = 0; slotIndex < numSlots; slotIndex++) {
+            ItemStack stack = inventory.getStackInSlot(slotIndex);
+            if(!stack.isEmpty())
+                inventoryStacks.add(stack.copy());
+        }
```
LAGIdiot, line 61: This should be extracted to own method so there is not such complexity

src/main/java/gregtech/api/util/InventoryUtils.java
```diff
@@ -0,0 +1,93 @@
...
+
+        // Since we're mutating the collection during iteration, use an iterator.
+        Iterator<ItemStack> outIter = itemStackList.iterator();
+        while(outIter.hasNext()) {
```
LAGIdiot, line 65: I had quite hard time reading this. Is it really easiest way how to do this?

src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
```diff
@@ -54,12 +60,29 @@ protected IMultipleTankHandler getOutputTank() {
...
 
+    //Logic Mostly copied from AbstractRecipeLogic, but with some additional checking for Multiblock output spacing to prevent voiding
+    protected boolean multiBlockSetupAndConsumeRecipeInputs(Recipe recipe) {
+        int[] resultOverclock = calculateOverclock(recipe.getEUt(), getMaxVoltage(), recipe.getDuration());
+        int totalEUt = resultOverclock[0] * resultOverclock[1];
+        IItemHandlerModifiable importInventory = getInputInventory();
+        IItemHandlerModifiable exportInventory = getOutputInventory();
+        IMultipleTankHandler importFluids = getInputTank();
+        IMultipleTankHandler exportFluids = getOutputTank();
+        List<ItemStack> itemOutputs = recipe.getAllItemOutputs(exportInventory.getSlots());
```
LAGIdiot, line 77: This can still void items if for example you use same recipe as before and you have only Export Bus with 1 slot. Then recipe will process but it will gave only Amethyst and void Tiny Pile of Dark Ashes. Which is probably not something we want

src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
```diff
@@ -54,12 +60,29 @@ protected IMultipleTankHandler getOutputTank() {
...
+        return (totalEUt >= 0 ? getEnergyStored() >= (totalEUt > getEnergyCapacity() / 2 ? resultOverclock[0] : totalEUt) :
+            (getEnergyStored() - resultOverclock[0] <= getEnergyCapacity())) &&
+            MetaTileEntity.addItemsToItemHandler(exportInventory, true, itemOutputs) &&
```
LAGIdiot, line 80: I took another look at this and this function does exactly what you are doing with new one. But this one does not work correctly when more then one item is given and it takes up last available slot. As it simulates input for every item one by one without considering others.
I think we should fix this function not creating new one that does same thing. But we sure can use knowledge obtained when designing new merge function.

LAGIdiot (changes requested):
Thank you for updating this PR. I give it another look and there are sine new insights and comments from me so please take a look at them

src/main/java/gregtech/api/util/InventoryUtils.java
Exaxxion: If one wishes to change the contents of a collection while they are iterating through it, the only safe way to do so is with `Iterator#remove()`, otherwise you will get a `ConcurrentModificationException`.

src/main/java/gregtech/api/util/InventoryUtils.java
Exaxxion: That one's on me, I was writing the snippet in a plaintext editor and couldn't recall the precise syntax of the method reference operator signature.

This lambda is functionally equivalent to
```java
Comparator.comparingInt(ItemStack::getCount)
```

src/main/java/gregtech/api/util/InventoryUtils.java
ALongStringOfNumbers: This has been done.

src/main/java/gregtech/api/util/InventoryUtils.java
ALongStringOfNumbers: This has been done.

src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
ALongStringOfNumbers: This has been done. The logic in addItemsToItemHandler has been adjusted to call the new method logic, and the simulate case from the original addItemsToItemHandler method is also handled.

src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
ALongStringOfNumbers: As mentioned on discord, this is currently not a bug due to the logic of getAllItemOutputs truncating the number of outputs to the maximum number of slots. This should probably be looked at more in depth in a different issue to make sure that all consequences of changing it, if that is what is desired, are handled.

src/main/java/gregtech/api/util/InventoryUtils.java
ALongStringOfNumbers: There has been some attempt to make the logic here clearer.

src/main/java/gregtech/api/capability/impl/MultiblockRecipeLogic.java
LAGIdiot: Yeah we can skip this one here and follow up on it in #1215

src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java
```diff
@@ -166,4 +203,46 @@ protected Recipe findRecipe(long maxVoltage, IItemHandlerModifiable inputs, IMul
...
+            }
+
+            // if there is a partial stack, add it too
+            if(remainder > 0) {
+                ItemStack partial = outputStack.copy();
+                partial.setCount(remainder);
+                recipeOutputs.add(partial);
+            }
+        }
+    }
```
LAGIdiot, line 246: Move this to `MultiFurnaceWorkable` as only that inner class is using it.

LAGIdiot (changes requested):
Thank you for updating this PR. There is only one last thing that need to get fixed.

Also I think that comments in some cases are quite extensive and not necessary as what is going on can be picked from code (like max items allowed in a stack) or documentation comment for private functions. But there is no need to do anything about it.

src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java
ALongStringOfNumbers: This has been done.

LAGIdiot (approved):
Good work finishing it up.

PR #1348, opened 2020-12-13
Fix incorrect Lapotron Crystal crafting recipe

### LINKED ISSUE #1336: [BUG] Lapotron Crystals craft to wrong maximum charge.
**Describe the bug**
In #1189, the maximum charge of the Lapotron Crystals was increased to 10M EU. However, the crafting recipe still crafts to Lapotron crystals with 4M EU Maximum charge.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.7.596
_Modpack:_ None
_Addons:_ None

**Setup**
Playing Solo
_New world generated_ Yes

**Steps To Reproduce**
1) Look at the Lapotron crystal recipe in JEI and see that it has 10M EU max charge.
2) Look at the crafting recipe and see that it crafts to one with 4M EU max charge

**Expected behavior**
To craft to 10M EU maximum charge

ALongStringOfNumbers:
**What:**
This PR fixes the incorrect Lapotron Crystal crafting recipe, where the Lapotron Crystal had a capacity of 10M EU, but the crafting recipe only crafted to one with 4M EU.

**How solved:**
A new method was added, which takes an additional parameter of the maximum charge, which then sets the output of the crafting recipe to have a maximum charge of that amount. This is used in conjunction with a new helper method in `MetaItem` that will return the maximum charge of an electric item.

**Outcome:**
Fixes Lapotron Crystal crafting to one with incorrect EU capacity. Closes #1336.

**Additional info:**

This is the case for all variants of the Lapotron, they craft to the correct amount

None expected

### DIFF
--- a/src/main/java/gregtech/api/items/metaitem/MetaItem.java
+++ b/src/main/java/gregtech/api/items/metaitem/MetaItem.java
@@ -835,16 +835,6 @@ public abstract class MetaItem<T extends MetaItem<?>.MetaValueItem> extends Item
             return itemStack;
         }
 
-        public long getChargeAmount() {
-            ItemStack itemstack = getStackForm(1);
-            IElectricItem electricItem = itemstack.getCapability(GregtechCapabilities.CAPABILITY_ELECTRIC_ITEM, null);
-            if(electricItem == null) {
-                throw new IllegalStateException("Not an electric item");
-            }
-
-            return electricItem.getMaxCharge();
-        }
-
         public ItemStack getInfiniteChargedStack() {
             ItemStack itemStack = getStackForm(1);
             IElectricItem electricItem = itemStack.getCapability(GregtechCapabilities.CAPABILITY_ELECTRIC_ITEM, null);
--- a/src/main/java/gregtech/api/items/metaitem/MetaItem.java
+++ b/src/main/java/gregtech/api/items/metaitem/MetaItem.java
@@ -835,6 +835,16 @@ public abstract class MetaItem<T extends MetaItem<?>.MetaValueItem> extends Item
             return itemStack;
         }
 
+        public long getChargeAmount() {
+            ItemStack itemstack = getStackForm(1);
+            IElectricItem electricItem = itemstack.getCapability(GregtechCapabilities.CAPABILITY_ELECTRIC_ITEM, null);
+            if(electricItem == null) {
+                throw new IllegalStateException("Not an electric item");
+            }
+
+            return electricItem.getMaxCharge();
+        }
+
         public ItemStack getInfiniteChargedStack() {
             ItemStack itemStack = getStackForm(1);
             IElectricItem electricItem = itemStack.getCapability(GregtechCapabilities.CAPABILITY_ELECTRIC_ITEM, null);
--- a/src/main/java/gregtech/api/recipes/ModHandler.java
+++ b/src/main/java/gregtech/api/recipes/ModHandler.java
@@ -260,25 +260,10 @@ public class ModHandler {
     }
 
     public static void addShapedEnergyTransferRecipe(String regName, ItemStack result, Predicate<ItemStack> chargePredicate, boolean transferMaxCharge, Object... recipe) {
-        boolean skip = false;
-        if (result.isEmpty()) {
-            GTLog.logger.error("Result cannot be an empty ItemStack. Recipe: {}", regName);
-            GTLog.logger.error("Stacktrace:", new IllegalArgumentException());
-            skip = true;
-        }
-        skip |= validateRecipe(regName, recipe);
-        if (skip) {
-            RecipeMap.setFoundInvalidRecipe(true);
-            return;
-        }
-
-        IRecipe shapedOreRecipe = new ShapedOreEnergyTransferRecipe(null, result.copy(), chargePredicate, transferMaxCharge, finalizeShapedRecipeInput(recipe))
-            .setMirrored(false) //make all recipes not mirrored by default
-            .setRegistryName(regName);
-        ForgeRegistries.RECIPES.register(shapedOreRecipe);
+        addShapedEnergyTransferRecipeWithOverride(regName, result, chargePredicate, true, transferMaxCharge, recipe);
     }
 
-    public static void addShapedEnergyTransferRecipe(String regName, ItemStack result, Predicate<ItemStack> chargePredicate, long chargeAmount, boolean transferMaxCharge, Object... recipe) {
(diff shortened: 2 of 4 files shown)


src/main/java/gregtech/loaders/recipe/CraftingRecipeLoader.java
```diff
@@ -90,7 +90,7 @@ private static void loadCraftingRecipes() {
         for(Material material : new Material[] {Materials.Lapis, Materials.Lazurite, Materials.Sodalite}) {
             String recipeName = "lapotron_crystal_" + material.toString();
             ModHandler.addShapedEnergyTransferRecipe(recipeName, MetaItems.LAPOTRON_CRYSTAL.getStackForm(),
-                Ingredient.fromStacks(MetaItems.ENERGY_CRYSTAL.getStackForm()), false,
+                Ingredient.fromStacks(MetaItems.LAPOTRON_CRYSTAL.getStackForm()), MetaItems.LAPOTRON_CRYSTAL.getChargeAmount(), false,
```
LAGIdiot, line 93: You need to pass in info from which items it is crafted so it can transfer energy from them - right now it just wastes energy from ENERGY_CRYSTAL

src/main/java/gregtech/api/util/ShapedOreEnergyTransferRecipe.java
```diff
@@ -25,13 +25,31 @@ public ShapedOreEnergyTransferRecipe(ResourceLocation group, @Nonnull ItemStack
...
         this.transferMaxCharge = transferMaxCharge;
         fixOutputItemMaxCharge();
     }
 
+    public ShapedOreEnergyTransferRecipe(ResourceLocation group, @Nonnull ItemStack result, Predicate<ItemStack> chargePredicate, long chargeAmount, boolean transferMaxCharge, ShapedPrimer primer) {
+        super(group, result, primer);
+        this.chargePredicate = chargePredicate;
+        this.transferMaxCharge = transferMaxCharge;
+        setMaxChargeInRecipe(chargeAmount);
+    }
```
LAGIdiot, line 44: This is wrong you are overriding crafted item max charge to it's max charge. This will produce unnecessary NBT mess and use calculation time.
For this case we should just not call `fixOutputItemMaxCharge` which means no override which means default max energy level of item is used

src/main/java/gregtech/api/recipes/ModHandler.java
```diff
@@ -278,6 +278,25 @@ public static void addShapedEnergyTransferRecipe(String regName, ItemStack resul
...
+            RecipeMap.setFoundInvalidRecipe(true);
+            return;
+        }
+
+        IRecipe shapedOreRecipe = new ShapedOreEnergyTransferRecipe(null, result.copy(), chargePredicate, chargeAmount, transferMaxCharge, finalizeShapedRecipeInput(recipe))
+            .setMirrored(false) //make all recipes not mirrored by default
+            .setRegistryName(regName);
+        ForgeRegistries.RECIPES.register(shapedOreRecipe);
+    }
+
```
LAGIdiot, line 299: This is nearly 100% code duplication except one param in method header and then passing it. This could be done by extending current one by that param and its passing down. And creating new function with old param set which just calls that updated one.

LAGIdiot (changes requested):
Thank you for creating this PR. There are some problems with it so please see to them

src/main/java/gregtech/loaders/recipe/CraftingRecipeLoader.java
ALongStringOfNumbers: This has been fixed.

src/main/java/gregtech/api/recipes/ModHandler.java
ALongStringOfNumbers: I have done so

src/main/java/gregtech/api/util/ShapedOreEnergyTransferRecipe.java
ALongStringOfNumbers: This is now done. A boolean is passed in for the new method to determine if `fixOutputItemMaxCharge` will be called, with all the old recipes having this boolean as true by default due to a redirected method.

src/main/java/gregtech/api/items/metaitem/MetaItem.java
```diff
@@ -835,6 +835,16 @@ public ItemStack getChargedStack(long chargeAmount) {
...
+        public long getChargeAmount() {
+            ItemStack itemstack = getStackForm(1);
+            IElectricItem electricItem = itemstack.getCapability(GregtechCapabilities.CAPABILITY_ELECTRIC_ITEM, null);
+            if(electricItem == null) {
+                throw new IllegalStateException("Not an electric item");
+            }
+
+            return electricItem.getMaxCharge();
+        }
+
```
LAGIdiot, line 847: I don't think this is needed anymore

src/main/java/gregtech/api/recipes/ModHandler.java
```diff
@@ -260,6 +260,10 @@ public static void addShapedRecipe(String regName, ItemStack result, Object... r
     }
 
     public static void addShapedEnergyTransferRecipe(String regName, ItemStack result, Predicate<ItemStack> chargePredicate, boolean transferMaxCharge, Object... recipe) {
+        addShapedEnergyTransferRecipeWithOverride(regName, result, chargePredicate, true, transferMaxCharge, recipe);
+    }
+
+    public static void addShapedEnergyTransferRecipeWithOverride(String regName, ItemStack result, Predicate<ItemStack> chargePredicate, boolean overrideCharge, boolean transferMaxCharge, Object... recipe) {
```
LAGIdiot, line 266: I am not fond of this name but I understand why you named this function differently.
But I am not getting any idea for better method name myself.

LAGIdiot (changes requested):
Good work on this PR. There are few last comments then needs to be addressed and then it should be finished.

src/main/java/gregtech/api/items/metaitem/MetaItem.java
ALongStringOfNumbers: I have removed it

src/main/java/gregtech/api/recipes/ModHandler.java
ALongStringOfNumbers: If you think of a better name, let me know and I will change it

LAGIdiot: Fair enough, let's go with current one.

LAGIdiot (approved):

PR #1351, opened 2020-12-16
Fix the Primitive Blast Furnace Voiding outputs

ALongStringOfNumbers:
**What:**
After the latest version of GTCE (1.10.8.599) was released, it was noticed that the Primitive Blast Furnace was voiding both the steel and the ashes outputs when the ash output in the output slot was too high and thus the recipe outputs could not be merged into the output slots

**How solved:**
Since the PBF is somewhat special and checks for room based on only the recipe outputs (not including the ash), and then adds the ash outputs after, there were issues with the outputs voiding. To keep with the previous behavior of the PBF, it has been adjusted to void the ash outputs when those are full/almost full, while continuing to produce steel. The steel will not be voided however, when its output slot fills up. The PBF will instead stop. However, this is easy to change so that the PBF will not function is there is not room for both the Ash and the recipe output. Please let me know what you prefer in this case.

**Outcome:**
Fixes recipe outputs being voided in the Primitive Blast Furnace if one output slot was full

### DIFF
--- a/src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityPrimitiveBlastFurnace.java
+++ b/src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityPrimitiveBlastFurnace.java
@@ -47,6 +47,8 @@ import net.minecraftforge.items.IItemHandlerModifiable;
 import net.minecraftforge.items.ItemHandlerHelper;
 import net.minecraftforge.items.ItemStackHandler;
 
+import static gregtech.api.util.InventoryUtils.simulateItemStackMerge;
+
 import javax.annotation.Nonnull;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


LAGIdiot (commented):
This seems strange. If I understand it correctly when there is no place for ashes just produce main output? That should not be like that. It should be if there is not enough space for both then don't produce right?

LAGIdiot:
Sorry I did not finished reading you PR opening comment and go straight to code. So you can skip my first comment.

I think this should have been open as bug first when we could iron out that question regarding ash voiding so you don't have to change implementation (no you don't have to open bug now).

Anyway for situation when ash is full but there is still space for steel then it should not process as we don't want any voiding.

LAGIdiot (changes requested):
Please update as per my last comment

ALongStringOfNumbers:
Ok, the PR has been updated to no longer void the ash when it reaches near maximum stack size, instead the PBF will stop until the outputs are cleared from it. I will suggest though that maybe the PBF needs another output slot now.

LAGIdiot (approved):
Thanks for finishing this PR.

Issue #1335, opened 2020-12-04
[Suggestion] Getting the unification preferred item via CT


Charnuz:
**Describe the solution you'd like**
Since output doesn't support ore dictionary entries, you have to do `<ore:fooBar>.firstItem` to get an acceptable output. This often times does not match what would be returned by unification.

LAGIdiot:
I am not sure exactly what you are trying to suggest here. Can I get some examples for it?

Charnuz:
Basically, getting the item from an ore dictionary entry like "_ingotCopper_" that follows GTCE's ore dictionary priority sorting. This really only matters for recipe output in most cases.

With ThermalFoundation and Forestry installed and `B:useCustomModPriorities` set to false:
- Forestry's copper ingot will be the preferred item for recipe output
- getting the first item in the ore dictionary entry via CT `<ore:ingotCopper>.firstItem` will usually match this (but not always, very mod load order dependent)



But with `B:useCustomModPriorities` set to true and the following custom mod priorities, ThermalFoundation's copper ingot will instead be the preferred item:
```
    S:modPriorities <
        thermalfoundation
        forestry
     > 
```
However, getting the first item in the dictionary entry will rarely match the actual item with most priority. Ideally there would be a function that accepts an ore dictionary entry and outputs the item with the from the mod with the most priority.

LAGIdiot:
Ah I see what you mean and it seems as reasonable request. But I am not sure if we can even do that as this functionality belongs to CT it self. @idcppl any idea how to do this?

PR #1352, opened 2020-12-17
Getting generated MetaItem's as an IItemStack

idcppl:
**What:**  
This would allow you to get material generated IItemStack's through the bracket handler, and would make looping over MetaItem's everytime not a thing. Also somewhat related to #1335 
So the bracket handler could also accept the oreDict name, for example, `<metaitem:plateIron>`


src/main/java/gregtech/common/CommonProxy.java
```diff
@@ -173,6 +174,7 @@ public static void registerRecipesLowest(RegistryEvent.Register<IRecipe> event)
         DecompositionRecipeHandler.runRecipeGeneration();
         RecyclingRecipes.init();
         WoodMachineRecipes.init();
+        MetaItemBracketHandler.rebuildComponentRegistry();
```
LAGIdiot, line 177: Based on what did you choose this place?

LAGIdiot (commented):
Can you please be more descriptive in PR opening comment and provide some real use cases.
As I am not that familiar with this matter provided information does not say much to me.

src/main/java/gregtech/common/CommonProxy.java
idcppl: This is the same registry spot that CraftTweaker uses for their bracket handlers.

idcppl:
This would allow for all Metaitems that gregtech generates to be gotten in an IItemStack form that doesn't have to rely on getting the first in oredict. There is probably some more things I could make smoother. If you really don't care for the material based generation items being in this, I could just remove that part. The old system looped everytime the bracket was called and very inefficient.

idcppl:
An example would be,  
```zs
RecipeMap.getByName("macerator").recipeBuilder()
.inputs([<ore:ingotSteel>])
.outputs(<metaitem:ingotSteel>)
.duration(60)
.EUt(30)
.buildAndRegister();

vs

RecipeMap.getByName("macerator").recipeBuilder()
.inputs([<ore:ingotSteel>])
.outputs(<ore:ingotSteel>.firstItem)
.duration(60)
.EUt(30)
.buildAndRegister();
```
It's almost the same, but the metaitem bracket handler guarantee's that it'll be GTCE ingot, only if it's generated by GTCE, else just use `<ore:ingotSteel>.firstItem`

src/main/java/gregtech/api/recipes/crafttweaker/MetaItemBracketHandler.java
```diff
@@ -6,34 +6,58 @@
 import crafttweaker.api.item.IItemStack;
 import crafttweaker.mc1120.item.MCItemStack;
 import crafttweaker.zenscript.IBracketHandler;
+import gregtech.api.items.materialitem.MaterialMetaItem;
 import gregtech.api.items.metaitem.MetaItem;
 import gregtech.api.items.metaitem.MetaItem.MetaValueItem;
+import gregtech.api.unification.OreDictUnifier;
+import gregtech.common.items.MetaItems;
```
LAGIdiot, line 13: This coupling is wrong. We should not access GTCE common from API. Please look for other way how to do it.

src/main/java/gregtech/common/CommonProxy.java
LAGIdiot: That seems like good place then.

LAGIdiot (changes requested):
Good work updating PR. There is one thing that need to be resolved everything else seems reasonable.

idcppl:
Believe this should be finished, now that common is out of api area.

LAGIdiot (approved):
Thank you for finishing this up.

PR #1353, opened 2020-12-17
Streamlining FuelRecipeMap to be like RecipeMap

idcppl:
**What:**
Made it so that you can get FuelRecipeMap's by doing `FuelRecipeMap.getByName("gas_turbine");`. Also exposed it properly to CraftTweaker.

### DIFF
--- a/src/main/java/gregtech/api/recipes/machines/FuelRecipeMap.java
+++ b/src/main/java/gregtech/api/recipes/machines/FuelRecipeMap.java
@@ -39,9 +39,9 @@ public class FuelRecipeMap {
     }
 
     @ZenMethod
-    public static FuelRecipeMap getByName(String name) {
+    public static FuelRecipeMap getByName(String unlocalizedName) {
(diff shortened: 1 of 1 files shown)


src/main/java/gregtech/api/recipes/machines/FuelRecipeMap.java
```diff
@@ -34,6 +38,13 @@ public FuelRecipeMap(String unlocalizedName) {
         return RECIPE_MAPS;
     }
 
+    @ZenMethod
+    public static FuelRecipeMap getByName(String name) {
```
LAGIdiot, line 42: That parameter should be `unlocalizedName` so it is consistent with RecipeMap and with filter

LAGIdiot (changes requested):
Good work on this PR. There is one minor thing which needs to be adjusted.

LAGIdiot (approved):
Thank you for updating this PR.

PR #1354, opened 2020-12-17
Fall through instead of returning null in traceCoverSide

eutropius225:
**What:**
This PR removes an illogical and unnecessary null return.

**Outcome:**
`gregtech.api.cover.ICoverable#traceCoverSide` will no longer return null under specific conditions.

It doesn't make sense for it to return null ever: there was still a block raytraced, and thus a side hit, after all.

None


LAGIdiot (approved):
Thank for creating this PR.

PR #1355, opened 2020-12-18
Fix large boiler heatEfficiencyMultiplier

amelentev:
**What:**
Fix typo in Large Boiler Heat Efficiency calc.
Heat Efficiency was always 100% before. Because `(boilerType.temperatureEffBuff * temperature) / 100.0` is always < 0.5 and `round(it)==0`.

**Outcome:**
Heat Efficiency now is up to 132% as designed in BoilerType.temperatureEffBuff.


LAGIdiot (approved):
Thanks for noticing this issue and fixing it.

Issue #1358, opened 2020-12-20
[BUG] Adding a material and adding both GENERATE_FLUID_BLOCK and GENERATE_PLASMA flags crashes the game


hjae78:
**Describe the bug**
When generating a material and giving it both the tags GENERATE_FLUID_BLOCK and GENERATE_PLASMA the game will not launch.

**Versions**
GTCE: Latest

**Setup**
Dev environment

**Steps To Reproduce**
1) Generate material and give it both GENERATE_FLUID_BLOCK and GENERATE_PLASMA flags.
2) Try to launch game.

**Expected behavior**
That the game should launch with a plasma for the material to be generated.

LAGIdiot:
Thank you for reporting this issue. I think that crashlog or at least log should be provided for this type of error.

Anyway problem is related to way we register this fluids.
We call https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/common/MetaFluids.java#L166 twice first time for fluid version and second time for plasma version. And if flag GENERATE_FLUID_BLOCK is set we will generate two FluidsBlocks with same name https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/common/MetaFluids.java#L191 which is obviously not allowed. 

Now we should not crash game. Which means this should be fixed. But GTCE did not ever expected need for place able plasma liquid block.

So do you need place able plasma block?

hjae78:
No, my intention was not to generate a placeable plasma block, i simply added plasma to a pre-existing material that had the fluid block flag.

LAGIdiot:
I understand your intention and what happened.

Right now we don't have any use for placeable plasma blocks. So there is no need to generate placeable variant of it. If we need it later we can add it.

PR #1359, opened 2020-12-20
Fixed FluidMaterial crashing the game if GENERATE_FLUID_BLOCK and GENERATE_PLASMA are present

idcppl:
**What:**
fixed issue #1358 
It was trying to generate the normal version and plasma version with the same name.

### DIFF
--- a/src/main/java/gregtech/common/MetaFluids.java
+++ b/src/main/java/gregtech/common/MetaFluids.java
@@ -190,8 +190,7 @@ public class MetaFluids {
             BlockFluidBase fluidBlock = new BlockFluidClassic(fluid, net.minecraft.block.material.Material.WATER);
             if(fluidType == FluidType.PLASMA) {
                 fluidBlock.setRegistryName("plasma." + materialName);
-            }
-            if(fluidType == FluidType.NORMAL) {
+            } else {
                 fluidBlock.setRegistryName("fluid." + materialName);
             }
             MetaBlocks.FLUID_BLOCKS.add(fluidBlock);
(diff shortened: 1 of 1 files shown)


src/main/java/gregtech/common/MetaFluids.java
```diff
@@ -163,7 +163,7 @@ private static void setAlternativeFluidName(FluidMaterial material, FluidType fl
         alternativeFluidNames.put(fluidType.getFluidName(material), alternativeName);
     }
 
-    public static Fluid registerFluid(FluidMaterial material, FluidType fluidType, int temperature) {
+    public static Fluid registerFluid(FluidMaterial material, FluidType fluidType, int temperature, boolean isPlasma) {
```
LAGIdiot, line 166: You don't need to pass in additional param as you can already get that information from `fluidType`

LAGIdiot (commented):
I am not sure how we want to handle this so please wait for direction from related bug.

idcppl:
Fixed it, didn't even think about that.

LAGIdiot:
Please update this PR regarding conclusion from connected bug.

You should fix you opening comment to get GitHub correctly link bug that this fixes.
And also please make commit comments descriptive in regard to what they change and don't put useless text to them (LUL).

LAGIdiot:
Are you still working on this PR? Do you need help with anything?

idcppl:
I don't know where you want actually want to go with this, kind of made the PR before actually looking at how to tackle it through suggestions. I'm just going to close this and leave the bug to someone else.

PR #1362, opened 2020-12-21
Check all inputs to the Furnace recipe map

### LINKED ISSUE #1360: [BUG] The Furnace recipe map accepts a list of inputs, but only tests the first item in the list
**Describe the bug**
In `RecipeMapFurnace`, its override of the `findRecipe` method accepts a list of ItemStacks as input, but only searches for a recipe for the first Itemstack in the list. This can cause issues some times, especially in the addon mod Shadows of Greg, in which the Processing Array, which requires machines in the input buses as well as the recipe inputs, can return a null recipe if a set of Electric Furnaces is the first item in the input bus and recipe inputs after. This is because the `findRecipe` method only checks the first entry in `inputs`, which is the Electric Furnace, which will return null for the recipe check.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.8.599
_Modpack:_ none
_Addons:_ SoG 2.15

**Setup**
Playing Solo
_New world generated_ Yes

**Steps To Reproduce**
1) Join a wolrd with SoG and GTCE.
2) Build a Processing Array with 1 input bus and put electric furnaces into the first input slot in the input bus
3) Add ingredients from the GTCE furnace recipe map and see that nothing happens

**Expected behavior**
For the recipe to function, because a correct set of inputs were provided. This can be done by checking all members of inputs, and returning the first existing recipe, or null if none exist.

ALongStringOfNumbers:
**What:**
This PR changes the Furnace recipe map to check all input ItemStacks for a possible recipe, instead of just checking the first item in the input list. This is especially useful for addons that may pass a list to the furnace recipe map with more than one ItemStack, such as is the case with the Processing Array in Shadows of Greg.

**How solved:**
Changed the code to check each input ItemStack for a recipe, returning a constructed recipe for the first non-null recipe found, or else returning null

**Outcome:**
Check all inputs for recipes for the Electric Furnace. Closes #1360.

None Expected

### DIFF
--- a/src/main/java/gregtech/api/recipes/machines/RecipeMapFurnace.java
+++ b/src/main/java/gregtech/api/recipes/machines/RecipeMapFurnace.java
@@ -28,20 +28,15 @@ public class RecipeMapFurnace extends RecipeMap<SimpleRecipeBuilder> {
         for(ItemStack input : inputs) {
             ItemStack output = ModHandler.getSmeltingOutput(input);
 
-            if(output == null) {
-                continue;
-            }
-            else {
+            if(!output.isEmpty()) {
                 return this.recipeBuilder()
                     .inputs(GTUtility.copyAmount(1, input))
                     .outputs(output)
                     .duration(128).EUt(4)
                     .build().getResult();
             }
-
         }
 
         return null;
-
     }
 }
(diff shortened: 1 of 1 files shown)


LAGIdiot (changes requested):
Good work, PR looks solid from functional point. But formatting of new code is terrible. Please use automatic formatter and remove empty which does not separate logic blocks.

ALongStringOfNumbers:
I have cleaned up the formatting slightly, and also changed the check to correctly check `item.isEmpty()`

LAGIdiot (approved):
Thanks for finishing PR.

PR #1378, opened 2020-12-30
Ensure that input stacks don't exceed their maximum stacksize in ItemStack merge logic

ALongStringOfNumbers:
**What:**
This PR adds some defensive measures to the new itemstack merge logic added in #1337 to ensure that the ItemStack passed to the merge method does not exceed its maximum stacksize, or else it breaks down the given ItemStack into several stacks of the maximum stack size and an optional remainder stack if required. It was noticed that this measure was needed because after the new logic was merged, some crashes were noticed in the addon mod SoG, with the Processing Array, because it created an output stack with a count that was greater than the stack's maximum stack size, and then had it expand in output bus of the mutliblock. Under the new itemstack merge logic, this could lead to a crash because there might not be enough room to fully expand the stack in the output bus.

**How solved:**
This issue could have been solved in SoG directly, but I decided to fix it here, since it was a shortcoming in my previous PR that did not consider input stacks that exceeded their maximum stack size. Also, this fix could benefit other addon mods besides just SoG. The fix entails splitting the overlarge Itemstack into a new input list consisting of however many full stacks the overlarge stack flattened into, and a remainder stack if one was required.

**Outcome:**
Add defensive measures to new ItemStack merge logic to deal with input stacks exceeding their maximum stacksize.

None expected, but it should fix some potential crashes in addon mods

### DIFF
--- a/src/main/java/gregtech/api/util/InventoryUtils.java
+++ b/src/main/java/gregtech/api/util/InventoryUtils.java
@@ -52,14 +52,27 @@ public class InventoryUtils {
      * <b>Precondition:</b> the target inventory must actually accept the types of items
      *               you are trying to insert.
      * <br /><br />
-     * @param items the items you want to insert
+     * @param inputItems the items you want to insert
      * @param inventory the target inventory receiving items
      * @return {@code true} if inventory contains sufficient slots to merge and
      *         insert all requested items, {@code false} otherwise.
      */
-    public static boolean simulateItemStackMerge(List<ItemStack> items,
+    public static boolean simulateItemStackMerge(List<ItemStack> inputItems,
                                                  IItemHandler inventory)
     {
+        //First Ensure that the passed list of ItemStacks is actually within its max stacksize
+        List<ItemStack> items = new ArrayList<>();
+        for(ItemStack stack : inputItems) {
+            if(!stack.isEmpty()) {
+                if(stack.getCount() > stack.getMaxStackSize()) {
+                    computeItemStacks(items, stack);
+                }
+                else {
+                    items.add(stack.copy());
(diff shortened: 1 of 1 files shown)


src/main/java/gregtech/api/util/InventoryUtils.java
```diff
@@ -115,4 +128,33 @@ private static void mergeItemStacks(Collection<ItemStack> source, Collection<Ite
...
+            fullStack.setCount(maxCount);
+            collapsedStack.add(fullStack);
+        }
+
+        if(remainder > 0) {
+            ItemStack partialStack = stack.copy();
+            partialStack.setCount(remainder);
+            collapsedStack.add(partialStack);
+        }
+    }
```
LAGIdiot, line 159: I don't like implementation of this. Because you need to pass in collection which gets modified as side effect.
I would rather see this as function taking in "big size" ItemStack and returning ArrayList of "normal size" ItemStacks. Needed to say this should be also handle situation if "normal size" ItemStack or empty one is passed in.

This way it will be more readable (or at least I think so). And you could also use it in case like https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java#L217

LAGIdiot (changes requested):
Good job spotting this. Please see my comment regarding way I would like to see it handled

Exaxxion:
@LAGIdiot for what it's worth, this PR is essentially obviated by #1384.

LAGIdiot:
I am taking #1384 instead of this as it addresses more problems which are in InventoryUtils

PR #1382, opened 2021-01-05
Add JUnit 4 Support

Exaxxion:
**What:**
JUnit is a Java testing framework, nominally for unit tests but also capable of broader testing scopes like integration testing.

Testing frameworks are useful for demonstrating correctness of code under specific conditions, and these tests also double as useful documentation of the intended usage of an API through real examples. It is useful for verifying the conditions that cause a bug and that a proposed fix does actually fix that bug. Regressions become harder to introduce after these tests exist, and make refactoring safer by ensuring they do not change functionality.

**Outcome:**
Properly annotated test classes may be added in the `src/test/` directory tree and will be executed during the `test` build step automatically.

**Additional info:**
Failing tests by design terminate a build. This is useful for preventing the release of regressions or otherwise broken functionality.

For an example of this, remove the `@Ignore` annotation on the test case which is noted to fail due to an existing bug, then execute a build.

None expected.


LAGIdiot (approved):
Thank you for taking your time and creating this PR. Not that you just added unit testing framework but you also provided first example of usage. And it works like a charm.

PR #1384, opened 2021-01-06
Fix Logic Bugs and Improve Utility of Multiblock Voiding Fix Code

Exaxxion:
**What:**
This PR addresses several shortcomings with the multiblock voiding fix code provided via #1333 and revisions thereto in #1337, which were released as part of GTCE 1.10.8.599.

The specific issues are as follows:
1. `ItemStack`s provided to `InventoryUtils#simulateItemStackMerge()` as items to merge, where `ItemStack#getCount()` exceeds `ItemStack#getMaxStackSize()`, fail to be accounted for and result in an exception when the stack passes simulation but cannot actually be merged in subsequent code.
2. NBT data is incorrectly omitted from consideration when comparing `ItemStack`s for equality.
3. `simulateItemStackMerge()` fails if a recipe is defined to produce multiple partial stacks of the same item that could actually fit in the target inventory (example: inserting two stacks of 32 ingots into an empty small wooden chest).

This PR supersedes the proposed fixes from #1378 by @ALongStringOfNumbers, which targeted (1) above. The other two bugs were discovered while I was implementing comprehensive branch coverage tests for `InventoryUtils`.

**How solved:**
Issue (1) is solved by normalizing oversized stacks to a minimal number of valid stacks not exceeding `ItemStack#getMaxStackSize()`.
Issue (2) is solved by additionally ensuring that `ItemStack`s have the same NBT tags to be considered the same item.
Issue (3) is solved by generating compacted copies of the input `ItemStacks` in `simulateItemStackMerge()`.

**Outcome:**
The discovered bugs are corrected and the test suite is updated to contain 100% branch coverage of `InventoryUtils` methods.

**Additional info:**
To facilitate these changes, I have incorporated two additional classes:
1. `ItemStackHashStrategy` designed by @eutropius225, which provides a flexible `ItemStack` hashing strategy for use with FastUtils maps
2. `StreamUtils`, which provides convenience functions for using Minecraft constructs more cleanly with for-each and Stream semantics.

None anticipated, unless addon mods were relying on specific incorrect behavior introduced by these PRs.

### DIFF
--- a/src/main/java/gregtech/api/util/InventoryUtils.java
+++ b/src/main/java/gregtech/api/util/InventoryUtils.java
@@ -90,8 +90,8 @@ public final class InventoryUtils {
      * @return a list containing the resulting ItemStacks.
      */
     static List<ItemStack> compactItemStacks(Collection<ItemStack> inputItems) {
-        Hash.Strategy<ItemStack>      strategy = ItemStackHashStrategy.comparingAllButCount();
-        final Map<ItemStack, Integer> map      = new Object2IntOpenCustomHashMap<>(strategy);
+        Hash.Strategy<ItemStack> strategy = ItemStackHashStrategy.comparingAllButCount();
+        final Map<ItemStack, Integer> map = new Object2IntOpenCustomHashMap<>(strategy);
 
         return inputItems.stream()
 
@@ -190,7 +190,7 @@ public final class InventoryUtils {
 
         final ArrayList<ItemStack> splitStacks = new ArrayList<>();
 
-        int count     = stack.getCount();
+        int count = stack.getCount();
         int numStacks = count / maxCount;
         int remainder = count % maxCount;
 
--- a/src/main/java/gregtech/api/util/InventoryUtils.java
+++ b/src/main/java/gregtech/api/util/InventoryUtils.java
@@ -48,6 +48,10 @@ public class InventoryUtils
      * simulating merging like items into existing stacks, then checking if there
      * are enough empty stacks left to accommodate the remaining items.
      * <br /><br />
+     * <b>Precondition:</b> the target inventory must not virtualize ItemStacks such that
+     *               they can exceed the maximum stackable size of the item as
+     *               defined by {@link net.minecraft.item.ItemStack#getMaxStackSize()}.
+     * <br /><br />
      * <b>Precondition:</b> the target inventory must actually accept the types of items
      *               you are trying to insert.
      * <br /><br />
--- a/src/main/java/gregtech/api/util/InventoryUtils.java
+++ b/src/main/java/gregtech/api/util/InventoryUtils.java
@@ -9,9 +9,10 @@ import java.util.*;
 import java.util.function.Function;
 import java.util.stream.*;
 
+import static gregtech.api.util.Predicates.*;
 import static gregtech.api.util.StreamUtils.*;
 
-public class InventoryUtils {
+public final class InventoryUtils {
     /**
      * @param inventory the target inventory
      * @return the number of empty slots in {@code inventory}
--- a/src/main/java/gregtech/api/util/InventoryUtils.java
+++ b/src/main/java/gregtech/api/util/InventoryUtils.java
@@ -1,25 +1,29 @@
 package gregtech.api.util;
 
+import it.unimi.dsi.fastutil.*;
+import it.unimi.dsi.fastutil.objects.*;
 import net.minecraft.item.*;
 import net.minecraftforge.items.*;
 
 import java.util.*;
+import java.util.function.Function;
+import java.util.stream.*;
 
-public class InventoryUtils {
+import static gregtech.api.util.StreamUtils.*;
 
+public class InventoryUtils
+{
     /**
      * @param inventory the target inventory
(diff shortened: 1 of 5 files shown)


Exaxxion:
Since #1382 has been squash-merged, I'll rebase this PR onto 0ae8688b, dropping the first two commits.

src/main/java/gregtech/api/util/InventoryUtils.java
```diff
@@ -90,29 +131,93 @@ public static boolean simulateItemStackMerge(List<ItemStack> items,
...
+     *
+     * @param stack    a non-empty ItemStack to split into stacks of at most {@code maxCount} items.
+     * @param maxCount the maximum number of items in each stack. Must be a positive, non-zero integer.
+     * @return the resulting zero to many whole stacks and zero to one partial stack.
+     */
+    public static List<ItemStack> apportionStack(ItemStack stack,
+                                                 final int maxCount)
+    {
+        assert !stack.isEmpty();
+        assert maxCount > 0;
```
LAGIdiot, line 199: Please replace these with `if` check and `IllegalArgumentException`

src/main/java/gregtech/api/util/StreamUtils.java
```diff
@@ -0,0 +1,103 @@
...
+import net.minecraftforge.items.*;
+
+import java.util.*;
+import java.util.function.*;
+import java.util.stream.*;
+
+/**
+ * Various Quality-of-Life methods for working with Java 8 Streams.
+ */
+public class StreamUtils
```
LAGIdiot, line 13: I this should be separated to two files. Because mixing highly abstract logic and concrete one in file is hard on readers. Please move ItemsStack specific logic to own file.

src/main/java/gregtech/api/util/InventoryUtils.java
```diff
@@ -90,29 +131,93 @@ public static boolean simulateItemStackMerge(List<ItemStack> items,
...
+
+    /**
+     * Normalizes a potentially problematic ItemStack by splitting it into copied stacks of at most
+     * {@link ItemStack#getMaxStackSize()} items. Empty ItemStacks produce an empty
+     * result list, and stacks already in normalized form are copied as a singleton list.
+     *
+     * @param stack an ItemStack to process.
+     * @return an immutable List of the resulting ItemStacks, in descending size order.
+     */
+    public static List<ItemStack> normalizeItemStack(ItemStack stack)
```
LAGIdiot, line 174: I think it would be worth it even apply this function to all places which could use it as port of this PR.

For example https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java#L217

LAGIdiot (changes requested):
Honestly I am quite impress seeing introduction of some of these concepts. I really like what you are doing here. This looks like it can become core part of GTCE and could save us some time in long run.

I have added some comments to specific parts which needs to be addressed. And for overall PR please fix formatting (possibly with some build in formatter like IntelliJ IDEA's one).

Exaxxion:
> Honestly I am quite impress seeing introduction of some of these concepts. I really like what you are doing here. This looks like it can become core part of GTCE and could save us some time in long run.
> 
> I have added some comments to specific parts which needs to be addressed. And for overall PR please fix formatting (possibly with some build in formatter like IntelliJ IDEA's one).

I am glad that you are largely pleased with the PR, but I am unclear on your standards for formatting. The code as it appears has already been formatted. Is there something in particular that needs to be changed?

I've seen formatting come up in your comments on multiple issues and it seems as though this is a recurring problem. Perhaps formalizing how you want the code formatted with something like [CheckStyle](https://checkstyle.org/) would be useful here.

LAGIdiot:
Fair enough point regarding formatting. I will make issue for myself to come up with some way how to correctly handle it.

For now if you can please run it through IntelliJ IDEA formatter using default Java scheme I would be really glad so I don't have to fix it later.

Exaxxion:
I have reformatted the code to more closely match the rest of the project in 7e41eef.

src/main/java/gregtech/api/util/StreamUtils.java
Exaxxion: Addressed in 1208289

src/main/java/gregtech/api/util/InventoryUtils.java
Exaxxion: Addressed in 894b0cd

LAGIdiot (approved):
Thank you for addressing these issue. This looks really good now.

src/main/java/gregtech/api/util/InventoryUtils.java
Exaxxion: I've been working on adding this change since you requested it, haven't committed that just yet though

LAGIdiot: Let's get this done in another PR so I can get this in current release

Exaxxion: All right, that's fine. I can move it to a new feature branch and PR it in later, then.

PR #1387, opened 2021-01-08
Make InputIngredient implement IIngredient

eutropius225:
**What:**
This PR makes InputIngredient implement CraftTweaker's IIngredient.

**How solved:**
The CountableIngredient is converted to an IIngredient, which is then delegated to by all implemented methods of IIngredient.

Ideally, that IIngredient would be returned directly by methods instead of wrapping with InputIngredient, but that may not be backwards compatible with addon mods.

**Outcome:**
Scripts using InputIngredient can now use it directly as an IIngredient.

For example, scripts copying recipes between recipe maps no longer have to create their own IIngredient wrappers.

Any mods that foolishly reflected the private field will no longer see it.


LAGIdiot (approved):
Thanks for your hard work, this seems like reasonable change which will bring easier usage of this to modpack makers.

PR #1388, opened 2021-01-08
Remove Extra Fuel recipe additions

### LINKED ISSUE #1377: [BUG] Hydrogen is added to the gas generator fuel recipemap twice
**Describe the bug**
Hydrogen is added to the gas generator fuel recipeMap twice and has different power values and durations, see: [Example 1](https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/loaders/recipe/FuelRecipes.java#L27), and [Example 2](https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/loaders/recipe/FuelRecipes.java#L31)

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.8.599
_Modpack:_ None
_Addons:_ None

**Setup**
Playing Solo
_New world generated_ Yes

**Steps To Reproduce**
1) Check the code sections and see that Hydrogen is added twice

**Expected behavior**
For Hydrogen not to be added to the recipemap twice. After checking in game, it appears that the second addition, with 8mb input and 5 ticks duration is the recipe that shows up in game, therefore, the first addition can most likely be removed, as players would not have seen it.

Edit: The same thing happens for Methane, see [Example 1](https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/loaders/recipe/FuelRecipes.java#L28) and [Example 2](https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/loaders/recipe/FuelRecipes.java#L36)

Screenshots: 
The current recipes in the game use the second recipe additions, so I would suggest removing the first ones so the recipes don't change

ALongStringOfNumbers:
**What:**
This PR removes some extra additions of Hydrogen and Methane to the Gas Generator Fuel RecipeMap. These additions did not show up in game, as the additions that came after them were always the ones that showed up. Therefore, these additions were removed to not change any recipes in game.

**How solved:**
Removed the extraneous additions

**Outcome:**
Removed extraneous Hydrogen and Methane additions to the Gas Generator. Closes #1377.

**Additional info:**
Recipes in game before change:

Recipes in game after change:

None expected. If an addon mod was removing this recipe, then at most the recipe removal will fail, which would not crash.

### DIFF
--- a/src/main/java/gregtech/loaders/recipe/FuelRecipes.java
+++ b/src/main/java/gregtech/loaders/recipe/FuelRecipes.java
@@ -24,9 +24,6 @@ public class FuelRecipes {
         registerSteamGeneratorFuel(Materials.Steam.getFluid(60), 1, GTValues.LV);
 
         //low-tier gas turbine fuels
-        registerGasGeneratorFuel(Materials.Hydrogen.getFluid(1), 1, GTValues.LV);
-        registerGasGeneratorFuel(Materials.Methane.getFluid(1), 2, GTValues.LV);
-
         registerGasGeneratorFuel(Materials.NaturalGas.getFluid(8), 5, GTValues.LV);
         registerGasGeneratorFuel(Materials.Hydrogen.getFluid(8), 5, GTValues.LV);
         registerGasGeneratorFuel(Materials.CarbonMonoxde.getFluid(8), 6, GTValues.LV);


LAGIdiot (approved):
Good job fixing this issue.

PR #1394, opened 2021-01-15
Bug fix for ItemHandlerList::extractItem

solidDoWant:
**What:**
Mistake in ItemHandlerList::extractItem cases 'slot' number of items to be extracted instead of 'amount'. This PR fixes that mistake.

**How solved:**
Swapped 'slot' for 'amount' in the 'itemHandler.extractItem' method call.

**Outcome:**
Fixed extraction amount bug in ItemHandlerList::extractItem.

I am about 90% sure that this method is not called anywhere in the base mod but it may be by addons. My limited testing didn't show any issues but others could potentially arise if this method is used elsewhere.


LAGIdiot (approved):
Thank you for spotting and fixing this mistake. It seems that it is not used at all se were not aware of this problem but better to have it fixed.

PR #1395, opened 2021-01-15
Add Better TOP Info for Transformers

DStrand1:
**What:**
This PR adds more helpful information to the TOP display when looking at a Transformer. Currently, transformers only display the normal energy bar that every machine displays. I brought this idea up in the Discord, and was inspired to create a PR for it after many people said it seemed like a good idea to add to GTCE.

**How solved:**
I added some code that checks if the machine is a transformer, and if it is, it displays if the transformer is in step-up, or in step-down mode. Additionally, I added a check for if the player is looking at an input or output face, and it displays that information, along with the voltage and amperage, accordingly.

**Outcome:**
Added better TOP info for transformers.

**Additional info:**
Some examples of possible scenarios when looking at an I/O side of a transformer:
stepdown-input
stepup-output
This was inspired by how GT5U handles displaying its transformer data in Waila, as shown below:
stepdown

Gets voltage tier names from the GTValues class, so addons that add higher-tier transformers will most likely not have this extra TOP information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 ## Changelog
 
+### 1.10.9
+* Updated Fuel recipe map to be similar to recipe map (#1353) - idcppl
+* Updated Furnace recipe map to check all inputs (#1362) - ALongStringOfNumbers
+* Removed Extra Fuel recipe additions (#1388) - ALongStringOfNumbers
+* Fixed incorrect Lapotron Crystal crafting recipe (#1348) - ALongStringOfNumbers
+* Fixed the Primitive Blast Furnace Voiding outputs (#1351) - ALongStringOfNumbers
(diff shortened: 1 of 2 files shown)


LAGIdiot (changes requested):
Thanks for your hard work on this PR.

I think you should create whole new provider for this functionality which will be specific for Transformers. Right now you are putting specific logic to file which is designed for general use.

DStrand1:
I changed some things around to pull the specific Transformer code out into a new Provider, called AdditionalInfoProvider. This allows us to easily just add an if-block, checking if a MetaTileEntity is a specific machine, and can add more TOP info there with very little work needing to be done. If this change is not how you would like the code organized, I can easily change it to be specifically for the Transformer, and make one more commit changing it to that. Behavior is identical now as it was in the initial commit.

PR #1396, opened 2021-01-15
Add GT5u and GTNH Electric Blast Furnace coil bonuses

TechLord22:
**What:**
This PR adds the Electric Blast Furnace Coil bonuses from GT5u and GTNH. 
GT5u:
1) Every 900K above the recipe's temperature gives a 5% energy discount after overclocking. 
2) Every 1800K above the recipe's temperature will overclock the recipe once with a 75% speed increase, if the EBF is capable of overclocking.

GTNH:
1) EBF temperature gets bonuses of 100K according to the formula: `100 * (tier - 2)`, where tier is the corresponding voltage tier (LV = 1, MV = 2, etc.). The EBF also displays the Bonus Temperature above the potentially changed Max Temperature in the multiblock's GUI. 

Each of the values above can be configured in the config, and the GT5u and GTNH feature groups can be disabled independently of one another. The default values are the same as the GT5u and GTNH defaults. The EBF has a tooltip explaining each of the three features, which only display if the specific feature is enabled. The tooltip also changes to reflect the values specified in the config.

**How solved:**
Added code for each feature to the Electric Blast Furnace. The GT5u features are implemented by making changes to recipes before they are run in the EBF. The GTNH features are implemented by modifying the EBF temperature on structure form. 

**Outcome:**
Adds EBF energy discounts and recipe speedups according to coil temperature. Also adds EBF temperature bonuses according to multiblock voltage tier. 

**Additional info:**
An image of the new EBF tooltip without GT5u Features:
image

An image of the new EBF tooltip without GTNH Features:
image

An image of the new EBF tooltip with both GT5u and GTNH Features:
image

An image of the EBF tooltip with no new features:
image

An image of the modified EBF controller GUI with GTNH features:
image

An image of the modified EBF controller GUI with no features:
image

My monitor is unusually large, requiring Auto GUI scale, so tooltip appearance will look better for other users. 

Some recipes will be possible at earlier coil tiers when using the GTNH feature, even at default values. Only one recipe could potentially interfere with progression: Hot Iridium - producable at kanthal coils instead of nichrome. This allows the player to make the nano saber and energy field projector slightly earlier, however nichrome is immediately achievable post kanthal, so there isn't much issue.


LAGIdiot:
Thank you for your hard work on this PR.

I look through it and I am not sure if this is something that should be included in GTCE as it does not fit with it's current vision. Especially that GTNH part I don't think that by putting more energy in will heat it up better I would expect coils to melt first as they are not design to take in that much heat (and that is exactly reason you need to build new more powerful EBFs). Also mixing more then one overclocking mechanics seems like overkill which will be hard to explain and use.

I will leave this open for discussion so if anyone has something to add please do it. And I will take a look at it later.

LAGIdiot:
Based on discussion for this PR which had 0 comments after mine opening one I come to conclusion that this will not be included in GTCE as for reasons mentioned above.

PR #1404, opened 2021-01-19
Add Better TOP Transformer Info

DStrand1:
Linking Pull Request #1395 as I screwed up my git hierarchy pretty bad, and force pushed my branch linked to that PR back to origin/master. Here are my changes since your last review:

I changed some things around to pull the specific Transformer code out into a new Provider, called AdditionalInfoProvider. This allows us to easily just add an if-block, checking if a MetaTileEntity is a specific machine, and can add more TOP info there with very little work needing to be done. If this change is not how you would like the code organized, I can easily change it to be specifically for the Transformer, and make one more commit changing it to that. Behavior is identical now as it was in the initial commit.


LAGIdiot (changes requested):
I am not fond of idea to have `AdditionalInfoProvider` which will have mix of different things in it. I would rather see it as `TransformerInfoProvider` which will handle only that specific machine type. This way it will be on first look obvious what it does. So please change it.

Also I see some unnecessary changes regarding file endings.

DStrand1:
Alright, should only take me a moment

DStrand1:
Changed to simply "TransformerInfoProvider"

LAGIdiot (approved):
Thanks for finishing this PR.

PR #1405, opened 2021-01-19
Add Cable Recipes and 4x Cable/Wire Recipes

DStrand1:
**What:**
Currently, there are no recipes being created to combine or deconstruct cables. For example, crafting two 1x cables into a 2x cable, or crafting one 2x cable into two 1x cables. Additionally, you can currently only do cable and wire recipes like this in pairs. I added recipes to be able to combine four 1x cables/wires into a 4x cable/wire, and same with four 2x and four 4x.

**How solved:**
Added some simple code to the WireRecipeHandler class to allow you to do this.

**Outcome:**
Cables are now much more easily used, and can be simply crafted or uncrafted into whichever size you need at the moment.

**Additional info:**
Quick example of one of the added recipes:
cable

None expected.


LAGIdiot (approved):
Thank you for your hard work this seems like reasonable useful addition.

PR #1406, opened 2021-01-19
Fix Positive Crafting Loops in Chemical Recipes

### LINKED ISSUE #1318: [BUG] Phosphorus positive Crafting loop exists via Posphorus Pentoxide
**Describe the bug**
A Phosphorus positive loop exists via making phosphorus pentoxide, allowing for infinite phosphorus to be generated with the addition of water and oxygen.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.5.582
_Modpack:_ None
_Addons:_ None

**Setup**
Playing Solo
_New world generated_ Yes

**Steps To Reproduce**
The Recipe chain is:

1) 4 Phosphorus dust + 10B Oxygen in a Chemical Reactor -> 14 Phosphorus Pentoxide Dust
2) 1 Phosphorus Pentoxide dust + 6B water in a chemical reactor -> 4B Phosphoric Acid
3) 7B Phosphoric Acid in an Electrolyzer -> 4 Phosphorus dust and 3B Hydrogen.

This is a recipe loop, as you get much more Phosphorus via Phosphoric Acid then you put in.

**Expected behavior**
No recipe loop

**Additional context**
An easy solution would be to disable the decomposition of Phosphoric Acid, or remove the Phosphoric Acid from Phosphoric Pentoxide recipe in the chemical reactor.

### LINKED ISSUE #1329: [BUG] Methane Positive crafting loop possible
**Describe the bug**
A Methane positive crafting loop exists, such that with the addition of solely water, infinite methane and oxygen can be generated.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.7.594
_Modpack:_ None
_Addons:_ None

**Setup**
Playing Solo
_New world generated_ Yes

**Steps To Reproduce**
1) Make Methane in a Chemical Reactor via the recipe: 1 Carbon dust + IC:0 + 4k mB of Hydrogen -> 5k mB Methane
2) In a Chemical Reactor, use the recipe: IC:1 + 1k mB Methane + 2k mB Water -> 8k mB Hydrogen + 1k mB CO2.
3) In an Electrolyzer, break down the CO2 -> 1 Carbon dust + 2k mB Oxygen.
4) A methane and oxygen positive infinite loop presents itself due to the input amount of Methane in step 2 and the output amount of Hydrogen in step 2

**Expected behavior**
No positive infinite loop for methane and oxygen should exist. Solving this would most likely involve adjusting the Chemical Reactor recipe present in Step 2 outlined above or by adjusting the Methane output in the recipe outlined in Step 1.

DStrand1:
**What:**
Fixes positive crafting loops demonstrated in issues #1329 and #1318 

**How solved:**
For #1318: I simply removed the Phosphorus Pentoxide + Water -> Phosphoric Acid recipe, as was the conclusion decided on in the issue thread.

For #1329: I rebalanced the recipe for Methane.
The chemical formula used here was: C + 4H => CH₄. However, in our recipes, it was producing 5k mb of Methane, meaning that the actual formula implemented was: C + 4H => 5CH₄. Obviously, this does not make sense from a chemistry standpoint, as 4 Carbon and 16 Hydrogen are being created from thin air. Unfortunately, this does make the production of Methane slightly more difficult from a player standpoint, but I think it is important to respect the chemistry and follow our rules of 1k mb being 1mol of a fluid.

**Outcome:**
Closes #1318 
Closes #1329 

Only issues seen are if someone is depending on the production of Phosphoric Acid using Phosphorus Pentoxide, but since this is not the only (nor primary) way of producing this resource, I don't see this being a significant issue.


DStrand1:
I readded the Phosphorus Pentoxide recipe. We are using the chemical formula: C4O10 + 6H2O => 4(H3PO4), but only had 1 phosphorus pentoxide dust in the recipe. I changed it to take the full 14, or one full mol, so that the equation is balanced. There does seem to be a strange problem where the electrolysis recipe for Phosphoric Acid (H3PO4) does not make any sense chemically, but it does not lead to a positive loop, so it should be "okay" to stay as-is.

DStrand1:
Closing this as it will be addressed in a different PR, in a more robust way

PR #1408, opened 2021-01-19
Rotor Mold

DStrand1:
**What:**
Currently, Rotors are one of the most tedious microcomponents to make, and as a result, pumps are even more tedious. This PR addresses this, by adding a new Solidifying Mold for Rotors.

**How solved:**
I added a Fluid Solidifier Mold that makes a rotor with 4 ingots of the material. As a result, I had to change a few things around to avoid duplication exploits. The total material for Rotors in OrePrefix was changed from `M * 4 + M / 4` to simply `M * 4`. As a result, Arc furnace, macerator, and fluid extractor recipes now give 4 of the material, rather than 4 and a quarter. Additionally, I removed the now outdated assembling machine recipe for rotors, that takes 4 plates, a ring, and soldering alloy/tin. Lastly, I went ahead and made the stack size 64, rather than 32.

The reason I chose 4 ingots, is because I either had to choose 4 or 5. Currently, the hand recipe at its best is 4 plates, a screw, and a ring, being 4.375 material. The old assembler recipe is 4 plates and a ring, plus some soldering/tin, being 4.25 material. 4 ingots seemed like the best place to put this, as it saves only a little more material than the old machine recipe, and still has benefits over the hand recipe.

**Outcome:**
Rotor Mold added to make rotors easier to make.

**Additional info:**
Couldn't show tooltip for fluid, but all are 576mb (4 ingots)
moldrecipe
moldcraft
This also has a Forming Press recipe that is automatically generated along with the other molds.

Players who have rotors automated with a system like AE2 or RS could have recipes that do not work in their system. But, 
archquote


smudgerox:
a mold for fluid solidifying would be better imo.

 I don’t really see how you could extrude a rotor considering the shape, but a mold would work perfectly in this case for a touch more realism. 

agree with the idea overall, though.

DStrand1:
I did it this way since this is how GT5 did it. Also, I like the idea of it being even ingot amounts, and this way is much easier to automate in the long run.

Archengius:
Fluid solidifying recipe can be added additionally, I don't see how it prevents recipe from using even ingot amounts.
Also, "I'm doing it this way because GT5 did in that way" reasoning does not work, we are not GT5, we are GTCE.

DStrand1:
Should we have both options then, extruder and fluid solidifier?

Archengius:
Sounds like a good idea to me.

DStrand1:
There is now a Mold and an Extruder Shape for Rotors, both using 4 ingots of material

LAGIdiot:
This overall seems good. Please updated your opening comment with information regarding Mold Shape and provide pictures for crafting of Mold and Extruder Shape.

Regarding removal of Assembly recipe. I am OK with it but this needs to be documented in change log (done on release) and should be released at least in Minor version.

DStrand1:
Updated screenshots in opening comment.

smudgerox:
I very much appreciate you taking my suggestion on board. 

That said, adding it with that solidification recipe time makes the extruder recipe obsolete before you even get to it. There would be no reason to use it, only people who don't know any better would. 

I'm not really sure what you would want to do about that, might want to increase the solidification time, or just remove the extruder portion entirely.

DStrand1:
Good catch. I was lazy when adding the Mold, and didn't modify its EUt and duration from the source I copy-pasted from (plate). I adjusted the EUt and duration of the Mold to be the same as the Extruder, allowing the player to choose which option they want to use.

idcppl:
> I very much appreciate you taking my suggestion on board.
> 
> That said, adding it with that solidification recipe time makes the extruder recipe obsolete before you even get to it. There would be no reason to use it, only people who don't know any better would.
> 
> I'm not really sure what you would want to do about that, might want to increase the solidification time, or just remove the extruder portion entirely.

It takes 4 seconds for a fluid extractor recipe, so it should be fine at 2 seconds. Keeping materials in fluid form can be considered inefficient time wise.

Exaxxion:
Extruders push a material blank through a die to create a specific-shaped length of material, which is usually then sliced to produce final products which have an identical cross-section.

This makes sense for elongated continuous shapes like wires and pipes, and excusably extrapolates to flat objects like gears (since it's extra complexity to have a gear-shaped tube item you then have to slice), but rotors have curved blades to scoop fluid. I concur with @smudgerox that casting liquid in a mold makes far more sense for this particular case.

DStrand1:
I agree that an extruder recipe for this doesn't make the most sense, but if we are removing the old assembler recipe for rotors, I feel like locking the only automatable recipe for rotors to a Fluid Solidifier is not a great solution. I believe for this reason alone we should have both, but if others do not feel like the Fluid Solidifier recipe is too difficult, I can remove the Extruder recipe.

Archengius:
Fluid solidifier recipe is alright, no need to duplicate it in extruder.

DStrand1:
Alright. I removed the extruder shape and left only the mold.

LAGIdiot (approved):
After all these updates it still seems reasonable so it is accepted.

But please again update opening comment regarding what really this PR does.

DStrand1:
Updated opening comment.

Issue #1412, opened 2021-01-20
[BUG] Multiple Fluid Extractor recipes for seeds.


thecodingchicken:
**Describe the bug**
7 Recipes exist in the fluid extractor to make seed oil.  
Currently, two recipes exist that make 10mb from <minecraft:wheat_seeds>, <minecraft:melon_seeds> give both 10mb and 3mb, and <minecraft:pumpkin_seeds> give both 10mb and 6mb.
The one other vanilla seed in the list, <minecraft:beetroot_seeds> give you 10mb.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.109.612
_Modpack:_ GTCE Testing Environment, the only other mods added are JEI(4.16.1.301) and CodeChicken Lib(3.2.3.358)
_Addons:_ None

**Setup**
Playing Solo
_New world generated_

**Steps To Reproduce**
1) Lookup the recipes for seed oil in jei, and multiple recipes that use said seeds can be seen.  

**Expected behavior**
4 recipes for seed oil should exist, given the 4 vanilla seeds being used.  

**Screenshots**
![Wheat seeds](https://user-images.githubusercontent.com/29822509/105119942-493d8300-5a9f-11eb-91d3-78451c9f52a2.png)
![Melon seeds](https://user-images.githubusercontent.com/29822509/105119966-55c1db80-5a9f-11eb-9b3c-5212bc8b5d90.png)
![Pumpkin Seeds](https://user-images.githubusercontent.com/29822509/105120017-70945000-5a9f-11eb-8648-6b1d91508d2b.png)

LAGIdiot:
Accepted, based on provided description and PR.

PR #1413, opened 2021-01-20
Fixing duplicate recipes for seed oil

thecodingchicken:
**What:**
Remove Recipes for Wheat seeds, Melon seeds, and Pumpkin seeds, fixes issue #1412 

**How solved:**
Wheat seeds, Melon seeds, and Pumpkin seeds are all covered in GTUtility.getGrassSeedEntries(), however, Beetroot seeds are not.  So, by removing the 3 recipes for those seeds, a recipe for it will not be built twice.  

**Outcome:**
Removed Duplicate Seed Oil recipes from Fluid Extractor
Fixes: #1412 

**Additional info:**
Old recipes:
image
image
image
image
New recipes:
image
image

None, as this only removes duplicate recipes.

### DIFF
--- a/src/main/java/gregtech/loaders/recipe/ChemistryRecipes.java
+++ b/src/main/java/gregtech/loaders/recipe/ChemistryRecipes.java
@@ -262,9 +262,6 @@ public class ChemistryRecipes {
                 .buildAndRegister();
         }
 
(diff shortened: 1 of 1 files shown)


LAGIdiot (approved):
Thanks for spotting this and fixing it.

PR #1415, opened 2021-01-22
Fix Tooltip Bounding Box

### LINKED ISSUE #1347: [BUG] Overlapping tooltips in machine GUI
**Describe the bug**
It is possible to get overlapping tooltips in a machine GUI. Note: this was only tested with the HV Electrolyzer, but it could happen in multiple machines.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.7.594
_Modpack:_ Omnifactory, on development branch
_Addons:_ SoG: 2.15

**Setup**
Playing Solo
_New world generated_ No

**Steps To Reproduce**
1) Place down an HV eletrolyzer and put something in the item input slot
2) Mouse over the item input and then slowly move your mouse towards the fluid input slot.
3) Near the dividing line between the two slots, it can be seen that the tooltips overlap

**Expected behavior**
For the Tooltips to not overlap

**Screenshots**

DStrand1:
**What:**
This PR fixes the bounding box issue for `isMouseOver()`, causing tooltips to overlap when a fluid slot and an item slot are directly next to each other.

**How solved:**
The fix was a simple off-by-one error, and all I had to change was a check that was previously `>=`, but should have simply been `>`.

**Outcome:**
Closes #1347 

**Additional info:**
Tooltip of item shown when hovering over item slot:
itemhover

Tooltip of fluid slot, and only fluid slot, shown when mouse is moved 1 pixel to the left:
fluidhover

(Don't mind the missing textures in my inventory, they are left over from my Rotor extruder shape/mold work from a different PR)

None expected.

### DIFF
--- a/src/main/java/gregtech/api/gui/Widget.java
+++ b/src/main/java/gregtech/api/gui/Widget.java
@@ -120,7 +120,7 @@ public abstract class Widget {
     }
 
     public static boolean isMouseOver(int x, int y, int width, int height, int mouseX, int mouseY) {
-        return mouseX >= x && mouseY >= y && x + width >= mouseX && y + height >= mouseY;
+        return mouseX >= x && mouseY >= y && x + width > mouseX && y + height > mouseY;
     }
 
     /**


LAGIdiot (approved):
Thanks for your hard work

PR #1416, opened 2021-01-22
Add "Regular" to dust oredict entries

### LINKED ISSUE #1315: Make normal dust sortable with ore dict filter with wildcard
I'm trying to sort my dusts in an ore processing chain.

I can sort tiny dust and small dust using wildcard symbols like dustSmall* and so on in the ore dict filter.
The problem is sorting normal dusts, is impossible since they have no "prefix".
Writing dust* moves every type of dust, also tiny, small, crushed and so on.

A solution could be to name normal dust something like dustNormalRuby instead of dustRuby.

DStrand1:
**What:**
This PR addresses issue #1315 and adds an oredict entry to full piles of dust, "dustRegular<material>". I did not remove the other oredict, as that would introduce compatibility issues, so normal dusts have 2 oredict entries.

**How solved:**
I created a generic oredict `registerOre()` method that creates an oredict from a String rather than OrePrefix. I also adjusted the OrePrefix `registerOre()` method to call the String one, as to not have duplicate code, and to maintain compatibility in the code. While there is already a `registerOre()` method that uses a String as the oredict prefix, it does not support using materials, so I decided this route was the best implementation.

**Outcome:**
Closes #1315

**Additional info:**
Chest of input items:
sampleinventory

Conveyor settings with Oredict filter:
conveyorsettings

Chest with output items, showing only the regular dust:
exportinv

None expected.

### DIFF
--- a/src/main/java/gregtech/api/items/materialitem/MaterialMetaItem.java
+++ b/src/main/java/gregtech/api/items/materialitem/MaterialMetaItem.java
@@ -54,8 +54,9 @@ public class MaterialMetaItem extends StandardMetaItem {
             OrePrefix prefix = this.orePrefixes[metaItem / 1000];
             Material material = Material.MATERIAL_REGISTRY.getObjectById(metaItem % 1000);
             OreDictUnifier.registerOre(new ItemStack(this, 1, metaItem), prefix, material);
+
             if(prefix.name().equals("dust"))
-                OreDictUnifier.registerOre(new ItemStack(this, 1, metaItem), prefix.name() + "Regular", material);
+                OreDictUnifier.registerOre(new ItemStack(this, 1, metaItem), "dustRegular", material);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/gregtech/api/items/materialitem/MaterialMetaItem.java
```diff
@@ -54,6 +54,8 @@ public void registerOreDict() {
             OrePrefix prefix = this.orePrefixes[metaItem / 1000];
             Material material = Material.MATERIAL_REGISTRY.getObjectById(metaItem % 1000);
             OreDictUnifier.registerOre(new ItemStack(this, 1, metaItem), prefix, material);
+            if(prefix.name().equals("dust"))
+                OreDictUnifier.registerOre(new ItemStack(this, 1, metaItem), prefix.name() + "Regular", material);
```
LAGIdiot, line 58: I would rather see it like this:
```suggestion

            if(prefix.name().equals("dust"))
                OreDictUnifier.registerOre(new ItemStack(this, 1, metaItem), "dustRegular", material);
```
As it is more obvious on first look

LAGIdiot (changes requested):
Good work on this PR.

I have added one comment which will enchant readability of given solution. You can merge it from web or I will do it on merging day.

LAGIdiot (approved):

Issue #1152, opened 2020-07-17
Ore generation height measurement is bugged


talchas:
In CachedGridEntry findOptimalSpot returns the offset within a chunk that is closest to the the center of the grid, but the return value is used as if it is a global coordinate, not a chunk-relative one. This means that masterEntry is always set to the heights of one of 0,0/15,0/etc, not the height of a block in the chunk in question. This measurement is then used by doGenerateVein to completely eliminate veins early based on this height vs the vein's min_height and size. The most obvious result is that if the 0,0 chunk is deep ocean, any ore with too high of a min_height will never spawn anywhere in the world. With the default settings tin, tetrahedrite, salts, and several others can get hit by this fairly easily.

This bug is easy to fix, just add the missing chunk offset before asking the world for the height.

Even with this fixed there's still a lesser problem, which is that one height measurement is used to abort the entire vein generation, and this height measurement does not even come from inside of the grid cell that the vein is centered in. It usually comes from whichever chunk in whichever _neighboring_ grid cell happens to be generated first, which then triggers CachedGridEntry creation of all its neighbors, but using the current chunk to guess the height. The height/GTWorldGenCapability isn't saved in any of the grid's chunks, so it will get replaced, but the CachedGridEntry and its veinGeneratedMap will remain with these first values until the server restarts.

So for anything next to ocean whether or not a vein gets generated can depend on what direction you are exploring from, since that determines the order chunks are generated in. If you were really careful, you could probably generate some chunks with one veinGeneratedMap, and then restart the server, go back there from a different direction and have a different veinGeneratedMap for generating other chunks in the same grid cell (The list of possible veins for the cell will remain the same, but the early exit of the first vein would also cause the location of any other veins to shift, since there are many more rng values consumed after the abort point). 

Fixing this bug without just removing the early abort seems difficult, and that would increase the cost of generating grid cells where veins legitimately fail. Of course, at the moment due to the first bug, worlds either always have them fail or never, so it would be no worse than the current situation in worlds that aren't broken, and it isn't too slow now.

LAGIdiot:
Thank you for reporting this problem to us. And providing us with all that data. This seems as serious problem and should be investigated soon. I will mark this as unverified until we have data from our side to prove your claims.

Saereth:
[...]

PR #1418, opened 2021-01-22
Make oregen check local chunk coordinates instead of origin corners

ALongStringOfNumbers:
**What:**
This PR adjusts the oregen to check the height of the chunk in the local chunk coordinates, instead of always at the defined origin chunk corners. In doing so, this PR addresses the first issue mentioned in #1152.

Previously, the `findOptimalSpot` method would check which corner of the chunk was closest to the center of the defined chunk grid. When doing so, it would add the chunk's coordinates to the defined origin chunks, see `double diffX = (chunkBaseX + pos.getX()) - gridCenterX;`, and would use this to check which corner is closest to the center of the chunk grid. However, when returning the result, it would instead just return the `mostClosePos`, which was one of these defined origin chunks, and not take into account the offset used when finding the corner closest to the center of the chunk grid.

After exiting the `findOptimalSpot` method, the height of returned position would then be gathered, however since the returned pos was one of the defined origin corners, the height would always be the height of the chunk at 0,0, instead of the local chunk where the vein was trying to generate. This means that there could be some issues trying to generate ore veins if the origin chunk had a very low height, such as a deep ocean biome as mentioned in #1152.

**How solved:**
This PR simply adds the chunk offset to the returned value of the defined origin corners, so that when the height of the chunk is checked, the height is checked at the chunk, instead of at the origin chunk

**Outcome:**
Fixes an issue where oregen for some veins could fail if the 0,0 chunk had extremely low height.

None expected

### DIFF
--- a/src/main/java/gregtech/api/worldgen/generator/CachedGridEntry.java
+++ b/src/main/java/gregtech/api/worldgen/generator/CachedGridEntry.java
@@ -94,6 +94,7 @@ public class CachedGridEntry implements GridEntryInfo, IBlockGeneratorAccess, IB
         if (masterEntry == null) {
             Chunk primerChunk = world.getChunkFromChunkCoords(primerChunkX, primerChunkZ);
             BlockPos heightSpot = findOptimalSpot(gridX, gridZ, primerChunkX, primerChunkZ);
+            heightSpot = heightSpot.add(primerChunkX * 16, 0, primerChunkZ * 16);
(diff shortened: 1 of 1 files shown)


LAGIdiot (approved):
Thanks for fixing this issue.

Now that I see it implemented and I read through provided explanation it seems as just dumb mistake. Not big implementation problem I imagine it would be.

PR #1420, opened 2021-01-24
Fixes #1365

### LINKED ISSUE #1365: [BUG] Surface Rocks can generate in (under) water
unknown (1)
unknown

Rongmario:
**What:**
Fixes #1365 - where surface rocks would generate in pools of water (same level of the liquid)

**How solved:**
Now places surface rocks only when the most block is air, and if the block underneath has a solid top face.

**Possible Considerations**
Do discuss if `isReplaceable` should be checked resulting where surface rocks could replace the likes of grass and whatnot.


LAGIdiot (approved):
Thanks for your hard work.

We can iterate on replacing grass and other things later. For now let's go with this.

Issue #1401, opened 2021-01-17
Missing Textures


Rundas01:
stargate liquids have no texture when placed in the world

**Versions**
Forge: 14.23.5.2854
GTCE: 1.10.9.612

**Setup**
Playing Both

**Steps To Reproduce**
1) get a bucket of a stargate liquid
2) place it in the world

**Expected behavior**
The liquid should have a texture

**Screenshots**
Is attached
![Bild_2021-01-17_173847](https://user-images.githubusercontent.com/38040866/104849604-dbe8f100-58ea-11eb-94dd-ca935330d099.png)

ALongStringOfNumbers:
This sounds similar to #1085

LAGIdiot:
This is same problem as #1085 in regards how it behaves.

PR #1421, opened 2021-01-24
Fixes #1085

### LINKED ISSUE #1085: Potion liquids textures missing when pouring on the ground
As the title says, if I pour a bucket filled by potion like Lucky Potion on the ground, it becomes purple and black texture. This problem has existed since last version.

Rongmario:
**What:**
Fixes potion fluids not having textures when placed.

**How solved:**
Delegated registering fluid blocks to after potions fluids have initialized.

**Outcome:**
They now have textures when placed.
Fixes: #1085 

**Additional info:**
There was some talks on discord about changing potion fluids and how they inherently work.

Most common way being 3 fluids are registered, `normal`, `splash` and `lingering`. And brewing recipes are done via FluidStack differentiation (`PotionType` nbt stuff: see `PotionUtils`). I also haven't checked this commit against #1401 as that may be something different?

### DIFF
--- a/src/main/java/gregtech/common/CommonProxy.java
+++ b/src/main/java/gregtech/common/CommonProxy.java
@@ -81,13 +81,13 @@ public class CommonProxy {
         SURFACE_ROCKS.values().stream().distinct().forEach(registry::register);
         FRAMES.values().stream().distinct().forEach(registry::register);
         ORES.forEach(registry::register);
-        FLUID_BLOCKS.forEach(registry::register);
     }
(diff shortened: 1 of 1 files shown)


LAGIdiot (approved):
Thanks for your hard work.

I will take this as fix for potion textures. Other problems can be handled later. Also thanks for inside regarding Potion registration.

LAGIdiot:
I have fixed missing close link for issue in opening comment.

PR #1422, opened 2021-01-25
Fix Turbine Running When It Shouldn't

### LINKED ISSUE #1367: [BUG] The turbine increases the speed without rotor
**Describe the bug**
The turbine continues to execute the recipe, without the rotor.
The bug works in solo and on the server

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.8.599
CodeChickenLib: 1.12.2-3.2.3.358-universal
CraftTweaker2: 1.12-4.1.20.586

**Steps To Reproduce**
1) Build a plasma turbine (this is clearly visible on it, because of the duration of the recipe)
2) Run the recipe
3) Pull out the rotor

**Screenshots**
image

Took the holder while retaining the NBT
image

Took the same holder after about 10 seconds
image

**Additional context**
The turbine does not generate power, without a rotor.

Moreover, the controller remains active after the rotor is removed.
In addition, when you stop the process with a soft-hammer, the work stops, but the controller texture remains active.

It seems to me that the problem is with the controller than with the rotor holder. Controller activity is the cause, and continued overclocking is the consequence.

DStrand1:
**What:**
Fixes the bug demonstrated by Issue #1367 
Additionally fixes the bug that when the Turbine multiblock invalidated, the rotor speed is not reset.

**How solved:**
I added a check to the RotorHolder to check if it has a rotor in its inventory, and reset speed and disable controller if not. I also had to add a handful of checks around to get all the edge cases ironed out, since `FuelRecipeLogic` was not able to properly check for the secondary condition of having a rotor present, so I had to override it. There was also an issue where setting the controller to inactive didn't reset the current recipe's processing time. I couldn't set the multiblock to inactive in the rotor holder, since there was a problem where `MetaTileEntityMultiblockPart::getController()` was returning null on world load, causing the turbine speed to be reset on a relog.

**Outcome:**
Closes #1367

**Additional info:**
This code brought up an edge case where when the Controller block itself was broken, the rotor speed was not reset since the lower level class `MultiblockControllerBase` was not calling `invalidateStructure()` directly, but instead simply removing each part from the multiblock. I'm not sure of the reasoning of this, but if this fix is reverting a purposeful design decision, it can be reverted. I left the old code in a comment in that file. If reverted, it would only mean that the rotor speed wouldn't be reset when specifically the controller was broken, but it may have other implications.

None expected.

### DIFF
--- a/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
@@ -85,15 +85,11 @@ public class FuelRecipeLogic extends MTETrait implements IControllable {
         }
     }
 
-    public void clearRecipe() {
-        recipeDurationLeft = 0;
-    }
-
     protected boolean shouldVoidExcessiveEnergy() {
         return false;
     }
 
-    private void tryAcquireNewRecipe() {
+    protected void tryAcquireNewRecipe() {
         IMultipleTankHandler fluidTanks = this.fluidTank.get();
         for (IFluidTank fluidTank : fluidTanks) {
             FluidStack tankContents = fluidTank.getFluid();
@@ -111,7 +107,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable {
         return isActive;
     }
 
-    private int tryAcquireNewRecipe(FluidStack fluidStack) {
+    protected int tryAcquireNewRecipe(FluidStack fluidStack) {
         FuelRecipe currentRecipe;
         if (previousRecipe != null && previousRecipe.matches(getMaxVoltage(), fluidStack)) {
             //if previous recipe still matches inputs, try to use it
@@ -174,6 +170,10 @@ public class FuelRecipeLogic extends MTETrait implements IControllable {
     }
 
     protected void setActive(boolean active) {
+        // If we are changing states (active -> inactive), clear remaining recipe duration
+        if (this.isActive && !active && recipeDurationLeft != 0) {
+            recipeDurationLeft = 0;
+        }
         this.isActive = active;
         if (!metaTileEntity.getWorld().isRemote) {
             metaTileEntity.markDirty();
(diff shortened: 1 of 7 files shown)


src/main/java/gregtech/common/metatileentities/multi/electric/generator/MetaTileEntityLargeTurbine.java
```diff
@@ -86,6 +86,7 @@ protected void formStructure(PatternMatchContext context) {
 
     @Override
     public void invalidateStructure() {
+        getAbilities(ABILITY_ROTOR_HOLDER).get(0).resetRotorSpeed();
```
LAGIdiot, line 89: Please make public function in class which will provide rotor so that it is easily accessible

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
```diff
@@ -169,7 +169,11 @@ public static long getTieredVoltage(long voltage) {
         return GTValues.V[GTUtility.getTierByVoltage(voltage)];
     }
 
-    private void setActive(boolean active) {
+    protected void setActive(boolean active) {
+        // If we are changing states (active -> inactive), clear remaining recipe duration
+        if (this.isActive && !active && recipeDurationLeft != 0) {
+            recipeDurationLeft = 0;
```
LAGIdiot, line 175: No need to check for `recipeDurationLeft` you can just set it

src/main/java/gregtech/api/metatileentity/multiblock/MultiblockControllerBase.java
```diff
@@ -198,7 +198,8 @@ public void invalidateStructure() {
     public void onRemoval() {
         super.onRemoval();
         if (!getWorld().isRemote && structureFormed) {
-            this.multiblockParts.forEach(part -> part.removeFromMultiBlock(this));
+            //this.multiblockParts.forEach(part -> part.removeFromMultiBlock(this));
+            invalidateStructure();
```
LAGIdiot, line 202: I feel that calling `invalidateStructure` is correct move here

LAGIdiot (changes requested):
Thanks for submitting this PR. There are few areas which needs to be changed.

DStrand1:
Alright, requested changes have been implemented.

LAGIdiot (approved):
Thanks for finishing this up.

PR #1424, opened 2021-01-25
Distribute Ticks More Evenly

### LINKED ISSUE #1356: [Suggestion] Distribute expensive ticks more evenly
I've heard people say but haven't profiled that using many covers (or many machines) can be laggy. I read the cover source code and noticed that it will tend to do expensive things on every 20th or 5th tick. If all the covers/machines in a chunk or large area load in after a reload on the same tick then load on the server will be concentrated and every 20th tick will be slower than other ticks.

This doesn't reduce total load but makes it a more smooth lower level of load rather than regular large lag spikes. I noticed at least LogisticsPipes has code to evenly distribute expensive ticks like I'm suggesting.

Change https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/api/metatileentity/TickableTileEntityBase.java to start `timer` from a random number per block, or from a number based on the hash of the tile entity position or ID.

**Additional context**
This may be even a 3 line change. I'd have done it myself already except that I've never set up a mod dev environment before so I'd need to learn how to build and test mods first, and I'm not sure if I'll ever get around to that.

It's possible I missed some distribution in how entities are loaded that effectively does this already, in which case sorry about that.

DStrand1:
**What:**
This PR attempts to fix the issue highlighted in #1356, that ticks are typically distributed in a way that they line up with eachother often, causing especially expensive ticks.

**How solved:**
As described in the comments of the issue, I created a new method `MetaTileEntity#getOffsetTimer()` that returns the timer value plus a new offset, which is a random number between 0 and 20 exclusive. I chose to use Random and not Math.random() here because it is a better random distribution. The old `MetaTileEntity#getTimer()` still exists, but is deprecated.

**Outcome:**
Closes #1356 

**Additional info:**
I do want to point out that the `TickableTileEntity#update()` method still checks for if `timer == 0` to do first tick behavior, and ignores offset completely in timer incrementation. As a result, I don't see that behavior being an issue with this change.

None expected. Addons are expected to change to `getOffsetTimer()` if they wish to see this performance increase, otherwise they can stick with the old `getTimer()` without issues.

### DIFF
--- a/src/main/java/gregtech/api/metatileentity/TickableTileEntityBase.java
+++ b/src/main/java/gregtech/api/metatileentity/TickableTileEntityBase.java
@@ -8,9 +8,9 @@ public abstract class TickableTileEntityBase extends SyncedTileEntityBase implem
 
     private long timer = 0L;
 
-    // Create an offset [0,20] to distribute ticks more evenly
+    // Create an offset [0,20) to distribute ticks more evenly
     private Random random = new Random();
-    private int offset = random.nextInt(21);
+    private int offset = random.nextInt(20);
 
     /**
      * @deprecated This method distributes ticks unevenly.
--- a/src/main/java/gregtech/api/metatileentity/TickableTileEntityBase.java
+++ b/src/main/java/gregtech/api/metatileentity/TickableTileEntityBase.java
@@ -2,12 +2,31 @@ package gregtech.api.metatileentity;
 
 import net.minecraft.util.ITickable;
 
+import java.util.Random;
+
 public abstract class TickableTileEntityBase extends SyncedTileEntityBase implements ITickable {
 
     private long timer = 0L;
 
+    // Create an offset [0,20] to distribute ticks more evenly
+    private Random random = new Random();
+    private int offset = random.nextInt(21);
+
+    /**
+     * @deprecated This method distributes ticks unevenly.
+     * Use {@link TickableTileEntityBase#getOffsetTimer()} instead.
+     */
+    @Deprecated
     public long getTimer() {
         return timer;
+    }
+
+    /**
+     * Replacement for old {@link TickableTileEntityBase#getTimer()}.
+     * @return Timer value with a random offset of [0,20].
+     */
+    public long getOffsetTimer() {
+        return timer + offset;
     }
 
     @Override
--- a/src/main/java/gregtech/api/metatileentity/TickableTileEntityBase.java
+++ b/src/main/java/gregtech/api/metatileentity/TickableTileEntityBase.java
@@ -1,16 +1,14 @@
 package gregtech.api.metatileentity;
 
+import gregtech.api.util.GTUtility;
 import net.minecraft.util.ITickable;
-
-import java.util.Random;
 
 public abstract class TickableTileEntityBase extends SyncedTileEntityBase implements ITickable {
 
     private long timer = 0L;
 
     // Create an offset [0,20) to distribute ticks more evenly
-    private Random random = new Random();
-    private int offset = random.nextInt(20);
+    private final int offset = GTUtility.getRandomIntXSTR(20);
 
     /**
      * @deprecated This method distributes ticks unevenly.
--- a/src/main/java/gregtech/api/capability/impl/RecipeLogicSteam.java
+++ b/src/main/java/gregtech/api/capability/impl/RecipeLogicSteam.java
@@ -141,7 +141,7 @@ public class RecipeLogicSteam extends AbstractRecipeLogic {
     public void update() {
         if (getMetaTileEntity().getWorld().isRemote)
             return;
-        if (this.needsVenting && metaTileEntity.getTimer() % 10 == 0) {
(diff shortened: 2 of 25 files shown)


DStrand1:
I will note that this does work, but it is a little weird how things operate initially. For example, say the initial offset is "3" for a certain tile entity. When it checks the timer, and needs to check, for example, `getTimer() % 5`, the first operation of this is going to occur in 2 ticks, and every subsequent operation occurs in 5. I'm not sure if this addressable, or if it is even that big of a deal. But I wanted to point it out.

DStrand1:
I've tested this (with some help) fairly extensively in the presence of several addons and in a large OF world and found no issues aside from the above.

warjort:
Shouldn't this generate offsets between 0 and 19, i.e. [0,20) not [0,20]?
For 0 to 20, 2 out 21 tiles (the ones with offsets 0 and 20) will effectively be on the same tick.

For the getTimer() % 5 issue, you could add a suggested timer granularity. e.g. 5 would only allow offsets of 0, 5, 10, 15
This would only be used when your change causes a problem.

Something like:
````
// must divide into 20 (i.e. valid values are 1,2,4,5,10)
int getTimerGranularity() {
    return 1; // default is don't care
}
private int offset = getTimerGranularity() * random.nextInt(20 / getTimerGranularity());
}
````

or even simpler is let the tile choose the offset if it needs to:

````
// default implementation
int generateTimerOffset() {
    return random.nextInt(20);
}

@Override // derived class
int generateTimerOffset() {
    // Doesn't work properly with random offsets
    return 0;
}
````

**Comment by Archengius** [comment]
Why did you even touch TickableTileEntityBase#getTimer()? No addons have ever used it directly, my initial concern was about *add-on/GTCE meta tile entities relying on MetaTileEntity#setTimer starting from zero*. You should deprecate getTimer without adding any offsets to it, introduce getOffsetTimer and migrate GTCE machines to using it, carefully investigating every one of them and making sure no startup behaviours are broken.

**Comment by DStrand1** [comment]
I did it like that to try to make an easy fix without having to change every machine, but I understand that your stated solution is more thorough and compatible. I will work on changing it to how you suggested.

**Comment by DStrand1** [comment]
I adjusted the code so that `MetaTileEntity#getTimer()` still behaves exactly as before, but is now deprecated. There is a new method, `MetaTileEntity#getOffsetTimer()`, which implements `ICoverable#getOffsetTimer()`. I left the offset calculation in `TickableTileEntityBase`, as it made the most sense to do this calculation in that class rather than in `MetaTileEntity`. Each instance of `getTimer()` in all of our classes is now replaced with `getOffsetTimer()`, and there is no unexpected behavior after this change.

File: src/main/java/gregtech/api/metatileentity/TickableTileEntityBase.java [hunk 1]
```diff
@@ -2,14 +2,33 @@
...
 
+import java.util.Random;
+
 public abstract class TickableTileEntityBase extends SyncedTileEntityBase implements ITickable {
 
     private long timer = 0L;
 
+    // Create an offset [0,20) to distribute ticks more evenly
+    private Random random = new Random();
+    private int offset = random.nextInt(20);
```
LAGIdiot, line 13: This should be final

src/main/java/gregtech/api/metatileentity/TickableTileEntityBase.java
```diff
@@ -2,14 +2,33 @@
...
 import net.minecraft.util.ITickable;
 
+import java.util.Random;
+
 public abstract class TickableTileEntityBase extends SyncedTileEntityBase implements ITickable {
 
     private long timer = 0L;
 
+    // Create an offset [0,20) to distribute ticks more evenly
+    private Random random = new Random();
```
LAGIdiot, line 12: I don't feel like we need to have Random in every `TickableTileEntityBase` when we use it only once. How about getting random int from some utility class. And maybe even XSTR when we are at it.

LAGIdiot (changes requested):
Thank you for work on this PR. I have found few more places which could use adjustments.

Also did you go through all classes which needs change? For some reason `PipeCoverableImplementation` is missing it and is not buildable.

src/main/java/gregtech/api/metatileentity/TickableTileEntityBase.java
DStrand1: Moved this to `GTUtility` as a private field, and added a public method `getRandomIntXSTR(int bound)`

DStrand1:
For `PipeCoverableImplementation`, I decided to just implement the new `getOffsetTimer()` and have it call `getTimer()` since I feel this is a place that could easily cause problems for the future. Alternatively, I could create a default in `ICoverable` to call `getTimer()` as to not cause other issues for addons if you would like.

LAGIdiot (approved):
Thank you for finishing this up. I will give it quick spin to ensure that it works correctly.

LAGIdiot:
I have tested this on my testing world and everything seemed to run correctly and placing new machines did not produce and problems either. When I take in account testing you did it should work correctly.

PR #1428, opened 2021-01-26
Allow Fluid to be Plasma and Placeable

### LINKED ISSUE #1358: [BUG] Adding a material and adding both GENERATE_FLUID_BLOCK and GENERATE_PLASMA flags crashes the game

DStrand1:
**What:**
This PR fixes the crash demonstrated in #1358 and revitalizes the PR #1359. You are now able to generate a FluidMaterial with both flags `GENERATE_FLUID_BLOCK` and `GENERATE_PLASMA`.

**How solved:**
I simply added a check to `registerFluid()` that checks for if the current fluid type is Plasma. This does not block the fluid from being placeable, but instead just prevents it from being registered twice, which was causing the crash.

**Outcome:**
Closes #1358 

**Additional info:**
Demonstration of a test fluid being placed:
placeFluid

Demonstration showing that a Plasma form of the same material exists, and cannot be placed:
cantPlacePlasma

This solution fixes the issue very simply, but it does close the door on having a Plasma able to be placed in the world. I don't expect that to be a problem, though, based on discussion in the issue and the old PR.

None expected.

### DIFF
--- a/src/main/java/gregtech/common/MetaFluids.java
+++ b/src/main/java/gregtech/common/MetaFluids.java
@@ -186,7 +186,7 @@ public class MetaFluids {
 
         FluidRegistry.addBucketForFluid(fluid);
 
-        if (material.hasFlag(MatFlags.GENERATE_FLUID_BLOCK) && fluid.getBlock() == null) {
+        if (material.hasFlag(MatFlags.GENERATE_FLUID_BLOCK) && fluid.getBlock() == null && fluidType != FluidType.PLASMA) {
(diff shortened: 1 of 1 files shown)


LAGIdiot (approved):
Thanks for your hard work

PR #1434, opened 2021-01-31
Formatting proposal

LAGIdiot:
**What:**
This is proposal to unite our formatting styles. Provided file is extract of my current IntelliJ IDEA project profile which was put together from code style used in this project. Styles for all files used in this project are available.

This will automatically apply to our IDEs (as long that option is not turn off). Which will save us some time on PRs.

This topic is open for discussion but given that it is hot theme if it become hate fest I will lock it. Also this is not about creating new formatting style but about providing guide lines for current one. So keep it in mind.


warjort:
I have seen other mods using spotless:

https://github.com/diffplug/spotless/tree/main/plugin-gradle

This is integrated into gradle so will apply regardless of ide (or even not using an ide).
It does more than simple formatting if you want.

But see the requirements section for the versions of gradle supported.

LAGIdiot:
I don't feel like gradle plugin is right way. Because you don't have to use gradle to work on project. I basically work without it whole time and use it only for generating Changelog or when I need to reset workspace.

PR #1435, opened 2021-01-31
Fix NPE When Placing Covers

### LINKED ISSUE #1343: [BUG] NullPointerException when placing shutter and facades
**Describe the bug**
NullPointerException when placing down shutter and facades.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.7.594

**Setup**
Playing Solo
New world generated

**Steps To Reproduce**
1) place down 3x3 wall of wires or machines.
2) have a stack of shutters/facades in hands
3)hold right mouse button while circling the wall
4)notice log full of null point exceptions

**Expected behavior**
Clean log

method `rayTraceCoverableSide` in ICoverable.java may return null, which leads to `coverable.getCoverAtSide(coverSide)` to get a cover for a 'null' side, leading to the NPE.
`rayTraceCoverableSide` calls for `retraceBlock` which is part of chickenlib.

I sugest a `coverSide == null` check before calling `coverable.getCoverAtSide(coverSide)`

DStrand1:
**What:**
This PR addresses the issue shown in #1343, where when placing a cover on a cable, your log would be filled with NPE errors.

**How solved:**
I added more robust null checking in a few places. There was a section where previously, we had a `Preconditions.checkNotNull()` call on `side`, but this would get hit and still cause a log error, so I instead changed this to return false if null.

**Outcome:**
Closes #1343 

**Additional info:**
Previously, I was able to get an error in the log following the steps to reproduce very consistently and almost immediately. With these changes, I tested it on 10 sets of cables and was unable to get an error.

None expected.

### DIFF
--- a/src/main/java/gregtech/api/cover/ICoverable.java
+++ b/src/main/java/gregtech/api/cover/ICoverable.java
@@ -114,6 +114,9 @@ public interface ICoverable {
     }
 
     static boolean doesCoverCollide(EnumFacing side, List<IndexedCuboid6> collisionBox, double plateThickness) {
+        if (side == null) {
+            return false;
+        }
         if (plateThickness > 0.0) {
             Cuboid6 coverPlateBox = getCoverPlateBox(side, plateThickness);
             for (Cuboid6 collisionCuboid : collisionBox) {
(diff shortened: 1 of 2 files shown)


LAGIdiot (approved):
Good work fixing this issue.

Issue #1211, opened 2020-08-23
[Suggestion] Easier Multiblock build preparation/execution


LAGIdiot:
**Is your feature request related to a problem? Please describe.**
1. When building multiblocks too much time is taken by finding what blocks are needed. Mostly when searching structure where are all inputs and outputs.
2. Controller for multiblocks (`MultiblockWithDisplayBase`) is showing only "Invalid structure" when there is something wrong with multiblock.

**Describe the solution you'd like**
1. Extend JEI page for multiblocks to show list of used blocks. With highlighting minimal number of them.
2. Show more specifically what is wrong with structure. Like "Input Bus is missing", "Wrong block on position [17,67,-300] should be Air" or "Structure not finished".

**Help requested:**
There is need to iterate on this as I am not sure how to do it as best as possible right now. I would be glad for any input on this.

huneau:
We talk with @Shattered2909 last time for a tool to auto-build multiblock. 
I take the question on another side ... if its to complex to display issues info of what is missing. maybe make a tool that make a copy of the JEI pattern and reproduce it by placing blocks.
Last time I try to debug a multiblock design, and I was a bit frustrating of what was wrong and I code small thing into the predicate that highlight the wrong/missing block like now MB Tweaker. I think its should be doable with a minimum effort.

As a side suggestion related, @Shattered2909 (always the same guy that harras me :P) ask me to make multiblock configurable, in some sort, he would like to say what casing and what render should multiblock be. Maybe for multiblock we/you can implement the same thing we do (its @hjae78 who dev the thing) this will avoid me to break your code by replacing your metatileentity id and also break compatibility with MBtweaker.

ALongStringOfNumbers:
There could also be some mention of the functionality added in #1090 as well as the previously existing shift + left click to pitch the  structure.

ALongStringOfNumbers:
With the zoom feature added in #1090, it would be nice if we could start the JEI preview of the mutliblocks at a custom level of zoom. This would be nice for multiblocks defined via multiblocktweaker or addon mods, which could be somewhat large and not all display in the JEI preview at the default level of zoom.

PR #1452, opened 2021-02-08
Update JEI Multiblocks to show block counts and allow bookmarking

### LINKED ISSUE #1393: [Suggestion] The ability to save items to the JEI favorites list from the multiblock preview
When looking at the preview of multiblocks, you have to remember the component parts of the multiblock when exiting the screen. This can be somewhat solved when the list of component items introduced in #1211 is implemented, but before then there are not many good solutions

I would like the ability to add items from the multiblock preview to the JEI favorites list, preferably with the JEI favorite button, `a`. This would be implemented best by sending the item that you are hovering your cursor over to the favorite list when pressing `a`.

**Additional context**
An additional improvement that could be done is pulling up the recipes for items when `r` is pressed while hovering over the item in the preview. And then going back to the preview screen after pressing the back button from the recipe.

warjort:
**What:**
This shows the number of blocks required to build the currently shown pattern in the JEI multiblock view and allows bookmarking of those blocks.

**How solved:**
Added 20 slots at the bottom of the screen. 
These are filled from the block drops of the multiblock in the dummy world, which was sort of already being collected (just not the amounts).
The new slots are part of an IGuiItemStackGroup to enable all the extra JEI features like bookmarking and r/u behaviour. 
The left/right buttons to switch between multiblock variants have been moved to the right hand side below the layer switch button.

**Outcome:**
Closes #1393 
and does part 1 of #1211

I made it 20 slots because I know gregicality has some complicated multiblocks.

### DIFF
--- a/src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoCategory.java
+++ b/src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoCategory.java
@@ -4,6 +4,7 @@ import gregtech.api.GTValues;
 import gregtech.common.metatileentities.MetaTileEntities;
 import gregtech.integration.jei.multiblock.infos.*;
 import mezz.jei.api.IJeiHelpers;
+import mezz.jei.api.IGuiHelper;
 import mezz.jei.api.IModRegistry;
 import mezz.jei.api.gui.IDrawable;
 import mezz.jei.api.gui.IRecipeLayout;
@@ -20,11 +21,13 @@ public class MultiblockInfoCategory implements IRecipeCategory<MultiblockInfoRec
 
     private final IDrawable background;
     private final IDrawable icon;
+    private final IGuiHelper guiHelper;
 
     public MultiblockInfoCategory(IJeiHelpers helpers) {
-        this.background = helpers.getGuiHelper().createBlankDrawable(176, 166);
+        this.guiHelper = helpers.getGuiHelper();
+        this.background = this.guiHelper.createBlankDrawable(176, 166);
         ResourceLocation iconLocation = new ResourceLocation(GTValues.MODID, "textures/gui/icon/coke_oven.png");
-        this.icon = helpers.getGuiHelper().createDrawable(iconLocation, 0, 0, 16, 16, 16, 16);
+        this.icon = this.guiHelper.createDrawable(iconLocation, 0, 0, 16, 16, 16, 16);
     }
 
     public static final Map<String, MultiblockInfoRecipeWrapper> multiblockRecipes = new HashMap<String, MultiblockInfoRecipeWrapper>() {{
@@ -78,7 +81,7 @@ public class MultiblockInfoCategory implements IRecipeCategory<MultiblockInfoRec
 
     @Override
     public void setRecipe(IRecipeLayout recipeLayout, MultiblockInfoRecipeWrapper recipeWrapper, IIngredients ingredients) {
-        recipeWrapper.setRecipeLayout((RecipeLayout) recipeLayout);
+        recipeWrapper.setRecipeLayout((RecipeLayout) recipeLayout, this.guiHelper);
     }
 
 }
--- a/src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoRecipeWrapper.java
+++ b/src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoRecipeWrapper.java
@@ -4,12 +4,15 @@ import codechicken.lib.raytracer.CuboidRayTraceResult;
 import codechicken.lib.raytracer.IndexedCuboid6;
 import codechicken.lib.vec.Cuboid6;
 import codechicken.lib.vec.Vector3;
+import gregtech.api.gui.GuiTextures;
 import gregtech.api.render.scene.SceneRenderCallback;
 import gregtech.api.render.scene.WorldSceneRenderer;
 import gregtech.api.util.BlockInfo;
 import gregtech.api.util.GTUtility;
 import gregtech.api.util.ItemStackKey;
+import mezz.jei.api.IGuiHelper;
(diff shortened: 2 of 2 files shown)


LAGIdiot:
Thanks for you hard work. Please update your opening comment with screenshots as you are changing GUI.

Regarding #1393 it provides requested functionality, probably not in way that @ALongStringOfNumbers expected but it should be usable right?

For #1211 It would be better to specify what part were implemented so we can update that issue after this one is closed.

I have quite few points regarding appearance of this update. 

I think we should render parts little bit closer to bottom and align it to center of window.
Scene should be rendered from top of GUI and possibly leave some space between.
Buttons with arrows should not touch parts area.


Parts should have same order for all variants and multiblock types. Possibly starting with Controller, followed by active blocks then passive blocks.


Sometimes hovering over some buttons highlight parts slots, and sometimes they grey out.


Hovering over bottom 1/3 of part does not show tooltip.

warjort:
For reference this is what I see with the auto gui scale:


Just so you understand why I crammed it in. There isn't really any space at the top or bottom of the GUI at this scale.

For centering, the left hand side of these slots is already at x=1, I think the only way to make it look centred is to reduce the number of slots to 18 and move the whole thing to the right.

I haven't done anything with tool tips. I will investigate what magic incantation I did to do to make JEI? do this properly. :-)
The only thing I am drawing is the background texture of the slot.
For me the tooltip only seems to work in the top half of the slot.

On the ordering, the only thing I have access to is the BlockInfo which is a position, IBlockState and a TileEntity.
In the original code, it had already gone through a HashMap by the time it calculates the block drops so the order is "random".

Given the limited info available, it could be ordered something like:
* TileEntity is a controller block
* other tile entities
* other blocks
and I could maybe include the number of blocks and the block id to try to keep ordering the same across variants?
Let me know if this is ok, or you have a different idea.

DStrand1:
I think it would be worthwhile to mention some details regarding some of the ingredients too. For example, the EBF requires a minimum of 10 Heat-Proof casings to form, and fluid hatches are optional for it to form.

warjort:
> I think it would be worthwhile to mention some details regarding some of the multiblock parts too. For example, the EBF requires a minimum of 10 Heat-Proof casings to form, and fluid hatches are optional for it to form. These could either be appended to the multiblock part tooltip, or replace it in some instances

That information is not available at this part of the code, if it is available at all anywhere.
All that is available are the blocks in the scene/variant you are viewing in the GUI.
The easiest way to provide this extra information atm would be to add something to the default tooltip of the controller blocks.

LAGIdiot:
> I think it would be worthwhile to mention some details regarding some of the multiblock parts too. For example, the EBF requires a minimum of 10 Heat-Proof casings to form, and fluid hatches are optional for it to form. These could either be appended to the multiblock part tooltip, or replace it in some instances

This all is mentioned in #1211 and I think we can come up with much more what could be useful.

ALongStringOfNumbers:
I am not sure how hard it would be, but to attempt to save space on the gui, you could reduce the number of rows to 1 and then use arrows to show more items. 

An example would be: The multiblock has 18 items, the first 9 get rendered in the slots and you have a `>` button (Similar to how the EBF has one for displaying different coils) that will render the next 9 items in the slots. And then you can go back to the first nine items with a `<` button.

Of course, these arrows could be somewhat big and might require reducing the number of slots show

warjort:
> I will investigate what magic incantation I need to do to make JEI? do this properly. :-)

I think I understand why this is not working. The bottom half of the top row of slots is outside the background we have to told JEI to use.

To make this work for the double row of slots, the background would need to be extended by 30 "pixels". Does that sound ok?

warjort:
For the space saving, could I move the left and right arrows for changing scenes to be together under the LA button for changing layers? That would free up space at the bottom for the parts slots.

LAGIdiot:
> For the space saving, could I move the left and right arrows for changing scenes to be together under the LA button for changing layers? That would free up space at the bottom for the parts slots.

Feel free

warjort:
> Sometimes hovering over some buttons highlight parts slots, and sometimes they grey out.

This looks like a bug in minecraft? 

The highlighting of the buttons is sometimes leaking into the draw operations for the parts background texture.
I don't exactly understand the circumstances to always reproduce it and it is not always the last button drawn, but it looks like I have fixed this by changing the code so the buttons are drawn last.

Very weird.

warjort:
I think I have done all the things you asked for?

The left/right buttons have been moved to make space for the slots so they now fit on the JEI background properly which fixes the tooltip issue. I didn't change the size of the background so it will now show properly in the auto gui scale.
See the new image in the top post.

I moved the slots to start at x=0 and they look less off centre now the positions of the buttons are not pointing it out.

I have sorted the parts as I suggested above. This was more complicated than expected as the data only becomes available in phases and so has to be collected in multiple steps. Not the most elegant code I ever wrote. :-)

I worked around the button highlight issue by changing the drawing order.

LAGIdiot:
I see what you did regarding my request and it seems like you did everything that is possible without rewrite. And I understand that other things are not possible without it.

One other thing that needs to be done is move render scene above parts slots. So rendered multiblock can be hidden behind it.

warjort:
I changed the height of the rendered scene so it doesn't try to draw under the part slots.

However, the line I changed looked like a typo. I guess it didn't matter since the area is pretty much square?
````
-int sceneHeight = recipeWidth;
+int sceneHeight = recipeHeight-36;
````

LAGIdiot:
> I changed the height of the rendered scene so it doesn't try to draw under the part slots.
> 
> However, the line I changed looked like a typo. I guess it didn't matter since the area is pretty much square?
> 
> ```
> -int sceneHeight = recipeWidth;
> +int sceneHeight = recipeHeight-36;
> ```

I guess it was a typo

LAGIdiot (approved):
Thanks for finishing this up.

PR #1461, opened 2021-02-13
Plunger as a voiding fluid itemstack

### LINKED ISSUE #1389: [Suggestion] Make Plunger a fluid container
Would be great if Plunger was usable as an always empty infinite sized fluid container. That way it would be more usable in GTCE machine UIs for example through clicking with it on exact tanks to clear them, which is a nicer solution than just blindly clicking with it on the machine block in world and hoping that it would void fluid from the right tank.

Implement fluid handler item capability on it.

warjort:
**What:**
Allow the plunger to be used as a fluid container that voids fluids in fluid slots.

**How solved:**
Implemented a fluid itemstack handler that voids any fluid that goes into it.
Gave the plunger this capability augmented with damage applied to the plunger, currently 1 durability per 1000mb (the same as the in world usage).

**Outcome:**
Closes #1389


warjort:
As per #1389 the plunger is infinite capacity. However this introduces an issue related to damage.
i.e what to do when there is not enough durability left on the plunger to soak up all the fluid?

As per the implementation for the in world usage, I decided to simply check there is enough durability to soak up the fluid and do nothing if there isn't.
But for the in world usage it only does 1000mb at a time with a cost of 1 durablity per 1000mb. 

For "infinite amounts", there is a legitimate expection that the plunger should void fluids until it breaks instead of doing nothing.
This would require doing a calculation based on IToolItem information, but there is a comment in that code saying not to do just that. i.e. you must call damageItem() and let it do the calculation.

I could do a loop and remove durability 1 damage point at a time and return the filled fluid amount accordingly, but I suspect this may cause issues interacting with "creative" fluid slots.

Archengius:
> I could do a loop and remove durability 1 damage point at a time and return the filled fluid amount accordingly, but I suspect this may cause issues interacting with "creative" fluid slots.

You cannot, since fluid containers can have their fill/drain simulated.

> For "infinite amounts", there is a legitimate expection that the plunger should void fluids until it breaks instead of doing nothing.
This would require doing a calculation based on IToolItem information, but there is a comment in that code saying not to do just that. i.e. you must call damageItem() and let it do the calculation.

I think it is a reasonable exception. Keep in mind though that situation where you have too little durability of the plunger and it refuses to drain fluid at all should be avoided. Just break if it there's not enough durability and transfer 1 bucket.

warjort:
>  Just break if it there's not enough durability and transfer 1 bucket.

If there is 64 buckets in a tank and the plunger only has 50 durability it will currently do nothing.
You're suggestion would break the plunger but leave 63 buckets in the tank.

That seems like a bad deal? :-)

The issue is in I don't know how much durability got used when it breaks unless I use the IToolItem data.

Archengius:
You got my message completely wrong. What I suggested is to retrieve amount of durability remaining from tool item, then drain corresponding amount of buckets. But if we don't have enough durability even for 1 bucket, drain it anyway and break the plunger.

warjort:
I have implemented the partial draining of the fluid slot if the plunger breaks.

Let me know if my use of IToolItem breaks the rules (it probably does if there was an electric plunger :-)

LAGIdiot (changes requested):
I think that warning is there for a reason and should be respect it as we don't have means to prevent anyone from creating electric plunger.
For calculating purposes I would suggest to do dry run of `IToolItem.damageItem` with required durability to clean all. If it fails then you can by bisecting durability to find max which can tool handle. Sure it is not as clean but I don't have anything better.

warjort:
If I understand correctly, you want me to do a binary chop to find the min damage that breaks it?

That won't work. All I have in simulate mode from IToolItem.damageItem() is "I did damage" yes or no.
It doesn't say if it actually broke it.

The complication comes from electric items that will return "no I didn't damage it" when they are out of charge, which is why you are not suppossed to use the durability directly.

The normal usage is for the caller to damage the item first then do something if it did damage. That doesn't work in this case where you are doing something potentially infinite things and want to know when to stop,

warjort:
> That doesn't work in this case where you are doing potentially infinite operations and want to know when to stop,

At least it doesn't work in simulate mode when nothing is recording the cumulative damage.

And I need simulate mode because the fluid extract can be simulated as @Archengius pointed out.

warjort:
The following is the outline of "nearly" backward compatible patch that changes the damageItem api so I can do it properly.
It schematically does
````
- boolean damageItem(ItemStack, damage, simulate);- 
+ int damageItem(ItemStack, damage, allowPartial, simulate);
````
So I can know in simulate mode what damage is applied and therefore how much liquid to drain.

It still doesn't work properly for electric items since they do an "all or nothing" with the damage like my original implementation of this feature. To make them work, we would need to do something with the allowPartial flag, see below:

````
diff --git a/src/main/java/gregtech/api/items/IToolItem.java b/src/main/java/gregtech/api/items/IToolItem.java
old mode 100644
new mode 100755
index fb27b874..9ec72d34
--- a/src/main/java/gregtech/api/items/IToolItem.java
+++ b/src/main/java/gregtech/api/items/IToolItem.java
@@ -1,5 +1,6 @@
 package gregtech.api.items;
 
+import gregtech.api.util.GTLog;
 import net.minecraft.item.ItemStack;
 
 /**
@@ -13,8 +14,29 @@ public interface IToolItem {
      * <p>
      * DO NOT USE METHODS BELOW TO CHECK IF TOOL CAN RECEIVE SPECIFIED AMOUNT OF DAMAGE,
      * Use this method with simulate = true to check so, because it can be different for electric items!
+     *
+     * PLEASE IMPLEMENT damageItemNew()
      */
-    boolean damageItem(ItemStack stack, int damage, boolean simulate);
+    default boolean damageItem(ItemStack stack, int damage, boolean simulate) {
+        return damageItemNew(stack, damage, false, simulate) > 0;
+    }
+
+    /**
+     * Tries to apply given amount of damage to item
+     * If simulated, actual damage won't be applied and durability won't be changed
+     * <p>
+     * DO NOT USE METHODS BELOW TO CHECK IF TOOL CAN RECEIVE SPECIFIED AMOUNT OF DAMAGE,
+     * Use this method with simulate = true to check so, because it can be different for electric items!
+     *
+     * @return the actual damage applied
+     */
+    default int damageItemNew(ItemStack stack, int damage, boolean allowPartial, boolean simulate) {
+        GTLog.logger.warn("Item damage is not accurate", new Exception(this.toString()));
+        if (damageItem(stack, damage, simulate))
+            return 1;
+        else
+            return 0;
+    }
 
     /**
      * @return amount of internal damage this item have
diff --git a/src/main/java/gregtech/api/items/toolitem/ToolMetaItem.java b/src/main/java/gregtech/api/items/toolitem/ToolMetaItem.java
old mode 100644
new mode 100755
index eb9a32c5..672370dc
--- a/src/main/java/gregtech/api/items/toolitem/ToolMetaItem.java
+++ b/src/main/java/gregtech/api/items/toolitem/ToolMetaItem.java
@@ -328,30 +328,37 @@ public class ToolMetaItem<T extends ToolMetaItem<?>.MetaToolValueItem> extends M
     }
 
     @Override
-    public boolean damageItem(ItemStack stack, int vanillaDamage, boolean simulate) {
+    public int damageItemNew(ItemStack stack, int vanillaDamage, boolean allowPartial, boolean simulate) {
         IElectricItem capability = stack.getCapability(GregtechCapabilities.CAPABILITY_ELECTRIC_ITEM, null);
         if (capability != null) {
             int energyAmount = ConfigHolder.energyUsageMultiplier * vanillaDamage;
             if (capability.discharge(energyAmount, capability.getTier(), true, false, true) < energyAmount) {
+
+//
+// Here we do something with allowPartial
+//
                 //if we can't discharge full amount of energy, just return false
                 //and don't attempt to discharge left amount of energy
-                return false;
+                return 0;
             }
             capability.discharge(energyAmount, capability.getTier(), true, false, simulate);
         }
         T toolMetaItem = getItem(stack);
         if (toolMetaItem == null) {
-            return false;
+            return 0;
         }
         IToolStats toolStats = toolMetaItem.getToolStats();
         if (!toolStats.isUsingDurability(stack)) {
-            return true;
+            return vanillaDamage;
         }
-        int newDamageValue = getItemDamage(stack) + calculateToolDamage(stack, itemRand, vanillaDamage);
+        int itemDamage = getItemDamage(stack);
+        int maxDamage = getMaxItemDamage(stack);
+        int damageRemaining = maxDamage - itemDamage;
+        int newDamageValue = itemDamage + calculateToolDamage(stack, itemRand, vanillaDamage);
         if (!simulate && !setInternalDamage(stack, newDamageValue)) {
             GTUtility.setItem(stack, toolStats.getBrokenStack(stack));
         }
-        return true;
+        return Math.min(vanillaDamage, damageRemaining);
     }
 
     public int regainItemDurability(ItemStack itemStack, int maxDurabilityRegain) {
`````

warjort:
I say above, it is "nearly" backward compatible above, because there is one other place in GTCE, Shadows of Greg and Gregicality that implements damageItem().

AbstractToolItemCapabilityProvider does it as a kind of proxy.

Invocations of damageItemNew would call the default interface method, even though the proxied object has an implementation.

Of course, I can fix it in this case and nothing would be actually using it that way (yet).
But if somebody is doing some clever dynamic proxying it is hard to make it fully backward compatible.

LAGIdiot:
I have to try this up on my environment and give it some real though as it is getting bigger. I will try to get to it on Saturday morning as first thing.

LAGIdiot:
I have given this some thoughts and I come up with similar idea. I think we should mark `boolean damageItem(ItemStack, damage, simulate)` and all it's overrides as deprecated. And add new `int damageItem(ItemStack, damage, allowPartial, simulate)` as you are suggesting to which we will switch all our calls. This should allow us to do new stuff which we need and warn our addons that we are planning to remove old ones (which we will probably never done).

Regarding `AbstractToolItemCapabilityProvider` I think we can do same and still be in clear.

Does it make sense to you or did I miss something critical which prevents us from doing so?

warjort:
First, I was wrong about the AbstractToolItemCapability.

While it does proxy requests to IToolItem.damageItem(), they are not from that method. 
Instead they are similarly named methods on the screwdriver, wrench, etc. capabilities.

So this requires no change.

warjort:
I have done the changes, but not quite as described above or you requested.
I made the "damageItemNew" method abstract and also removed the "new" from its name.

I also haven't deprecated the old method or changed the exisitng code to use the new method.
In my opinion, the existing method is better for the common case
````
if (tool.damageItem(itemStack, damage, simulate))
   doStuff();
````
it is better than
````
if (tool.damageItem(itemStack, damage, false, simulate) > 0)
   doStuff();
````

The new method is for if you need to know the damage inflicted or want to do a drain of "insufficient charge" on an electric item.

warjort:
Finally, I haven't implemented the "allowPartial" for electric items.

While I don't think it would be that difficult, something like:
````
diff --git a/src/main/java/gregtech/api/items/toolitem/ToolMetaItem.java b/src/main/java/gregtech/api/items/toolitem/ToolMetaItem.java
index a006c414..db887cf9 100755
--- a/src/main/java/gregtech/api/items/toolitem/ToolMetaItem.java
+++ b/src/main/java/gregtech/api/items/toolitem/ToolMetaItem.java
@@ -337,6 +337,9 @@ public class ToolMetaItem<T extends ToolMetaItem<?>.MetaToolValueItem> extends M
             // TODO when allowPartial is true, replace energyAmount with remaining charge
             // and recalculate the damage done/returned based on that
 
+                energyAmount = capability.discharge(Integer.MAX_VALUE, capability.getTier(), true, false);
+                vanillaDamage = roundUp(energyAmount/ConfigHolder.energyUsageMultiplier);
+
                 //if we can't discharge full amount of energy, just return false
                 //and don't attempt to discharge left amount of energy
                 return 0;
````

There is nothing like an electric plunger with which to test it. :-)

LAGIdiot:
I was thinking in much larger scope which is not really needed for this. But I am 100% OK with doing just necessary changes required for Plunger needs and we can expand on it later if we need.

Regarding discharge I think you need to calculate with `vanillaDamage` on input. Honestly I am not sure if we should introduce electric variant too (of course not electronic plunger). As I don't expect that there will be that much usage for it.

warjort:
I have added an implementation for allowPartial.

By the way, the invocation of discharge() just after this code looks redundant when simulate=true?

warjort:
While testing that I hadn't broken the existing electric tools I found a typo in BlockMachine where it was applying screwdriver damage to the wrench.

I have fixed it as part of this PR.

LAGIdiot:
Good work updating this PR. Code regarding partial discharge looks solid.

Regarding second discard yes it could be skipped in simulate mode. Thanks for spotting and fixing that problem with wrong damage being a applied.

Regarding todo for playing sound to player, we don't have sound for it...

warjort:
> Regarding todo for playing sound to player, we don't have sound for it...

Couldn't it just use the sponge sound?

LAGIdiot:
You can try and if it makes sense then why not.

warjort:
I still don't have a player or world/blockpos to control who to send to.

LAGIdiot:
It seems that sound is already played here: https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/common/items/behaviors/PlungerBehaviour.java#L52

warjort:
That is for the in world plunger usage where you right click the block.
I see it is using the slime sound instead of the sponge.

This new behaviour is when you drop it in a fluid slot. 
Ideally we would want to know the blockpos of whatever is invoking fill() and send the sound to nearby players, but I don't think that information is reachable.

LAGIdiot:
Yeah it does not seem that we are able to play sound on this end. I would rather not block this PR because of sound. And if needed we can return for sound later.

LAGIdiot (approved):
Good job on this PR. Let's take it as it is. And we can iterate on it later if needed.

PR #1463, opened 2021-02-13
Don't generate orebyproduct recipes when there are no byproducts fixe…

### LINKED ISSUE #1448: [BUG] Ore Byproduct page fails in some circumstances
**Describe the bug**
In some circumstances the ore byproduct page fails in that it does not display the output material in the output slots, but instead in the final slot in the second machine processing row.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.11.1.632
_Modpack:_ None
_Addons:_ None

**Setup**
Playing Solo
_New world generated_ Yes

**Steps To Reproduce**
1) Open JEI
2) Look at the byproducts page of Oilsands

**Expected behavior**
The final output to be in the output slots of the ore byproduct page

**Screenshots**
Oilsands

Bismuth

warjort:
**What:**
Ore Byproducts recipes showing with no byproducts in JEI.

**How solved:**
Does not add recipes to JEI for materials that have no by products.
A page will still exist if the material is a byproduct of something else.

**Outcome:**
Fixes #1448


LAGIdiot (approved):
Thanks for your hard work.

PR #1464, opened 2021-02-13
Don't place the same tile multiple times for JEI multiblock rendering

### LINKED ISSUE #1379: [BUG] Cracker Unit is not shown formed in JEI
**Describe the bug**
The Cracker Unit is not shown as formed in JEI, such that the inputs and the outputs take the casing texture.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.8.599
_Modpack:_ None
_Addons:_ None

**Setup**
Playing Solo
_New world generated_ Yes

**Steps To Reproduce**
1) Open Jei and look at the Cracker Unit preview in JEI

**Expected behavior**
That the Cracker Unit is shown formed in JEI

**Screenshots**

warjort:
**What:**
Multiblocks not forming in the JEI rendering scene

**How solved:**
Made sure individual tile entites are placed in the dummy world rather placing the same tile in multiple locations.
This allows the recipemap constraints to be satisified and the multiblock will now form.

**Outcome:**
Fixes #1379

### DIFF
--- a/src/main/java/gregtech/integration/jei/multiblock/MultiblockShapeInfo.java
+++ b/src/main/java/gregtech/integration/jei/multiblock/MultiblockShapeInfo.java
@@ -5,6 +5,7 @@ import gregtech.api.metatileentity.MetaTileEntityHolder;
 import gregtech.api.util.BlockInfo;
 import gregtech.common.blocks.MetaBlocks;
 import net.minecraft.block.state.IBlockState;
+import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.EnumFacing;
 
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


LAGIdiot (commented):
This does not seems right. I would understand if you use some factory to generated new instances. Is this best possible solution?

warjort:
MetaTileHolder.setMetaTileEntity() is the factory AFAIK.

This code is pretty much the same as that done in the where() method, execpt I have to steal the "frontFacing" from the tile created there, otherwise the multiblock wouldn't form properly.

warjort:
The part that feels wrong to me is that it is being done twice. Once in the where() method and now in BlockInfo array bake.

But I don't think I can fix that without breaking the api used for the multiblock jei structures which is used by addons including MultiblockTweaker kind of exposing it via crafttweaker.

LAGIdiot (commented):
I am looking at this now and problem if am correct this problem is only on Cracker Unit which is only multiblock which has 2 of same active blocks used. So it may be related that you have same object on two positions. So maybe for active blocks we should use new instances every time we are adding them in `bakeArray`? Question is what all needs to be new.

warjort:
This is my comment from the original bug report:
https://github.com/GregTechCE/GregTech/issues/1379#issuecomment-774716368

This bug fix makes new instances for all the tile entities since those are the ones that add MultiblockParts/Abilities used to determine if the RecipeMap constraints are satisfied and the multiblock can form.

Processing for them is only done for tiles, e.g. in MultiControllerBase
````
            ArrayList<IMultiblockPart> parts = new ArrayList<>(rawPartsSet);
            parts.sort(Comparator.comparing(it -> ((MetaTileEntity) it).getPos().hashCode()));
````

Normal blocks just add a block state. Typically just the block id, but it could be a facing or any other property.
Sharing this BlockState in the dummy world makes no difference.

If in the future a normal block could add an ability there are many places in the code that would need to be changed.

LAGIdiot:
Sorry I missed your comment on that issue. Your analysis is up to point. And your fix is sufficient if we don't want to rewrite it which is not desired.

LAGIdiot (approved):

PR #1465, opened 2021-02-13
Mechanism to copy text to the player's clipboard

### LINKED ISSUE #1391: [Suggestion] /gt util hand should copy the meta id to clipboard
After running the command `/gt util hand`, the metaitem name is output into the chat, after which you still have to type out this metaitem name into a script.

It would be nice if the usage of the `/gt util hand` command could copy the item's metaname into your clipboard, similar to the `/ct hand` command from Crafttweaker. This would save having to type out the item's metaitem name into the script, and instead simply paste the name into the script

warjort:
**What:**
Add a generic mechanism to copy text to the player's clipboard and use it in the "gt util hand" command for the metavalueitem's name

**How solved:**
Created a new gregtech.api.net.PacketClipboard which is just a string.

Added 2 new util methods
ClipboardUtil.copyToClipboard(String) - copies the string to the clipboard if the AWT desktop is available
ClipboardUtil.copyToClipboard(EntityPlayerMP, String) - sends the network packet to the client which will invoke the other method

When "/gt util hand" sends the chat message to display the MetaValueItem.unlocalizedName it now also sends that value to the player's clipboard

**Outcome:**
Closes #1391

### DIFF
--- a/src/main/java/gregtech/api/util/GTUtility.java
+++ b/src/main/java/gregtech/api/util/GTUtility.java
@@ -11,8 +11,6 @@ import gregtech.api.gui.impl.ModularUIContainer;
 import gregtech.api.items.IToolItem;
 import gregtech.api.metatileentity.MetaTileEntity;
 import gregtech.api.metatileentity.MetaTileEntityHolder;
-import gregtech.api.net.NetworkHandler;
-import gregtech.api.net.PacketClipboard;
 import gregtech.common.ConfigHolder;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockRedstoneWire;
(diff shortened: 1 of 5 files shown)


LAGIdiot (changes requested):
Good work on this PR.

Please move all newly added content from `GTUtility` to `ClipboardUtil` so it is in own file.

warjort:
Done

src/main/java/gregtech/api/util/GTUtility.java
```diff
@@ -50,6 +50,7 @@
 import net.minecraftforge.items.IItemHandlerModifiable;
 
 import javax.annotation.Nullable;
+
```
LAGIdiot, line 53: Please don't do unnecessary changes to files which are not directly related to given changes.

I know you were touching this file in previous commits but right now there is no to touch this file now.

LAGIdiot (changes requested):
Thanks for updating this PR. I have find one small thing which needs to be fixed.

LAGIdiot (approved):
Thank you for finishing this up.

LAGIdiot:
@warjort this should work every time or is there some specific case when this does not work?

warjort:
It only copies to the clipboard when MetaValueItem.unlocalizedName is sent to chat.
Not all blocks do that.

LAGIdiot:
Oh I see, thanks for info.

PR #1466, opened 2021-02-14
Don't set held item for non player listeners

### LINKED ISSUE #1020: workbench crash with FTBquesting
i have made a quest with FTB questing mod for a bronze ingot. it is already completed but whever i open the workbench GU, navigate to the chest/internal storage and left click bronze, i crash.

it appears when you left click an item in this mode, the item selected is automatically out inside without requiring you to manually move it, however this is what appears to cause FTB questing to crash

i have contacted Latvian modder on discord and his response is attached on the screenshot from "lat"

image

versions:
Forge:  14.23.5.2847
GTCE:  1.9.0.481
FTBQ:  1.8.0.237

crashlog:
https://pastebin.com/5tS3inyA

warjort:
**What:**
Crash when FTB Quests is installed and any ModularUIContainer is used, e.g. the crafting station

**How solved:**
Don't try to set the player's held item when the container listener is not a player.

**Outcome:**
Fixes #1020

### DIFF
--- a/src/main/java/gregtech/api/gui/impl/ModularUIContainer.java
+++ b/src/main/java/gregtech/api/gui/impl/ModularUIContainer.java
@@ -121,8 +121,10 @@ public class ModularUIContainer extends Container implements WidgetUIAccess {
     @Override
     public void sendHeldItemUpdate() {
         for (IContainerListener listener : listeners) {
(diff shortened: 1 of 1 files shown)


LAGIdiot (approved):
Thanks for fixing this

Issue #1323, opened 2020-11-23
[Suggestion] Clearer display for large numbers


ALongStringOfNumbers:
**Is your feature request related to a problem? Please describe.**
As a revival of https://github.com/GregTechCE/GregTech/pull/872 I would like to see a system making large sets of numbers more readable, as currently it can be hard to distinguish between large sets of numbers for higher power tiers. This would be especially prevalent for addon mods that add higher tier content, as their numbers could run on and become very hard to read. 

**Describe the solution you'd like**
For smaller numbers, such as things approximately of the magnitude of 1 million EU, the numbers could be delimited using commas (or another method if it is deemed better)

For larger numbers, collapsing the number to use suffixes such as K, M, G (Kilo, Mega, Giga) would make the number much more readable. However, in this case, when shift is held to show more tooltip, maybe the number could expand back to being comma delimited. 

The magnitudes I mention here are just suggestions, and it may be determined instead to use one format on one set of items, and the other format on another set of items. Such as using the suffixes on items in game, but being comma delimited in JEI.

**Additional context**
This of course should be open for addon mods to use, as they would have one of the main benefits from this as they add theirs beyond what currently exists in Gregtech CE, which means that they would be dealing with larger number sets.

Exaxxion:
Java 12+ has a built-in capability that essentially does this, respecting Locale (which is very useful for i18n), in the form of [CompactNumberFormat](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/text/CompactNumberFormat.html).

Obviously this isn't available in Java 8, so you'd have to implement something similar. For reference, the GPLv2 implementation of this class from OpenJDK 14 is available [here](http://hg.openjdk.java.net/jdk/jdk/file/ee1d592a9f53/src/java.base/share/classes/java/text/CompactNumberFormat.java).

million09:
Project EX has helpful formatter.
https://github.com/LatvianModder/Project-EX/blob/master/src/main/java/com/latmod/mods/projectex/gui/EMCFormat.java

Blood-Asp:
The old #872 merge request did work and was tested. I also did test multiple different formats (like Kilo, Mega, Giga) in the beginning and choose what worked the best
No idea why it was closed. Just reopening and merging should take care of this.

LAGIdiot:
[...]

PR #1467, opened 2021-02-14
Improved number formatting, with a couple of other fixes

warjort:
**What:**
Improved number formatting to include thousands seperators according to locale.

**How solved:**
Changed %d in the language file to %,d

**Outcome:**
Does some of #1323

**Additional info:**
While doing this I discovered a few places where %d was used when it should have been using %s or %f

I also found that the code for slot widget in the crafting table used gregtech.item_list.item_stored while the language file was gregtech.item_list.items_stored, I changed the language file.

There are some places like the multiblock boiler or the gt util hand command that use minecraft's TextComponentTransalation api. This doesn't work with %,d so I didn't change these. Other places that use this api in the code, format numbers using %s.

I haven't touched the russian or chinese translations. They will not have the thousands seperators until somebody that speaks those languages does the same excercise.

If one of the addon mods is reusing some of these translations, it may break their display if they have the same misuse as the %s and %f issue I fixed.


warjort:
One other change I did inadvertantly but decided to keep for the purposes of comment is that the GT tools attack damage and digging speed now show the first decimal. e.g. before it might show Attack Damage: 7, now it will show Attack Damage: 7.2

I don't know if the decimals are actually used by minecraft?

LAGIdiot:
I have tagged this as open for discussion. As I would like to hear opinion of others. Please don't turn this to hate fest...

@warjort also don't leave comments of what was changed to what in code we can see that from git history

warjort:
> @warjort also don't leave comments of what was changed to what in code we can see that from git history
But translators don't necessarily know how to read git diffs. The ==> is for them.

warjort:
Take a look at #1484 if you want to see what the %,d looks like.

LAGIdiot:
We are not using any translator services. So knowledge how to use git is required. Which means they should be able to use diff too.

LAGIdiot:
I am pinging this one from Discord so we can get discussion going

Issue #1022, opened 2020-02-04
[Suggestion] Add open chest/machine option for spectator


HonzaVinCZ:
I find this useful as server administrator.

PR #1473, opened 2021-02-15
Not count spectators in MetaTileEntityChest

CrafterKolyan:
**What:**
Do not count spectators in MetaTileEntityChest each 10 seconds.

**Outcome:**
Effectively nothing changed due to Spectators can not use Chests (see #1022)

Not possible

### DIFF
--- a/src/main/java/gregtech/api/util/GTUtility.java
+++ b/src/main/java/gregtech/api/util/GTUtility.java
@@ -78,7 +78,7 @@ public class GTUtility {
     }
 
     public static Stream<Object> flatten(Object[] array) {
-        return Arrays.stream(array).flatMap(o -> o instanceof Object[] ? flatten((Object[]) o) : Stream.of(o));
+        return Arrays.stream(array).flatMap(o -> o instanceof Object[] ? flatten((Object[]) o): Stream.of(o));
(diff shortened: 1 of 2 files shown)


warjort:
Why wouldn't you do the spectator counting in GTUtility instead of iterating again in the chest?
This reduces the work and makes the functionality available to others.

Something like:

````
// New method
static List<EntityPlayerMP> findPlayersUsing(MetaTileEntity metaTileEntity, doube radius, boolean includeSpectators) {
    // The old method with logic for spectators
}

// Old method for backwards compatibilty
static List<EntityPlayerMP> findPlayersUsing(MetaTileEntity metaTileEntity, doube radius,) {
     return findPlayersUsing(metaTileEntity, radius, false);
}
````

CrafterKolyan:
@warjort Done it, thanks for an idea.

CrafterKolyan:
This one is not needed as `net.minecraft.world.World#getEntitiesWithinAABB(java.lang.Class<? extends T>, net.minecraft.util.math.AxisAlignedBB)` actually calls `this.getEntitiesWithinAABB(p1, p2, EntitySelectors.NOT_SPECTATING)`

PR #1475, opened 2021-02-15
Fix chest placement render problem (Fix #1474)

### LINKED ISSUE #1474: [BUG] When placing chest it renders for 2 ticks as a solid block instead of transparent
**Describe the bug**
Placing a chest makes some strange artifacts appear

**Versions**
Forge: 14.23.5.2847
GTCE: 1.11.3

**Setup**
Playing Solo

**Steps To Reproduce**
1) Place some light source
2) Place chest near
3) See how it blinks for 0.1 second 

**Expected behavior**
No blinking, just as normal chest

**Screenshots**
Just placed a chest:
image
After 0.1 s:
image

CrafterKolyan:
**What:**
Fix chest placement render problem

**How solved:**
Change default light opacity to 0 which means that it's the same as air (and basically we can say that nothing is air)

**Outcome:**
Fix #1474 

Chest's light opacity now is 0 instead of 1 (the same as for Vanilla Chest)
If we on current position there is no `MetaTileEntityHolder` then we assume that it is air, so return opacity of 0 instead of 255.

Basically this problem occurs with any transparent block in Gregtech like Tank etc.

### DIFF
--- a/src/main/java/gregtech/api/block/machines/BlockMachine.java
+++ b/src/main/java/gregtech/api/block/machines/BlockMachine.java
@@ -18,7 +18,6 @@ import gregtech.api.cover.IFacadeCover;
 import gregtech.api.metatileentity.MetaTileEntity;
 import gregtech.api.metatileentity.MetaTileEntityHolder;
 import gregtech.api.render.MetaTileEntityRenderer;
-import gregtech.api.util.GTLog;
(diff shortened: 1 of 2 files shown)


LAGIdiot (approved):
I see we will start with expecting Machine being transparent and then update it based on tile entity which will be way less intrusive then current approach.

Thanks for your hard work

PR #1478, opened 2021-02-15
Fix some Russian translation

CrafterKolyan:
**What:**
Fix some Russian translation

**How solved:**
I am native Russian speaker

Not possible


LAGIdiot (approved):
Thanks for your hard work. I will accept this in good faith that it is correct as I am not able to confirm it.

PR #1479, opened 2021-02-15
Fix the caching for container items in the CraftingStation

### LINKED ISSUE #1451: [BUG] Crafting with empty tank in crafting station
**Describe the bug**
Can craft with empty GTCE tanks using GTCE crafting station

**Versions**
Forge: 14.23.5.2855
GTCE: 1.10.9.612
_Modpack:_ GSE 5.0.14QF2 https://www.curseforge.com/minecraft/modpacks/gregicality-skyblock-edition

**Steps To Reproduce**
1) Fill GTCE tank with any fluid required for recipe
2) Craft with this tank in GTCE crafting station

**Expected behavior**
Can't craft without fluid in tank.

**Video**

https://user-images.githubusercontent.com/15050771/107158916-40005180-698d-11eb-8e4e-38fb84eacf35.mp4

warjort:
**What:**
The crafting station lets you craft with empty fluid containers when a fluid is specified in the recipe.

The issue is caused because a call to IRecipe.getRemainingItems() can change the cached and validated recipe held in CachedRecipeData. e..g For a fluid container with fluid inside, it can be updated to be the empty fluid container.

Because the data is assumed valid in the cache, it would from then on be using the changed/broken recipe as though it were valid. All it wants is the empty fluid container to process the recipe.

This change also gets copied into the crafting grid displayed to the user when it tries to show what recipe was actually used.

**How solved:**
When calling getRemainingItems(), a deep copy of the recipe is passed. i.e. the item stacks are copied
This avoids any random recipe implementation changing the cache/validated recipe.

**Outcome:**
Fixes #1451

**Additional info:**
The 2nd part of this issue would also lead to other issues like a broken GT tool getting replaced with "air" in the crafting grid and stopping further crafting even if there is a valid tool in the crafting station's inventory.

### DIFF
--- a/src/main/java/gregtech/api/util/GTUtility.java
+++ b/src/main/java/gregtech/api/util/GTUtility.java
@@ -22,7 +22,6 @@ import net.minecraft.client.renderer.GlStateManager;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.EntityPlayerMP;
 import net.minecraft.init.Items;
-import net.minecraft.inventory.InventoryCrafting;
 import net.minecraft.inventory.Slot;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
@@ -86,14 +85,6 @@ public class GTUtility {
             ItemStack itemStack = src.getStackInSlot(i);
             dest.setStackInSlot(i, itemStack.isEmpty() ? ItemStack.EMPTY : itemStack.copy());
         }
-    }
-
-    public static InventoryCrafting deepCopyInventoryCrafting(final InventoryCrafting source) {
-        final InventoryCrafting result = new InventoryCrafting(new DummyContainer(), source.getWidth(), source.getHeight());
-        for (int i = 0; i < result.getSizeInventory(); ++i) {
-            result.setInventorySlotContents(i, source.getStackInSlot(i).copy());
-        }
-        return result;
     }
 
     public static <T> IntStream indices(T[] array) {
--- a/src/main/java/gregtech/api/util/GTUtility.java
+++ b/src/main/java/gregtech/api/util/GTUtility.java
@@ -22,6 +22,7 @@ import net.minecraft.client.renderer.GlStateManager;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.EntityPlayerMP;
 import net.minecraft.init.Items;
+import net.minecraft.inventory.InventoryCrafting;
 import net.minecraft.inventory.Slot;
 import net.minecraft.item.Item;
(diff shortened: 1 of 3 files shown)


src/main/java/gregtech/api/util/GTUtility.java
```diff
@@ -87,6 +88,14 @@ public static void copyInventoryItems(IItemHandler src, IItemHandlerModifiable d
...
     }
 
+    public static InventoryCrafting deepCopyInventoryCrafting(final InventoryCrafting source) {
+        final InventoryCrafting result = new InventoryCrafting(new DummyContainer(), source.getWidth(), source.getHeight());
+        for (int i = 0; i < result.getSizeInventory(); ++i) {
+            result.setInventorySlotContents(i, source.getStackInSlot(i).copy());
+        }
+        return result;
+    }
+
```
LAGIdiot, line 98: Please move this to new file `InventoryCraftingUtil`

src/main/java/gregtech/api/util/GTUtility.java
```diff
@@ -87,6 +88,14 @@ public static void copyInventoryItems(IItemHandler src, IItemHandlerModifiable d
         }
     }
 
+    public static InventoryCrafting deepCopyInventoryCrafting(final InventoryCrafting source) {
+        final InventoryCrafting result = new InventoryCrafting(new DummyContainer(), source.getWidth(), source.getHeight());
+        for (int i = 0; i < result.getSizeInventory(); ++i) {
```
LAGIdiot, line 93: Please change implementation to use `i++` instead of `++i` as it is more common through project.

LAGIdiot (changes requested):
Thanks for creating this PR. There two items I would like to get changed.

src/main/java/gregtech/api/util/GTUtility.java
warjort: If you have a real issue with this code I will fix it. But I am not going to do this.

LAGIdiot: I am trying to keep this file from growing to unreasonable size and having everything mixed up in it (I know that it contains quite a bit of everything now but it should not continue). To make code more readable by separating it to smaller chunks with up to point naming.

I feel quite bad that I have to request these changes as it should be something obvious. But if I don't do it quality of project would go down and it will become even harder to get bearing and start developing on top of it. So please just do it.

warjort:
I have done the change, but I put it in InventoryUtils instead of creating a whole new class.

I would suggest you add a comment to the top of GTUtility stating your policy. It will save everybody a lot of time.

LAGIdiot (approved):
Thank you for updating this as requested.

PR #1481, opened 2021-02-15
Use material color in particle texture, optimize color convertion

### LINKED ISSUE #1480: [BUG] Particle texture for chest does not represent a material
**Describe the bug**
If you break chest the particles you see are always white.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.11.3

**Setup**
Playing Both

**Steps To Reproduce**
1) Place Bronze Chest
2) Start trying to break it
3) See white particles

**Expected behavior**
Particles of chest should be equal to bronze

**Screenshots**
Just broke chest
image
Broke chest and instantly placed it (to see difference)
image

CrafterKolyan:
**What:**
Now material color is used in particle texture the same way as for rendering.

**How solved:**
Multiply `material.materialRGB` to previous result with converting RGB to RGBA and vice versa.

**Outcome:**
Fix #1480 

**Screenshots:**
image

Not possible, only new methods appeared. Old methods return the same result.

Take into account that we don't need to do and operation `& 0xFFFFFF` when doing right shift for 8 bits due to `int` overflow.
Also consider that in `convertOpaqueRGBA_CLtoRGB` we need a zero right shift instead of simple right shift in order not to get 1s in first 8 bits in case first bit of OpaqueRGBA_CL would be 1 (which is possible when red channel >= 128).

### DIFF
--- a/src/main/java/gregtech/api/util/GTUtility.java
+++ b/src/main/java/gregtech/api/util/GTUtility.java
@@ -54,7 +54,6 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
-import java.util.List;
 import java.util.*;
 import java.util.Map.Entry;
 import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


CrafterKolyan:
Here is how it works with blue paint on different tanks
image

LAGIdiot (approved):
Thanks for your hard work on this PR.

PR #1482, opened 2021-02-15
Optimize Render Bounding Box for Chest and Safe

CrafterKolyan:
**What:**
The chest doesn't open downwards so it's not needed to subtract 1 by Y axis in render bounding box
The safe doesn't open nor downwards nor upwards so it's not needed to subtract 1 and add 1 by Y axis in render bounding box

**Outcome:**
Nothing changes except for some small optimization in render mechanism when some chests that are higher than your Field of View won't be rendered. Same for safes (but also some of which were lower than your Field of View).

**Additional info**
For testing I've commented line 98 here, so that the chest was always opened after first opening:
https://github.com/GregTechCE/GregTech/blob/f80e5dcf581f6490126988620ca952cebfd1f754/src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityChest.java#L95-L99
After that I checked all possible angles and most importantly looking on chest from below. At first I incorrectly changed it to be `(0, 0, 0), (1, 2, 1)`, but then I understood that chest's top part is actually on the neighbouring block when it is opened, so we need to have bounding box of size 3x2x3 which is done in this pull request.

For safe it's the same but its size should be 3x1x3

For `MetaTileEntityHolder` there is just a refactoring - no logic changed.

Not possible

### DIFF
--- a/src/main/java/gregtech/api/metatileentity/MetaTileEntityHolder.java
+++ b/src/main/java/gregtech/api/metatileentity/MetaTileEntityHolder.java
@@ -214,7 +214,7 @@ public class MetaTileEntityHolder extends TickableTileEntityBase implements IUIH
     @Override
     public void onLoad() {
         super.onLoad();
-        if(metaTileEntity != null) {
+        if (metaTileEntity != null) {
             metaTileEntity.onLoad();
         }
     }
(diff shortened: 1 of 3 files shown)


LAGIdiot (approved):
Thanks for spotting this problem and fixing it.

PR #1483, opened 2021-02-16
Add ability to set Input Allowed via the machine grid

### LINKED ISSUE #1350: [Suggestion] The ability to set input allowed on the output side via the machine grid
I would like the ability to set "Allow Input from Output Side", which is done using a screwdriver, using the machine grid. Currently, you have to use the screwdriver on the actual output side of the machine to enable this option, which gets cumbersome when placing a lot of machines. This is because you have to either temporarily move the output, hit the output with a screwdriver, and the move the output back, or walk to the other side of the machine.

Right Clicking the screwdriver on the machine grid on the side that the output is on should toggle the "Allow Input from Output Side" option.

ALongStringOfNumbers:
**What:**
This PR adds the ability for Input Allowed from Output Side to be set via the machine grid by holding shift when right clicking on the machine grid with a screwdriver. 

The behavior of screwdrivers has also been slightly changed so that using while holding shift when accessing the output side of a machine with a cover on the output side will set Input Allowed via Output Side instead of accessing the cover. This was done to ensure uniformity of the behavior of shift + right clicking with a screwdriver.

**How solved:**
Adds some checks to determine if the player is holding shift when accessing the output side, and if the player is attempting to access the output side of a machine that has a cover on the output side

**Outcome:**
Allows setting Input Allowed via Output Side from the machine grid by holding shift when right clicking with a screwdriver. Closes #1350

I do not expect any compatibility issues.


LAGIdiot (commented):
I don't like introduced coupling between `MetaTileEntity` and `SimpleMachineMetaTileEntity`. I think this could be done bit clearer by creating new interface which will contain OutputFacing and AllowInputFromOutputSide. Then this interface can be used in `MetaTileEntity` instead of whole class.

What you think?

ALongStringOfNumbers:
The reason I introduced the specific check of if the `MetaTileEntity` was a `SimpleMachineMetaTileEntity` was that they are the ones that have an output side, while other `MetaTileEntities` don't have an output side. In addition I needed to ensure that there would be consistent behavior while using a screwdriver on a cover that is on the output side, because the cover would get checked first, opening the UI instead of setting output side.

As for your idea of an Interface that will contain OutputFacing and AllowInputFromOutputSide, that could work, although it would need to be checked in `MetaTileEntity` that the properties exist, which should not be that big of an issue.

LAGIdiot:
I mean code will be nearly same but instead of `SimpleMachineMetaTileEntity` which is concrete implementation you will check interface which is abstract.

ALongStringOfNumbers:
I have resolved the class coupling as per the suggestion of Arch on discord by creating a new Capability that is implemented by `SimpleMachineMetaTileEntity`. This capability contains basic information about the output side and status of the machine, and is used in place of the `instanceof` check.

LAGIdiot (commented):
Good work this looks solid from Capability implementation standpoint.

I see you used open variant when whole block has capability and you can get all info from it. I was thinking about close variant when only output side would provide this capability. I am not sure which variant is better.

I will open this for discussion what is better. Because this could be used on other places.

LAGIdiot:
I was testing this when I found out I would like to use more close variant when capability is only provided on output side.

So I reworked it a bit. I hope you don't mind.

Also I cleaned up few places to allow `SimpleMachineMetaTileEntity` to make it more usable for extension.

LAGIdiot (approved):
It is kind of strange to approve this when I partially wrote it. But let's be consistent.

PR #1484, opened 2021-02-16
Add an IFuelable capability to display fuel info in TOP

### LINKED ISSUE #1486: Opening up of FuelRecipeLogic to subclasses
In #1422 the method(s) to find fuel recipes was opened up in FuelRecipeLogic to allow subclasses to take part.

Quoting @DStrand1
>since FuelRecipeLogic was not able to properly check for the secondary condition of having a rotor present, so I had to override it

In fact, the change in the subclass is nothing to do with finding recipes and instead tests whether you are ready to process recipes at all.

Something like this change would revert this implementation back to private and introduce a proper api for what is trying to be achieved. It adds an isReadyForRecipes() method where subclasses can stop the processing.

````
diff --git a/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java b/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
index 3760ed23..758f729b 100755
--- a/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
@@ -102,7 +102,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
                     }
                 }
             }
-            if (recipeDurationLeft == 0) {
+            if (recipeDurationLeft == 0 && isReadyForRecipes()) {
                 tryAcquireNewRecipe();
             }
         }
@@ -112,11 +112,15 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
         }
     }
 
+    protected boolean isReadyForRecipes() {
+       return true;
+    }
+
     protected boolean shouldVoidExcessiveEnergy() {
         return false;
     }
 
-    protected void tryAcquireNewRecipe() {
+    private void tryAcquireNewRecipe() {
         IMultipleTankHandler fluidTanks = this.fluidTank.get();
         for (IFluidTank fluidTank : fluidTanks) {
             FluidStack tankContents = fluidTank.getFluid();
@@ -134,7 +138,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
         return isActive;
     }
 
-    protected int tryAcquireNewRecipe(FluidStack fluidStack) {
+    private int tryAcquireNewRecipe(FluidStack fluidStack) {
         FuelRecipe currentRecipe;
         if (previousRecipe != null && previousRecipe.matches(getMaxVoltage(), fluidStack)) {
             //if previous recipe still matches inputs, try to use it
diff --git a/src/main/java/gregtech/common/metatileentities/multi/electric/generator/LargeTurbineWorkableHandler.java b/src/main/java/gregtech/common/metatileentities/multi/electric/generator/LargeTurbineWorkableHandler.java
old mode 100644
new mode 100755
index c7b09685..f4c659ac
--- a/src/main/java/gregtech/common/metatileentities/multi/electric/generator/LargeTurbineWorkableHandler.java
+++ b/src/main/java/gregtech/common/metatileentities/multi/electric/generator/LargeTurbineWorkableHandler.java
@@ -77,19 +77,9 @@ public class LargeTurbineWorkableHandler extends FuelRecipeLogic {
     }
 
     @Override
-    protected void tryAcquireNewRecipe() {
-        IMultipleTankHandler fluidTanks = this.fluidTank.get();
+    protected boolean isReadyForRecipes() {
         MetaTileEntityRotorHolder rotorHolder = largeTurbine.getAbilities(MetaTileEntityLargeTurbine.ABILITY_ROTOR_HOLDER).get(0);
-        for (IFluidTank fluidTank : fluidTanks) {
-            FluidStack tankContents = fluidTank.getFluid();
-            if (tankContents != null && tankContents.amount > 0 && rotorHolder.isHasRotor()) {
-                int fuelAmountUsed = tryAcquireNewRecipe(tankContents);
-                if (fuelAmountUsed > 0) {
-                    fluidTank.drain(fuelAmountUsed, true);
-                    break; //recipe is found and ready to use
-                }
-            }
-        }
+        return rotorHolder.isHasRotor();
     }
 
     @Override
````

My interest is that as part of #1484 I need to "simulate" the recipe search and  I need access to that
isReadyForRecipes()/isHasRotor() logic too in a clean way.

### PR DESCRIPTION
**Pull request by warjort**
**What:**
Shows fuel information in TOP

**How solved:**
Introduced a new capability IFuelable and a TOP provider to display the information.

**Outcome:**
Does most of #1397 - does not do unrelated things like the suggested change to the boiler's throttle.
Closes #1486 

**Additional info:**
This is the opposite of what LAGIdiot approved, but if you still want this info in the boiler's display panel it should be trivial to add.
Instead it supplies a consistent view of fuel inventories across all blocks, including single blocks.

### CODE REVIEW

**Comment by warjort** [comment]
First some "eye candy" :-)

Large Boiler


Coal boiler


Lava boiler


Steam Turbine


Diesel generator


Multiblock Diesel Generator

**Comment by warjort** [comment]
Now for the real discussion.

The main changes are in FuelRecipeLogic which means this behaviour will "leak" into most fuel based generators.
I had to do seperate implementations for the steam boilers because they don't use FuelRecipeLogic.

Issues:

1) Cosmetics: I am displaying the fuel's name, the amount of fuel remaining, the fuel capacity of the tanks,slots etc. and the estimated fuel burn time of the remaining fuel.
I am pretty sure for some Gregicality fuels this use of the name is going end up overflowing the space in the TOP panel.
I thought about drawing the item, but this isn't going to be helpful for fluids since you can't mouse over the TOP info.

2) Cosmetics: The colours are the same as that used for the EnergyContainer. Probably they should be different?

3) Reload glitch: I have implemented this in FuelRecipeLogic by using the lastFuelRecipe as the fuel to check and display.
However this is null after a reload of the game. So unless the generator starts consuming fuel there is no information to display.
For now it will show "No fuel/idle" but that is intended to fixed. I will make a new comment on a proposed solution.

4) Large Boiler algorithm: The implementation in the LargeBoiler is very long. It is basically just a copy and modify of the existing method used to find the recipe(s). It is complicated by having 3 different recipe sources and using both fluid and item fuels.

5) Large Boiler fuels: In the example above it shows a burn time for "Raw Oil". In practice this is untrue because there is a minimum amount of fuel required to activate the recipe and the example shown above is below that. AFAIK there is nowhere that informs the user of this "feature".

**Comment by warjort** [comment]
For point (3) above. A possible solution is when the "lastFuelRecipe" is null and the generator is idle, to run the current recipe search in "simulate" mode so it can find the correct information to display.

But the current recipe search is using protected methods meaning addons can change the behaviour and existing implementations won't know about me adding a simulate mode.

A different solution would be to just search the tanks and pick a fluid at random as the one to display. But without the recipe, there is no way to show an estimated burn time.

File: src/main/resources/assets/gregtech/lang/en_us.lang [hunk 1]
```diff
@@ -17,6 +17,8 @@ gregtech.top.transform_up=§cStep Up§r %s (%dA) -> %s (%dA)
 gregtech.top.transform_down=§aStep Down§r %s (%dA) -> %s (%dA)
 gregtech.top.transform_input=§6Input:§r %s (%dA)
 gregtech.top.transform_output=§9Output:§r %s (%dA)
+gregtech.top.fuel_info=/%,d %,d s
```
CrafterKolyan, line 20: Wouldn't it be better to add a little bit bigger space?
```suggestion
gregtech.top.fuel_info=/%,d\t%,d s
```
Also I think it would print these two in a column but not sure. Can you provide screenshots for first two with `\t` (tab character). I really don't like how space looks in first two screenshots, I think it can be easily fixed.

warjort:
> Wouldn't it be better to add a little bit bigger space?

I am already worried that names like "Dense Hydrazine Fuel Mixture" from gregicality are not going to fit on this panel.
And it is longer in Russian "Плотная гидразиновая топливная смесь" :-)

warjort:
One other thing I forgot to mention is that I made this a general capability rather than a tile capability, e.g. an extension of IWorkable.

This is because I think this same interface can be used to expose information for things like a fuel powered jetpack.

src/main/resources/assets/gregtech/lang/en_us.lang
```diff
@@ -17,7 +17,8 @@ gregtech.top.transform_up=§cStep Up§r %s (%dA) -> %s (%dA)
 gregtech.top.transform_down=§aStep Down§r %s (%dA) -> %s (%dA)
 gregtech.top.transform_input=§6Input:§r %s (%dA)
 gregtech.top.transform_output=§9Output:§r %s (%dA)
-gregtech.top.fuel_info=/%,d %,d s
+gregtech.top.fuel_info=/%,d for %,d:%02d mins
```
CrafterKolyan, line 20: I think it's pretty understandable that it's in measured in minutes
```suggestion
gregtech.top.fuel_info=/%,d for %,d:%02d
```

warjort:
Having slept on this, I decided to change it so the fuels are displayed on 2 lines.
This should remove any issues with information going outside the TOP panel.

I have also changed the time remaining to be in minutes:seconds format

This is what it now looks like:

src/main/java/gregtech/integration/theoneprobe/provider/FuelableInfoProvider.java
```diff
@@ -0,0 +1,61 @@
...
+            probeInfo.text(TextStyleClass.WARNING + I18n.format("gregtech.top.fuel_none"));
+            return;
+        }
+        for (IFuelInfo fuelInfo : fuels) {
+            final String fuelName = fuelInfo.getFuelName();
+            final int fuelRemaining = fuelInfo.getFuelRemaining();
+            final int fuelCapacity = fuelInfo.getFuelCapacity();
+            final int burnTime = fuelInfo.getFuelBurnTime()/20;
+            final int burnTimeMins = burnTime/60;
+            final int burnTimeSecs = burnTime%60;
```
CrafterKolyan, line 46:
```suggestion
            final int burnTime = fuelInfo.getFuelBurnTime() / 20;
            final int burnTimeMins = burnTime / 60;
            final int burnTimeSecs = burnTime % 60;
```
How about this refactoring?

src/main/resources/assets/gregtech/lang/en_us.lang
warjort: 12:45 Is it mins and secs? It could be hours and mins. A newbie isn't going to know.

src/main/java/gregtech/integration/theoneprobe/provider/FuelableInfoProvider.java
warjort: nfc - no functional change

src/main/resources/assets/gregtech/lang/en_us.lang
CrafterKolyan: How about `0:13 mins`? I think it's 13 minutes.

CrafterKolyan: A newbie will see that it updates each second and will easily understand the measurement of time

src/main/java/gregtech/integration/theoneprobe/provider/FuelableInfoProvider.java
CrafterKolyan: Of course, it's no functional change, because it's refactoring. It's needed just for better readability.

src/main/resources/assets/gregtech/lang/en_us.lang
CrafterKolyan: Basically `mins` doesn't solve the problem you are talking about because it's still may be **mins** and secs or hours and **mins**.

warjort: can you put the comments in the pr rather than a converstion where it is easier for others to see and discuss.

personally I find cosmetics one of the least interesting parts of software development and don't want to waste any time discussing it

warjort:
To resolve point (5) above, I haved added a minimum required fuel to the capability.
This will now be displayed instead of burn time when there is not enough fuel.


This should help with the large boiler where this number is not the same as that displayed in JEI.

warjort:
I have reworked FuelRecipeLogic to use a combination of the solutions I proposed to solve issue (3)

warjort:
I have changed the display of burn time back to seconds.

After some debugging and reading of code for String.format() and TOP, I eventually found that if you try to use a format like:
%02d as I was doing to create a 2 digit zero padded seconds display,
Minecraft's locale class will replace it with %s when it loads it from the language file.

One numeric qualifier infront of d or f is allowed.

I guess as some point a Minecraft dev didn't want to do the work of actually fixing all their language files manually. :-)

LAGIdiot:
I am really impressed. This looks like missing piece we (or at least I) did not know that it was missing. I think this will provide quite valuable information. Provided pictures looks really solid.

From conversation I can see that you solve most of problems. And I think that using whole line for fluid name is good move. Regarding point 3 solutions seems good. I quickly looked through code used in addons and did not found any references to now private methods but I will still mark this as minor release just to be sure. For point 2 it would be cool to have it look bit differently then other but I am not good at styling so I will it on you or others if they participate. For point 4 that one is what it is, boilers are bit difficult...

DStrand1:
I noticed there was a change to the LargeTurbineWorkableHandler due to making `tryAcquireNewRecipe` private again. This is a good change in my opinion, but did you thoroughly test the Large Turbine behavior after this change? I would hate to see a fix for it from just a version or two get undone.

LAGIdiot:
> I noticed there was a change to the LargeTurbineWorkableHandler due to making `tryAcquireNewRecipe` private again. This is a good change in my opinion, but did you thoroughly test the Large Turbine behavior after this change? I would hate to see a fix for it from just a version or two get undone.

I have look at it too and it should not be undone as there is `isReadyForRecipes()` which will prevent it searching new recipe in `update`.

DStrand1:
> I have look at it too and it should not be undone as there is `isReadyForRecipes()` which will prevent it searching new recipe in `update`.

Yeah, I see that now, but in my experience, the Large Turbine is exceptionally hard to predict its behavior, so I think just giving it a quick test would be a really good idea.

LAGIdiot (approved):
From code perspective this looks solid. I will approve it now but feel free to to still look at point 2.

There small discrepancies in code style but I will ignore them as proposed solution regarding that did not yet come to validity. And I would rather not have another heated discussion regarding that

Issue #1460, opened 2021-02-13
[BUG] Shift-click to move a stack of items in/out of a GTCE machine/chest is rarely failing


anpilley:
**Describe the bug**
Since I started testing a dev version of Omnifactory (snapshot e3e6711), I've been having an issue when interacting with GTCE machines that wasn't showing up on earlier versions of the pack (that have earlier builds of GTCE)

Shift-clicking a stack to move it into, or out of, a GTCE chest or machine/bus is rarely failing to complete successfully. This happens most often when shift-clicking multiple successive stacks (like 8 or so stacks of some kind of dust, for instance). The UI updates, initially, showing all the stack have been moved, but then a short period later, things sync back up, and one of the multiple stacks will not have been moved, and is in its original location.

This is happening in both directions, either into a GT inventory block, or out of a GT inventory block. 

This was not happening in earlier versions of the pack, and since it's fairly early game, there's no significant TPS lag (TPS is clamped at 20 with no noticeable lag spikes). This isn't happening with other mods containers (like AA crates or vanilla minecraft chests) in the same pack that I've noticed.

This is unfortunately relatively rare, it's only happening roughly every 30 minutes to an hour or so of normal play, so it's difficult to catch and reproduce, but it's happening often enough that it doesn't feel like I'm just not clicking or missing the stack with my mouse (and no other issues with the mouse are noticeable in other games, currently)

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.9.612
_Modpack:_ Omnifactory dev snapshot e3e6711 (https://github.com/OmnifactoryDevs/Omnifactory/releases/tag/latest-dev-preview). Unfortunately, that specific dev snapshot has been replaced with a newer one at this point, but I only picked up the dev snapshot a few days ago, so there shouldn't be significant changes.
_Addons:_ No extra addons.

**Setup**
Playing Solo, in void dimension (via void cake), standard default lost cities world gen for starting overworld.

**Steps To Reproduce**
1) Have chest with multiple stacks of items (3-8 stacks of 64 or so)
2) use shift-click to rapidly move all the stacks into/out of inventory into/out of GT bus/chest/machine. 
3) Observe that all of the stacks have been moved to player's inventory.
4) close block's inventory UI.
5) move to another container
6) open GT container, begin moving items in using shift-click
7) Notice that one of the stacks failed to make it across from the original starting location.

**Expected behavior**
[...]

Issue #1190, opened 2020-08-10
[Suggestion] Add support for Mouse Tweaks mod


LAGIdiot:
**Is your feature request related to a problem? Please describe.**
Adding Mouse Tweaks (https://www.curseforge.com/minecraft/mc-mods/mouse-tweaks) mod support would make GTCE Chests and Crafting Station more easily usable or at least on pair with others. This could bring more people to choose it over other crafting tables.

**Describe the solution you'd like**
Add support for Mouse Tweaks mod as optional dependency and  implement it's interface on SlotWidgets.

LAGIdiot:
As #1485 is fixing wheel actions for which was this issue created it can be closed.

PR #1485, opened 2021-02-17
Fix wheel action in MouseTweaks and Crafting Station voiding items

### LINKED ISSUE #1291: [BUG] Crafting Station can void items from crafts
**Describe the bug**
The crafting station can void items from crafts that have multiple numbers of the output item depending on which mouse button is used to take the item. It seems like the functionality to right click and take half of a stack is being combined with a click to take outputs from the output slot.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.4.574
_Modpack:_ tested in minimal environment
_Addons:_ None

**Setup**
Playing Solo (only tested solo)
_New world generated_ if applicable: yes

**Steps To Reproduce**
1) Place a crafting station down and enter the vanilla recipe: 1 log -> 4 wood planks
2) Left click on the output slot and receive 4 wood planks, as expected
3) Right click on the output instead and receive 2 wood planks, which is half of the expected output. 

**Expected behavior**
Receive all items from crafts that output multiple of the same item.

CrafterKolyan:
**What:**
Now you can use a wheel action added by MouseTweaks, also fix #1291 

**How solved (full explanation of bug):**
Do not wrap player inventory when passing it to SlotWidget. Consider how it is implemented in `forgeSrc`
image
I've debugged MouseTweaks and found out that the main problem is that it just doesn't understand which items correspond to the player's inventory and which to some other stuff. We always falling into `continue;` on these lines:

https://github.com/YaLTeR/MouseTweaks/blob/minecraft-1.12.2/src/main/java/yalter/mousetweaks/Main.java#L474-L480

When we create a `SlotItemHandler` it's just connecting all slots to a constant `emptyInventory`:
image

To be precise, `findInPlayerInventory` is always `false` as `selectedSlot.inventory == SlotItemHandler.emptyInventory`. The same applies for `slot.inventory` as well as for any `Slot` which was created with usage of `SlotWidget`.

All we need to do is just fix this stuff by providing correct `IInventory`.

**UPD:** Now also fixes #1291 by changing `IItemHandler` to `IInventory` in `CraftingSlotWidget`.

**Outcome:**

https://user-images.githubusercontent.com/9883873/108194144-4f1e9680-7127-11eb-83b9-90b2e35293a7.mp4

**Possible compatibility issues**:
~Now `SlotWidget.getHandle()` returns `SlotDelegate` instead of `SlotItemHandler` **which may lead to logical errors when checking result of `getHandle()` with `instanceof`** (see https://github.com/GregTechCE/GregTech/pull/1485#issuecomment-780576223)~ Fixed this
Renamed `SlotWidget.WidgetSlotDelegate` to `SlotWidget.WidgetSlotItemHandler`.
`SlotWidget.WidgetSlotDelegate` is now deprecated.
Now `SlotWidget.getHandle()` may return either `SlotWidget.WidgetSlot` or `SlotWidget.WidgetSlotItemHandler` instead of previous `SlotWidget.WidgetSlotItemHandler` so now it returns `Slot` instead of `SlotItemHandler`.

### DIFF
--- a/src/main/java/gregtech/api/gui/ModularUI.java
+++ b/src/main/java/gregtech/api/gui/ModularUI.java
@@ -10,7 +10,6 @@ import gregtech.api.util.Position;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.InventoryPlayer;
 import net.minecraftforge.items.IItemHandlerModifiable;
-import net.minecraftforge.items.wrapper.PlayerMainInvWrapper;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -192,7 +191,7 @@ public final class ModularUI implements ISizeProvider {
         public Builder bindPlayerInventory(InventoryPlayer inventoryPlayer, TextureArea imageLocation, int x, int y) {
             for (int row = 0; row < 3; row++) {
                 for (int col = 0; col < 9; col++) {
-                    this.widget(new SlotWidget(new PlayerMainInvWrapper(inventoryPlayer), col + (row + 1) * 9, x + col * 18, y + row * 18)
+                    this.widget(new SlotWidget(inventoryPlayer, col + (row + 1) * 9, x + col * 18, y + row * 18)
                         .setBackgroundTexture(imageLocation)
                         .setLocationInfo(true, false));
                 }
@@ -202,7 +201,7 @@ public final class ModularUI implements ISizeProvider {
 
         public Builder bindPlayerHotbar(InventoryPlayer inventoryPlayer, TextureArea imageLocation, int x, int y) {
             for (int slot = 0; slot < 9; slot++) {
-                this.widget(new SlotWidget(new PlayerMainInvWrapper(inventoryPlayer), slot, x + slot * 18, y)
+                this.widget(new SlotWidget(inventoryPlayer, slot, x + slot * 18, y)
                     .setBackgroundTexture(imageLocation)
                     .setLocationInfo(true, true));
             }
--- a/src/main/java/gregtech/api/gui/widgets/SlotWidget.java
+++ b/src/main/java/gregtech/api/gui/widgets/SlotWidget.java
@@ -251,4 +251,59 @@ public class SlotWidget extends Widget implements INativeWidget {
         }
     }
 
+    /**
+     * @deprecated
+     * Use {@link WidgetSlotItemHandler} instead. <br>
+     * {@link WidgetSlotDelegate} was renamed to {@link WidgetSlotItemHandler} since GregTech 1.15.0.<br>
+     * Explanation of deprecation: In order to fix mouse wheel action a new class was introduced ({@link WidgetSlot}).
+     * To have consistent names {@link WidgetSlotDelegate} was renamed to {@link WidgetSlotItemHandler}.
+     *
+     * @see <a href="https://github.com/GregTechCE/GregTech/pull/1485">GregTech#1495 (Pull request)</a>
+     * @see <a href="https://github.com/GregTechCE/GregTech/issues/1291">GregTech#1291 (Issue)</a>
+     */
+    @Deprecated
+    protected class WidgetSlotDelegate extends SlotItemHandler implements IScissored {
+
+        public WidgetSlotDelegate(IItemHandler itemHandler, int index, int xPosition, int yPosition) {
+            super(itemHandler, index, xPosition, yPosition);
+        }
+
+        @Override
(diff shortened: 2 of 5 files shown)


PrototypeTrousers:
Does this affect #1291 and/or #1190 ?

CrafterKolyan:
@PrototypeTrousers by my statical analysis of code I would say that #1190 is party fixed due to the Chests now perform all operations by MouseTweaker, but for Crafting Station I am not sure this is true. Also thanks for pointing to #1291, I've fixed it now with ease. Here are also a video how it is working (first two clicks were with the Right Click, then Shift + Left Click):

https://user-images.githubusercontent.com/9883873/108208302-eab90280-7139-11eb-96c7-ba7d6464ba43.mp4

CrafterKolyan:
One of the problems here is that `SlotDelegate` is actually not a `Delegate` because it `extends Slot` even though it shouldn't. Another possibility is to create multiple classes but I am not sure it's a good idea as that would lead to copy-paste code. We can delete inheritance from `Slot` due to `getHandle()` should return `Slot`. That's why I changed its prototype to show that what you actually get is a `Delegate` and not a real `Slot`.

CrafterKolyan:
My further investigation shows that it seems we can't do anything except of a creating more classes, because otherwise we won't be able to fix Crafting Station (Shift + LMB) because MouseTweaker check for `instanceof` of `Slot` which is actually our `Widget` which is `Delegate`

https://github.com/YaLTeR/MouseTweaks/blob/c980bc11b6218b2371ebf591c3cdef548fd2899e/src/main/java/yalter/mousetweaks/handlers/GuiContainerHandler.java#L116-L121

Archengius:
I'm heavily against making base changes to SlotWidget just because one mod assumes too much about modded inventories.

warjort:
#1460 is another possible issue related to mouse tweaks

CrafterKolyan:
> I'm heavily against making base changes to SlotWidget just because one mod assumes too much about modded inventories.

Well, this mod is not assuming anything, it's just wants to check if `Slot` corresponds to player. Talking about checking for `CraftingOutput` it does it the only way it is possible and it's not their wrong code but our (mostly my) wrong usage of Delegate pattern which externs from the class it delegates.

CrafterKolyan:
Ok, I think now it is ready. I will not fix interaction with Crafting Station fully in this pull request but at least I fixed voiding items and hasn't broken anything else. The main pull request message shows the whole updated information. I think it's a step to fixing #1460  but probably not full solution, of course it's good to provide correct information about whether the inventory corresponds to player or not but this doesn't mean that synchronization will now be ideal.

CrafterKolyan:
Talking about #1460, basically what MouseTweaks does when you do Shift + LMB + Drag is just performing Shift + LMB on everything you drag over:

https://github.com/YaLTeR/MouseTweaks/blob/c980bc11b6218b2371ebf591c3cdef548fd2899e/src/main/java/yalter/mousetweaks/Main.java#L281-L285

So I don't think the problem is with MouseTweaks or GregTech: it's simple Minecraft synchronization (unless we do some strange stuff on click but I haven't found anything wrong in the code by now).

LAGIdiot:
I will have to take really close look at this. I am really happy that our inventories could work better with MouseTweaks and also some of bugs were fixed but that incompatibility is quite big price for it.

DStrand1:
After reviewing this, I think that the changes being made are good, but that we should not rename our base Widgets and change their functionality, as addons are depending on them significantly. I think if we make these changes as new Widgets and leave the old ones as-is but deprecated, that this change is safe compatibility-wise.

ALongStringOfNumbers:
@CrafterKolyan Could you address the above comment?

CrafterKolyan:
@ALongStringOfNumbers ok, will try to do it till Friday evening

CrafterKolyan:
@ALongStringOfNumbers Sorry for being late by a day. I've rebased all changes to the latest `master` and made changes requested by @DStrand1. See 7cd104d.
Now `SlotWidget.WidgetSlotDelegate` and `SlotWidget.WidgetSlotItemHandler` have absolutely the same implementation but `SlotWidget.WidgetSlotDelegate` is deprecated.

ALongStringOfNumbers:
This PR looks good to me now. I will let everyone else review it as well to see if there is something that I missed.

DStrand1 (approved):
Looks good to me, functionality is great and should not have issues with compatibility

LAGIdiot (approved):
I am OK with this level of backward compatibility given what we are trying to fix.

I am playtesting it and so far it looks good.

Issue #1370, opened 2020-12-24
[Suggestion] Multiblock's should have their own OrientedOverlayRenderer's


idcppl:
**Is your feature request related to a problem? Please describe.**
Not really a problem, but streamline multiblock controllers to have their own OrientedOverlayRenderer's like single block machines have. Would help resource pack makers.

**Additional context**
If this is to be accepted, I'll make the PR for it.

idcppl:
I'd still leave the old OrientedOverlay, since addon's use them and wouldn't wanna break those.

LAGIdiot:
I am not sure if I am getting what you are trying to say here. Can you please provide example how exactly could this be used?

idcppl:
Currently all Multiblock controllers share the same [OrientedOverlayRenderer](https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/api/metatileentity/multiblock/RecipeMapMultiblockController.java#L160). If you changed it to each having their own, it would allow for texture pack artist to change how each one looks. It's not like we need new textures, just put the same textures into different OrientedOverlayRenderer's.

LAGIdiot:
I think I understand now. Sure we can go with it. For now we will use same textures for all but we can update them in feature.

PR #1494, opened 2021-02-21
Open Up Multiblock Overlay

DStrand1:
**What:**
This PR is a simple one, that allows subclasses of `RecipeMapMultiblockController` to set their own custom overlay for the multiblock controller.

**How solved:**
I created a protected method `getFrontOverlay()` that is called by `RecipeMapMultiblockController#renderMetaTileEntity()` so that a multiblock controller can have any overlay. It still defaults to `MULTIBLOCK_WORKABLE_OVERLAY` as before.

**Outcome:**
- Allow custom Multiblock Controller overlays

None expected.

### DIFF
--- a/src/main/java/gregtech/api/metatileentity/multiblock/RecipeMapMultiblockController.java
+++ b/src/main/java/gregtech/api/metatileentity/multiblock/RecipeMapMultiblockController.java
@@ -14,6 +14,7 @@ import gregtech.api.capability.impl.MultiblockRecipeLogic;
 import gregtech.api.multiblock.PatternMatchContext;
 import gregtech.api.recipes.Recipe;
 import gregtech.api.recipes.RecipeMap;
+import gregtech.api.render.OrientedOverlayRenderer;
 import gregtech.api.render.Textures;
 import gregtech.api.util.GTUtility;
 import net.minecraft.util.ResourceLocation;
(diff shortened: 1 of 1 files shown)


LAGIdiot:
Is this related to https://github.com/GregTechCE/GregTech/issues/1370?

DStrand1:
As implemented, no, since that would require splitting up the variable `MULTIBLOCK_WORKABLE_OVERLAY` to instead have a variable for each multiblock. This PR instead lets addons like SoG or Gregicality choose to have multiblocks use a completely different OrientedOverlayRenderer instead of the default multiblock one.

If you would like, I can include a fix for #1370 in this, as this is the first step towards doing that.

LAGIdiot:
Ah I see now how it is different. Let's leave the scope like it is.

src/main/java/gregtech/api/metatileentity/multiblock/RecipeMapMultiblockController.java
```diff
@@ -157,6 +158,14 @@ protected boolean checkStructureComponents(List<IMultiblockPart> parts, Map<Mult
...
         super.renderMetaTileEntity(renderState, translation, pipeline);
-        Textures.MULTIBLOCK_WORKABLE_OVERLAY.render(renderState, translation, pipeline, getFrontFacing(), recipeMapWorkable.isActive());
+        this.getFrontOverlay().render(renderState, translation, pipeline, getFrontFacing(), recipeMapWorkable.isActive());
+    }
+
+    /**
+     * Override this method to change the Controller overlay
+     * @return The overlay to render on the Multiblock Controller
+     */
+    protected OrientedOverlayRenderer getFrontOverlay() {
```
LAGIdiot, line 168: Please add @Nonnull to this so it is obvious that it needs to return something, hard null check is not needed in this case.

LAGIdiot (changes requested):

LAGIdiot (approved):
Thanks for finishing this up.

PR #1495, opened 2021-02-21
Remove Duplicate Dense Plate Recipe

DStrand1:
**What:**
This PR removes a duplicate recipe for dense plates, as shown below.
recipe1
recipe2

**How solved:**
I removed the one that took less time, mostly because the other recipe in `PartsRecipeHandler#processDensePlate` felt like the place it made most sense to leave this recipe. If you would rather I adjust its time to be what it was before, I can.

**Outcome:**
- Remove Duplicate Dense Plate Recipes

None expected.


LAGIdiot (approved):
Good job spotting and fixing this this. I would also remove this recipe handle.

PR #1502, opened 2021-02-23
Recipe Properties

ALongStringOfNumbers:
**What:**
This is an attempt to update PR #1371 as it has gone stale, and I could use the implementation for SoG for the Fusion Reactor.

I have attempted to implement the feedback present in that PR, but I feel like I went somewhat in circles, and so could use some feedback on the implementation.

**How solved:**

I have updated `Recipe` to have a new constructor which takes a `RecipeProperty` List, which offers the ability for multiple properties to be applied to a recipe. The old constructor will have the provided `Map<String, Object>` converted into a `RecipeProperty` of the appropriate type, of which there are two at the moment.

`DefaultRecipeProperty`, which does nothing to the display in JEI, and is used to cover cases where an empty Map was passed into Recipe.

`TemperatureProperty`, which is used to deal with the EBF's `blast_furnace_temperature` and get the corresponding coil level.

I did not implement a Property for the Fusion Reactor, as GTCE currently does not implement anything related to the fusion reactor, which is instead done by addon mods, and so I left the addon mods to write their own `EUToStartProperty`.

I could not find any other properties that needed implementing in based GTCE, besides the ones for the Amplifier, which currently does not look to be implemented in GTCE.

The `RecipeProperties` themselves can be created through a String and Object for the ability to covert the old Map implementation, but can also be created in other ways.

**Outcome:**
Creates Recipe Properties, and implements a Recipe Property for the Electronic Blast Furnace

**Additional info:**

The old properties should be compatible, and only new API classes were created other than that.

### DIFF
--- a/src/main/java/gregtech/api/recipes/DefaultRecipeProperty.java
+++ b/src/main/java/gregtech/api/recipes/DefaultRecipeProperty.java
@@ -1,0 +1,21 @@
+package gregtech.api.recipes;
+
+import net.minecraft.client.Minecraft;
+
+/*
+    This Class is used to deal with cases where there was no Recipe Property provided
+ */
+public class DefaultRecipeProperty extends RecipeProperty {
+
+    public DefaultRecipeProperty() {
+    }
+
+    public void drawInfo(Minecraft mc, int x, int y, int color) {
+        mc.fontRenderer.drawString("", x, y, color);
+    }
+
+    @Override
+    public RecipeProperty getRecipeProperty() {
+        return this;
+    }
+}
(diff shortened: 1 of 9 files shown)


ALongStringOfNumbers:
I have been informed that this has resulted in using common code in API, namely `TemperatureProperty` in the Blast Furnace Recipe builder. I can solve this by moving `TemperatureProperty` into the API, but then would need to refactor the coil detection, so that it does not rely on code from common. So I will work on doing that refactor, so that there is no common code used in API.

ALongStringOfNumbers:
Closing this for now because #1371 has had recent progress and is no longer stale

PR #1504, opened 2021-02-24
Progress Bar Updates

DStrand1:
**What:**
This PR has 3 main changes in it, all related to progress bars, each with their own commit.

## Progress Bar Bugs
This section fixes some problems with how our progress bars overlap when the machine is running.

- **Macerator**: The Macerator had a few issues. First of all, the texture is meant to be 21 pixels long, and some textures for it had it as such, and some were still 20. This caused shrink-to-fit issues, so I changed them all to 21. With this change, there were two major outcomes.
    - The final "dust" piece now fills on progress, giving the bar a little more clarity at the end, where before it would not fill.
    - I changed all progress bars for `RecipeMap` and `RecipeMapGroupOutput` to use a 21-pixel length progress bar texture, and adjusted all textures to be 21 wide rather than 20. This has no change for most progress bars, except Macerator.
    The Macerator progress bars were also misaligned, and did not overlap properly on machine progress.

    **Electric Macerators**:
    OLD:
MaceratorOld
    NEW:
MaceratorNew

    **Steam Macerators**:
    OLD:
SteamBronzeOld
SteamSteelOld
    NEW:
SteamBronzeNew
SteamSteelNew

- **Centrifuge**: This progress bar is used in a few places, but this is just one example. This texture was simply missing a pixel towards the bottom right, meaning it had a white pixel misplaced on progress.
OLD:
CentrifugeOld
NEW:
CentrifugeNew

- **Triangle Bar**: This progress bar was also misaligned like the Macerator.
OLD:
TriangleOld
NEW:
TriangleNew

## Progress Bar Swaps
This section swaps the Compressor and Bending Machine progress bars. I can justify this with two reasons:
- They were backwards compared to other versions of GT (and IC2's Compressor, where the asset originates)
- The Bending Machine item input slot has an overlay that looks very similar to the new progress bar I assigned it

New GUI's:
BendingNewGUI
CompressorNewGUI

## Mixer Progress Bar Change
This section adds the 4-color Mixer progress bar from GT5U, because it is a very cool progress bar that I feel like we should have. I did not modify the old progress bar in the code, and only added a new one for the Mixer. This means that the Chemical Bath and Ore Washer have the old blue-only one still, as do any addons that use that progress bar.
mixerNewGUI

If addons add custom progress bars, they will have them slightly compressed due to the progress bar width change (20 -> 21).


DStrand1:
Upon further thought, for section 3, I think the Chemical Bath should have the new 4-color progress bar, but the Ore Washer should be the blue-only one since it only ever uses Water.

LAGIdiot:
Thanks for your hard work. As I am not UI person so I did not even notice these problems. So thanks for fixing them.

Regarding colored progress bar for Mixer...I am not sure on one hand it makes sense as it is signaling mixing of items but on other it is goes against principles of other progress bars. I am bit torn regarding this.

DStrand1:
Well, I would like to make the argument in favor of it that the Mixer progress bar already broke convention, as it was blue rather than white

LAGIdiot:
Hm, I see, let's make precedence from this. **If it makes sense for machine to have crazy colored progress bar than it can.**

Now feel free to update it for Chemical Bath too.

LAGIdiot (approved):

eutro (approved):
This is beautiful.

PR #1511, opened 2021-03-03
Fix the Fuelable TOP provider to do its translation on the client

### LINKED ISSUE #1508: [BUG] TOP - Fuelable Provider Error Message
**Describe the bug**
Looking at a GTCE machine that uses a fuel (Steam Turbine, Diesel Generator, Large Boilers, etc), TOP shows an error message: "Error: gregtech:fuelable_provider"  This appears to be just a TOP display issue and does not appear to affect gameplay

**Versions**
Forge: 14.23.5.2854
GTCE: 1.12.0.662
Addons: The One Probe: 1.4.28

**Setup**
Multiplayer Server, custom lightweight modpack

**Steps To Reproduce**
I viewed Steam Turbine, Diesel Generator, Large Boiler, Gas Turbine and confirmed all are showing the same message.  Confirmed that the message shows up with or without fuel in the machine.  

**Screenshots**
image

warjort:
**What:**
The new Fuelable TOP provider is erroring in dedicated server mode.

**How solved:**
The error is due to it trying to do the translation on the server. This is now done on the client.

**Outcome:**
Closes #1508

**Additional info:**
The TOP progress bar does not allow its "suffix" to have translatable text inside it. 
So I have moved the extra info about burn time and minimum fuel into a seperate text component. 
This means that information now appears further to the right.

The getFuelName() for the Fluid and Item fuel info have been changed to use the "unlocalizedName", so the server can pass it to the client for translation.
SInce this code is new, it is unlikely something else will be using this.
But even if something was using it, it would also need this change to be able to translate the name on the client.


LAGIdiot (approved):
Thanks for fixing this issue. It did not occurred to me that this could happened but at least we know for future reference.

PR #1513, opened 2021-03-03
Distribute Ticks More Evenly

serenibyss:
**What:**
This PR is the second attempt of #1424, but this time is careful to check for if anything depends specifically on first tick behavior. In total, I found 3 examples where first tick behavior may have been important:
- **Multiblock structure forming:** This is what caused this PR to be removed previously. The JEI preview depended on `getTimer() % 20 == 0` in order to form the multiblock in the preview in one tick.
- **Energy Absorber:** There are a couple states tested initially, and then again every so often, for `DragonEggStatus` and `ConnectedCrystals`.
- **Pump:** The pump adds itself to the `blocksToCheck` list initially, and will re-add itself every 5 minutes.
If anyone can spot other cases like this, please let me know, but I think they are all respected now.

**How solved:**
In order to fix these, I changed the definition of the old `getTimer()` to be used strictly for checking first tick behavior. For example, when checking for multiblock formation, it now checks:
`if (getOffsetTimer() % 20 == 0 || getTimer() == 0`
I would have liked to use the `onFirstTick()` method in `TickableTileEntityBase`, but I did not see how to override it properly. If anyone has any ideas as to how to override this method, feel free to suggest.

**Possible Compatibility Issue:**
None expected, as `getTimer()` behavior is identical to before.


LAGIdiot:
I see how you fixed problem we found out last time. But I am not sure if there is not more hidden ones like it. I will take a look at first tick problem implementation.

PR #1514, opened 2021-03-04
Add notConsumable Fluids

serenibyss:
**What:**
This PR simply expands our `notConsumable()` methods to include options for passing in a FluidStack.

**Outcome:**
Fluids can now be used in recipes and not be consumed.

**Additional info:**
image

When testing this with Gregicality, their multis were not able to handle notConsumable fluids, so they will have to update their code to allow for this. Our multis can handle it just fine.

~~ISSUES:~~
- ~~Not consumed fluid does not allow you to use "U" or "R" on it in JEI. This is a limitation of JEI not having an `ISubtypeManager` for fluids with NBT data. See https://github.com/mezz/JustEnoughItems/issues/2262. I am working on a fix for this in JEI.~~

~~Edit:~~
~~Version of this feature is complete and awaiting release for 1.16.4, see https://github.com/mezz/JustEnoughItems/pull/2263. Once merged, the backport to 1.12 will start.~~

The above concerns are now addressed in #1516


mezz:
I think you should try using this callback instead of adding NBT to the item, I think it will be a bit cleaner:
https://github.com/mezz/JustEnoughItems/blob/301ce88364c23d909138d2af8c2dc8c506c85b8b/src/api/java/mezz/jei/api/gui/ingredient/IGuiIngredientGroup.java#L52-L55

LAGIdiot (commented):
This looks like good addition. I think we can find some usage for it and I believe that addons will put it in good use too (maybe even sooner then us).

Regarding your opening comment and Gregicality do I understand it correctly that when recipe contains `notConsumable` the recipe will just not work or it will break it whole mod?

DStrand1:
It should work just fine with this code included, just that if Gregicality's multis attempt to use not-consumable fluids on certain multiblocks, it will fail. But it is a known problem with them, so I don't have any problem fixing it there myself if this is merged.

DStrand1:
> I think we can find some usage for it

There are some reactions that we have, like Titanium Tetrachloride in recipes for some plastics to boost the outputs where the TiCl4 serves as a catalyst. I think this is a prime example of somewhere we could use this feature in GTCE, though it does have balance implications.

LAGIdiot (approved):
Thanks for clarification

PR #1516, opened 2021-03-04
Drop NBT For JEI Recipe Registration

serenibyss:
**What:**
This PR changes our JEI recipe registration to not use NBT data for notConsumed and chanced outputs, and instead store them internally for the tooltips.

**How solved:**
I changed `GTRecipeWrapper#getIngredients()` method to not put NBT data on these items, and instead store them in some custom hashing data structures to use later for adding the JEI tooltips.

**Outcome:**
- Remove NBT data from recipes

**Additional info:**
These screenshots show one example of a recipe that formerly had NBT data attached, but no longer do:
Before:
image

After:
image

Here is an example of chanced and notConsumed recipes, with no tags:
image
image

Some CraftTweaker scripts will break, but it is for the best, as having to modify NBT in CT is never a good time.

### DIFF
--- a/src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
+++ b/src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
@@ -79,7 +79,13 @@ public class GTRecipeWrapper implements IRecipeWrapper {
                 ItemNBTUtils.setInteger(chancedStack, "boost_per_tier", chancedEntry.getBoostPerTier());
                 recipeOutputs.add(chancedStack);
             }
-            recipeOutputs.sort(Comparator.comparing(stack -> ItemNBTUtils.getInteger(stack, "chance")));
+
+            recipeOutputs.sort(Comparator.comparing(stack -> {
+                NBTTagCompound tag = stack.getTagCompound();
+                if (tag != null)
+                    return ItemNBTUtils.getInteger(stack, "chance");
+                else return 0;
+            }));
             ingredients.setOutputs(VanillaTypes.ITEM, recipeOutputs);
         }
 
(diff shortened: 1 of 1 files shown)


DStrand1:
I reworked this PR significantly from its initial posting. It now also addresses the JEI concerns of non-consumable fluids in #1514

Exaxxion:
Have you tested the tooltips against CraftTweaker RecipeMap chance scaling functions?

DStrand1:
Exa and I talked about this issue outside of this issue thread, and we have determined that the above mentioned possible issue is not an issue with my new implementation, but more an issue with the system.

I don't really see any way to update the tooltip to the chance function, as it could be any mathematical formula, so I think we just have to keep it as "Chance: X% + Y% per tier"

Exaxxion:
Yes, it would not be feasible with the current API. Fortunately the tooltip may be overrided using a langfile entry to display the desired information, and that's good enough for my use case.

LAGIdiot:
Can you please explain why would TC scripts break after implementing this? I can't piece it together.

DStrand1:
Here is a random example of a recipe removal in a script of mine:
`unpacker.findRecipe(12, [<gregtech:meta_item_1:2996>,<gregtech:meta_item_1:32766>.withTag({Configuration: 1, not_consumed: 1 as byte})], null).remove();`

Following this change, I believe that the tag configuration `not_consumed: 1 as byte` will cause this recipe to not properly remove the recipe anymore. Though, I haven't tested it, so it may still work, and simply now be an unnecessary addition to the `withTag()` call.

LAGIdiot:
I have checked that potential compatibility problem and I don't think it is applicable as script with line as defined in your comment will still work correctly.

LAGIdiot (approved):
Thanks for your hard work.

PR #1517, opened 2021-03-07
Adjust the EU/t of the Rotors

### LINKED ISSUE #1491: [Suggestion] Move Rotor Mold Recipes to LV
After implementing the Rotor Mold, there was a slight unintended side effect of its development process. Since it was originally implemented as an Extruder Shape, I had its EU/t set at MV. However, now that it is a Mold (and the only machine recipe for rotors), I think this should be moved to LV. Thoughts?

Additionally, here is the issue where this discussion first came up:
https://github.com/OmnifactoryDevs/Omnifactory/issues/696

ALongStringOfNumbers:
**What:**
Adjusts the Eu/t of Rotor Recipes to bring them into LV

**How solved:**
Adjusts the EU/t of the recipe

**Outcome:**
Adjusts the EU/t of the Rotor Solidification recipes so that it is LV. Closes #1491.

This could mess up CT scripts that adjusted the EU/t on their own, but those scripts would most likely not even be needed after this.


LAGIdiot (approved):
This seems reasonable. Thanks for updating it.

PR #1506, opened 2021-02-25
Remap MCP Mappings

serenibyss:
**What:**
This PR addresses the issue discussed in Discord earlier,
> We ran into a problem with the JEI OrePage where it would crash sometimes, and not consistently. I found out that it was due to the fact that Biome#biomeName changed at some point from public to private access, and an IllegalAccessException was being thrown due to it being private.
> As a result, we think the safest way forward is to update our MCP mappings from snapshot_20170928 to stable_39. I believe that I have fixed all mapping changes in our files (a total of 65 files changed), but I am still working on fixing our build.gradle so that all dependencies compile and run correctly in dev environment.

**How solved:**
I updated our MCP mappings to `stable_39`. Notable method name changes:

**Block:**
- getBlockLayer() -> getRenderLayer()
- onBlockDestroyedByPlayer() -> onPlayerDestroy()
- onEntityCollidedWithBlock() -> onEntityCollision()
- getMobilityFlag() -> getPushReaction()

**World:**:
- getChunkFromChunkCoords() -> getChunk()

**ResourceLocation:**
- getResourceName() -> getNamespace()
- getResourcePath() -> getPath()

**NBTCompound**:
- hasNoTags() -> isEmpty()

**Translation:**
- setUnlocalizedName() -> setTranslationKey()
- getUnlocalizedName() -> getTranslationKey()

**EnumFacing**:
- getFrontOffsetX() -> getXOffset()
- getFrontOffsetY() -> getYOffset()
- getFrontOffsetZ() -> getZOffset()

**CreativeTabs:**
- getTabIconItem() -> createIcon()

**EnumRarity:**
- rarityColor -> color

**Biome:**
- `biomeName` is private, so I changed its references to `getBiomeName()`

**Outcome:**
Update MCP Mapping to stable_39

You can verify my work using the official [MCP Mapping names](https://mcp.thiakil.com/#/search) website, which is what I used to guarantee that we switch to the correct new method names.

**Additional info:**
I have not yet gotten the dev environment working without a crash due to dependencies deobfuscated files being outdated. I know that setting this up is possible, as both SoG and Gregicality are already on `stable_39`, and have successfully remapped SRG->MCP for all of their dependents (same as ours in dev, and more).

**Possible compatibility issue:**
There may be some work needed for developers to refresh their environments, but I am not sure of the extent of it since I haven't finished the build.gradle

PR #1518, opened 2021-03-07
Adjust Biome Name tooltip to use getter

ALongStringOfNumbers:
**What:**
This PR brings the important fix of changing the Biome Name tooltip to use the biome name getter out of PR #1506. This is done so that it can be merged without having the finalize the issues in that PR and reviewing the entire PR.

**How solved:**
In future MCP mappings, the `biomeName` field changed to private, which caused some players to get access errors when viewing the tooltip on the ore spawn page. This PR simply switches from using the field to using the getter.

**Outcome:**
Fixes crash when viewing tooltip on JEI Ore page.

There should be none


LAGIdiot (approved):
Thanks for fixing this issue.

PR #1522, opened 2021-03-08
Partial overclocking

### LINKED ISSUE #1498: Overclocking Button Enhancement
You cannot choose what overclocking voltage a machine runs at, only current tier voltage and base speed with no overclocking. When dealing with machines like the centrifuge, a higher tier machine gives a higher chance at bonus/chance based resultant products, which further exacerbates the issue of not being able to pick the voltage tier a machine can run at.

What would be nice is replacing the overclocking toggle switch with a button that cycles through available voltage tiers for that machine (much like EnderIO conduits and their channel colors), possibly including a popup text field that indicates what voltage tier it is set to and/or the voltage range.  This would minimize UI space used, while providing more functionality.

warjort:
**What:**
Allow the maximum voltage tier for overclocking to be configured on simple machines.

**How solved:**
A new parameter is introduced into AbstractRecipeLogic called "overclockVoltage". 

A call to enableOverclockVoltage() sets this new overclock voltage to getMaxVoltage() and the new behaviour is enabled.
The user/machine can now change the overclockVoltage and it will be used where maxVoltage was previously used in the overclock logic.

SimpleMachineMetaTileEntity and MetaTileEntityMacerator have been changed to enable this behaviour.

Additionally calls to setOverclockVoltage() or setOverclockTier() will enable the new behaviour.

The SimpleMachineMetaTileEntity now has a  CycleButtonWidget for the overclock configuration. This lets the user choose the overclock voltage between OFF and the max voltage tier of the machine.

**Outcome:**
Closes #1498

Many of the SoG and Gregicality machines should get this new behaviour. The exceptions will be Gregicality's UH* machines and any machines that override the createWorkable() method.
I have tried to make the change to enable this behaviour as easy as possible (even for the extra energy tiers) by adding and using some helper methods to AbstractRecipeLogic that can be overridden.

For existing worlds, I have provided an automatic "data fix". If overclocking was allowed before, the new voltage will be set to maxVoltage of the machine. If it was dissallowed, it will be set to OFF.


DStrand1:
Looks great! As far as Gregicality UHV+ machines, I don't think that will be difficult to fix.

With the button, are you able to right-click, or shift-click or something to have it cycle in reserse? I can see myself trying to set a machine to LuV, and accidentally clicking one too many times, and now I have to cycle through all tiers again.

warjort:
The CycleButtonWidget only allows cycling in one direction. see CycleButtonWidget::mouseClicked()

I would suggest a separate PR for adding a right click behaviour?

LAGIdiot:
Thanks for your hard work on this PR. I like general idea and that is reason why I engaged positively on connected issue. But I am not sold on result. I don't like idea about machine being always in overclock mode. I think players should be able to turn it off. For example before you could run 2EU/t recipe without overclock but now you can't. So I would like to see off as option in Cycle button. This could also simplify logic on backend as you could start with `overclockVoltage` set to voltage of machine and don't have to handle -1 and call `enableOverclockVoltage`.

warjort:
The purpose of the -1 is so the GUI knows whether the implementing machine understands the new behaviour.

If they have overriden the Workable they could have custom logic for overclocking. 
Or they may need to do extra work like the new energy tiers in Gregicality.
So I wanted them to explicitly enable it before the more fine grained control is exposed in the GUI.

warjort:
It didn't occur to me that you can still overclock within the ULV tier.
Obviously now I think about it, 8/4 =2

What about if we allow the overclockVoltage to be 0 (zero) - meaning truly off
I am trying to keep it simple so that we don't have to keep multiple flags in sync which will be prone to breakages.

This would then only require a couple of simple changes
````
    protected int[] calculateOverclock(int EUt, long voltage, int duration) {
 -       if(!allowOverclocking) {
+        if(!allowOverclocking || voltage == 0) {
            return new int[] {EUt, duration};

                // Calculate overclock voltage based on old allow flag
                if (this.allowOverclocking) {
                    this.overclockVoltage = getMaxVoltage();
                } else {
-                    this.overclockVoltage = getVoltageByTier(0);
+                    this.overclockVoltage = 0;
                }

````

The question comes in how to express this in the GUI. The new button is a text widget, but its not very wide because it has to fit in the same space as the old button.
Should it just display blank for off? Or some other value like "--"

I think "None" or "Off" might translate to something too long is some languages?

**Comment by LAGIdiot** [comment]
I am aware that they probably have override for overclocking mechanic. So nothing should change for them in this case it will be called same way (with different values but same way). Also you are enabling this value on `SimpleMachineMetaTileEntity` so as long they also don't override `createWorkable` they will get new implementation. So this won't save them.

I don't think we can truly do this without some consequences so I will mark this as Minor. We should provide addons with way how to use this new system but we should not block ourselves because they choose to implement additional energy systems and inject them to our system.

**Comment by LAGIdiot** [comment]
Let's make button for 3 chars wide we will have **OFF** in and other languages can out in X for example. With better description like "OFF means no overclocking is done and rest of values means that recipe will be overclocked to given value if initial price allows it" or something I am not good at creating user messages.

**Comment by warjort** [comment]
> Also you are enabling this value on SimpleMachineMetaTileEntity so as long they also don't override createWorkable they will get new implementation. So this won't save them.

But they do override it for the new energy tiers. This is why I am using it to detect when not enable the new behaviour.

My assumption is that if they haven't overridden the workable they won't have/need custom overclock logic?

**Comment by warjort** [comment]
> Let's make button for 3 chars wide we will have OFF in and other languages can out in X for example. 

Ok, it can already handle 3 characters, e.g. ULV.

**Comment by LAGIdiot** [comment]
It seems that they are not using `SimpleMachineMetaTileEntity` for higher tiers as they don't have way how to override our 
tier system so we should not be concerned about this change and possibility with breaking them up. As it won't happened. https://github.com/Gregicality/gregicality/blob/a468fc8080500c83f18a450b41ac81f3df28abe4/src/main/java/gregicadditions/machines/GATileEntities.java#L270`

**Comment by DStrand1** [comment]
Yes, for Gregicality it is a pretty hard rewrite of many parent classes to get around GTValues, among some other things, so if they want this functionality it would have to be implemented separately unfortunately

**Comment by warjort** [comment]
Wouldn't it be easier to make a PR that;

* changes all access to GTValues within GTCE to use functions
* makes the functions use new mutable data structures
* initialise the new mutable data structures from the existing data structures
* adds functions to change the new data structures
* deprecates the old data structures

GTCE will use the mutable datastructures via the functions
Gregicality can add its new energy tiers and use the new functions
Other addons like SoG can continue to use the old "static" data structures

**Comment by warjort** [comment]
I have made the changes discussed above.

**Comment by LAGIdiot** [comment]
> Wouldn't it be easier to make a PR that;
> 
> * changes all access to GTValues within GTCE to use functions
> * makes the functions use new mutable data structures
> * initialise the new mutable data structures from the existing data structures
> * adds functions to change the new data structures
> * deprecates the old data structures
> 
> GTCE will use the mutable datastructures via the functions
> Gregicality can add its new energy tiers and use the new functions
> Other addons like SoG can continue to use the old "static" data structures

This would be discussion for other day. But I am open for suggestions or collaboration requests from addons (of course I would not break GTCE for them). But it seems it is easier for them to hack over us then come to us with reasonable requests...

**Comment by LAGIdiot** [comment]
I am still seeing that you are still working with `overclockVoltage = -1` and need to enabled it. I think we can do it without it and use `overclockVoltage` directly with default value of max voltage for given tier. As we already confirmed that it wont break anyone. Same goes for GUI I don't feel there is need to check that state to show either button or cycle button.  Also I think we should continue using `allowOverclocking` as valid state not just adding additional check for `overclockVoltage == 0`.

**Comment by warjort** [comment]
I am not convinced we understand how every possible machine in other addon mod creates custom workables.

And what about other users of AbstractRecipeLogic? e.g. multiblocks
Understanding how they calculate their maxVoltage and work with the overclocking is bound to introduce complications.

I'd rather just let them use the old logic until somebody introduces overclocking into their guis and needs to do the analysis.
I don't want another feature creep PR like the fix for pipes that's still not accepted.

**Comment by warjort** [comment]
> Also I think we should continue using allowOverclocking as valid state not just adding additional check for overclockVoltage == 0

It would be great if we could get rid of the boolean altogether and use 0/maxVoltage as on/off
but allowOverclock is a protected field so subclasses could be changing it (even though they should really use the methods otherwise the changes won't be saved in the tile entity).

Anyway, this all boils down to wether the -1 is needed, I still think it is.

**Comment by LAGIdiot** [comment]
I am not convinced that -1 is needed as overclocking function will still be transparent for everyone and nothing will change in that matter. It will still be called with same values if you initialize `overclockVoltage` with `getMaxVoltage` and `allowOverclocking` will handle cases when overclocking is off. Also I insist on leaving `allowOverclocking` in system and it should be correctly set on incoming changes from GUI. As you mentioned it can be overwritten in subclasses so we need to leave it with working functionality.

What I have in mind seems as transparent change to me which should not broke anything in GTCE or anyone who depends on it. I can created new PR based on yours with implemented changes as prototype.

Regarding your remark for fix pipes PR I have transferred that case to Arch so please pick a beef with him on it.

**Comment by warjort** [comment]
I have gone half way, by trying to keep the two flags in sync when overrideVoltage is not -1

If you insist one me removing the -1, I will just abandon this PR because I don't feel happy about potentially breaking things I haven't analysed.

Somebody else can ressurect it if they have the time to do the analysis or are prepared to do the changes without the due diligence.

**Comment by LAGIdiot** [comment]
-1 and and special handling like this signalize to me that there is something wrong. It may be poor design of previous functionality current functionality not being fully coherent or fear of consequences (which again leads to bad design).

By adding this special handling there is more code which needs to be maintained, tested (and fixed in case of problems). This leads to more uses and possible missuses and makes it more coupled.

I would like to request external observer to take a look at it.

I fully understand your feeling and I respect that you are open with information about abandoning this if I push you in direction you don't want.

**Comment by warjort** [comment]
Let me more explicit about what my worries are, because I am not sure it is understood.

Although there is a lot of discussion above about the flags and gui, the real meat of this PR is this function I added to AbstractRecipeLogic
````
    protected int[] calculateOverclock(int EUt, int duration) {
        if (this.overclockVoltage == -1) {
            return calculateOverclock(EUt, getMaxVoltage(), duration);
        } else {
            return calculateOverclock(EUt, overclockVoltage, duration);
        }
    }
````

This says if we have been told the overclock voltage cap, then use it, otherwise we will do the old behaviour of asking the workable for its max voltage and use that as the cap.

But notice the getMaxVoltage() is dynamic. It can be laziliy constructed or even dynamically changed.
In practice this is determined by the energy container plugged into the workable.

To avoid making this a major change, I restricted this PR to those machines that have a known enery cap at construction time.
This just happens to coincide with the simple machines that have a button to toggle the overclock behaviour which was the original goal of this PR.

So I need a flag of some kind to differentiate the two cases.  I chose that -1. 
I also thought this was important to use in GUI because of things like the new voltage tiers in Gregicality, but this seems less relevant. But there could still be cases where the workable is changed in ways that need the old toggle button.

Now it could be my fears are unfounded. All the the max voltages can be reliably determined at workable construction.
But determining that is the case is a lot of digging and testing in the addons, which I don't have the time or inclination to do, and I could still get wrong because I missed a subtlety.

Instead I left the old behaviour in place, leaving it to the addon developers whether they want to take advantage of the new behaviour. They could even still have a dynamic max voltage and configurable overclock voltage if they also add relevant calls to setOverclockVoltage() when it changes.

All I am saying is that this PR features creeps into a full analysis of all the overclocking behaviour across the GT family just to avoid that -1, I will leave it somebody else to finish.

File: src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java [hunk 1]
```diff
@@ -237,8 +238,16 @@ protected boolean setupAndConsumeRecipeInputs(Recipe recipe) {
             recipe.matches(true, importInventory, importFluids);
     }
 
+    protected int[] calculateOverclock(int EUt, int duration) {
```
TedKnudsen, line 241: Returning an array to represent an arbitrary slice of data (where element 0 is one thing and element 1 is some other thing) feels really bad to me.  Is this idiomatic?

src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
```diff
@@ -224,7 +225,7 @@ protected static boolean areItemStacksEqual(ItemStack stackA, ItemStack stackB)
     }
 
     protected boolean setupAndConsumeRecipeInputs(Recipe recipe) {
-        int[] resultOverclock = calculateOverclock(recipe.getEUt(), getMaxVoltage(), recipe.getDuration());
+        int[] resultOverclock = calculateOverclock(recipe.getEUt(), recipe.getDuration());
```
pyure, line 228: Returning an array to represent an arbitrary slice of data (where element 0 is one thing and element 1 is some other thing) feels really bad to me. Is this idiomatic?

src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
```diff
@@ -237,8 +238,16 @@ protected boolean setupAndConsumeRecipeInputs(Recipe recipe) {
             recipe.matches(true, importInventory, importFluids);
     }
 
+    protected int[] calculateOverclock(int EUt, int duration) {
+        if (this.overclockVoltage == -1) {
```
pyure, line 242: This would feel slightly less magical if we were just testing for a value > 0

src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
pyure: (It appears it is, nevermind)

warjort: It is a list of string options. The first is a special value "OFF" and comes from the translation file, 
the others are idiomatic (they are the known enery tier names, I am not sure this is what you meant?) and are not translated.

src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
warjort: you mean >= 0
0 is that special case value of OFF
-1 is the infamous not configured value

warjort: You mean >= 0

0 means OFF
-1 is the contentious not configured value

pyure:
Would it help if we stored overclockVoltage as a (nullable) reference type?  If we were storing overclockAttributes or some such instead, we could pass around null when its not used, get rid of the magic number, and it would be extensible if we needed to add and factor in more attributes.

warjort:
I originally thought about making it a supplier. But I didn't want to over engineer it.

````
public interface VoltageSupplier() {
    long getVoltage();
} 
````

Then the two cases could be differentiated by
````
private VoltageSupplier overClockVoltagePolicy;

    protected int[] calculateOverclock(int EUt, int duration) {
            return calculateOverclock(EUt, overclockVoltagePolicy.get(), duration);
    }

() -> getMaxVoltage(); // old case
() -> getOverClockVoltage(); // new case

and even 
() -> some other rule
````

**Comment by warjort** [comment]
It would make the GUI and NBT code more ugly and problematic unless that overclock policy also contained code for that usage.

e.g.
````
booelan isToggle(); // or something
````

**Comment by warjort** [comment]
You will be pleased to here I have got rid of the -1 and replaced it with the overclock policy described above.

Instead I have replaced it with a documentation/api usage issue.
````
    /**
     * Sets the overclocking policy to use getOverclockVoltage() instead of getMaxVoltage()
     * and initialises the overclock voltage to max voltage.
     * The actual value will come from the saved tag when the tile is loaded for pre-existing machines.
     *
     * NOTE: This should only be used directly after construction of the workable.
     * Use setOverclockVoltage() or setOverclockTier() for more a dynamic use case.
     */
    public void enableOverclockVoltage() {
        setOverclockVoltage(getMaxVoltage());
    }
````

The other changes are:
* Any call to enableOverclockVoltage(), setOverclockVoltage() or setOverclockTier() will enable the new policy.
* The overclockVoltage is now always saved in the tile's tag.
* I have removed the option for the old toggle button from simple machines altogether.

**Comment by warjort** [comment]
Some additional notes:

I have checked SoG and Gregicality and neither overrides the workable for a simple machine, with the exception of the new voltage tiers in Gregicality which don't use GTCE's simple machine.
Incidently, neither uses the GTCE macerator for higher tier macerators, they just use a simple machine.

If there is such a machine somewhere it will need to be changed to call enableOverclockVoltage() otherwise the new button
will work as an on/off button regardless of the tier selected by the user. So no real loss of functionality, just some confusion.

I have kept the new overclockPolicy as private implementation. So if somebody in future wants to expand this into a new api they won't have any real backward compatibility issues to deal with.

**Comment by warjort** [comment]
> If there is such a machine somewhere it will need to be changed to call enableOverclockVoltage() otherwise the new button
will work as an on/off button regardless of the tier selected by the user. So no real loss of functionality, just some confusion.

Well actually that is not quite true. It will appear to work properly until the tile is unloaded and then it will go back to working as an on/off button when the tile is reloaded.

File: src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java [hunk 4]
```diff
@@ -414,6 +477,12 @@ public void deserializeNBT(NBTTagCompound compound) {
         if(compound.hasKey("AllowOverclocking")) {
             this.allowOverclocking = compound.getBoolean("AllowOverclocking");
         }
+        if (compound.hasKey("OverclockVoltage")) {
```
LAGIdiot, line 480: Can you make this constants so we don't have to have 3 literals with same text. And if possible do same for `AllowOverclocking` (I know you did not added but when you are working on it...)

src/main/java/gregtech/api/metatileentity/SimpleMachineMetaTileEntity.java
```diff
@@ -336,9 +345,9 @@ public void clearMachineInventory(NonNullList<ItemStack> itemBuffer) {
                 .setTooltipText("gregtech.gui.fluid_auto_output.tooltip"));
         }
 
-        builder.widget(new ToggleButtonWidget(rightButtonStartX, 60, 20, 20,
-            GuiTextures.BUTTON_OVERCLOCK, workable::isAllowOverclocking, workable::setAllowOverclocking)
-            .setTooltipText("gregtech.gui.overclock"));
+        builder.widget(new CycleButtonWidget(rightButtonStartX, 60, 20, 20,
```
LAGIdiot, line 348: Overclock button should be at least 1 point (maybe more) wider so 3 letters can be rendered without touching border on every GUI scale.

LAGIdiot (changes requested):
Now this looks lot better to me and I am glad I was able to get you to updated it to state I am able to accept it. I would even go bit further with it but I am not gonna break you for it.

I have tested functionality, saving, upgrading world and it works as expected.

There are only 2 small things to adjust and I will accept it.

warjort:
I have made your requested changes.

I was a bit loathe to change the size of the overclock button, because its position is changed dynamically by the recipemap.
````
            rightButtonStartX -= ((RecipeMapWithConfigButton) workable.recipeMap).getRightButtonOffset();
````

Anyway, I increased the width by 4 pixels to ensure the text has at least 1 pixel clearance from the bevel on both sides for "OFF" and "ULV".

LAGIdiot (approved):
This looks much better now, thanks for finishing this up.

PR #1523, opened 2021-03-08
Allow using Right mouse button to cycle some widgets in reverse

ALongStringOfNumbers:
**What:**
Brought up in #1522, this PR allows for cycling `CycleButtonWidgets` in reverse by clicking them with the Right Mouse Button.

**How solved:**
Checks if the button used for the event is the RMB, and if it is cycles the options in reverse, wrapping from the first to the last option if needed.

Note, I check for if `button == 1` because during my testing that was what the RMB was. If there is a better way to detect if it was the RMB that was pressed, please let me know

**Outcome:**
Allows for cycling some widgets in reverse using the right mouse button

None expected

### DIFF
--- a/src/main/java/gregtech/api/gui/widgets/CycleButtonWidget.java
+++ b/src/main/java/gregtech/api/gui/widgets/CycleButtonWidget.java
@@ -132,7 +132,14 @@ public class CycleButtonWidget extends Widget {
     public boolean mouseClicked(int mouseX, int mouseY, int button) {
         super.mouseClicked(mouseX, mouseY, button);
         if (isMouseOverElement(mouseX, mouseY)) {
-            this.currentOption = (currentOption + 1) % optionNames.length;
+            //Allow only the RMB to reverse cycle
+            if(button == 1) {
+                //Wrap from the first option to the last if needed
+                this.currentOption = currentOption - 1 < 0 ? currentOption + optionNames.length - 1 : currentOption - 1;
+            }
+            else {
+                this.currentOption = (currentOption + 1) % optionNames.length;
+            }
             writeClientAction(1, buf -> buf.writeVarInt(currentOption));
             playButtonClickSound();
             return true;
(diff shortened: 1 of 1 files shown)


src/main/java/gregtech/api/gui/widgets/CycleButtonWidget.java
```diff
@@ -132,7 +132,14 @@ public void readUpdateInfo(int id, PacketBuffer buffer) {
     public boolean mouseClicked(int mouseX, int mouseY, int button) {
         super.mouseClicked(mouseX, mouseY, button);
         if (isMouseOverElement(mouseX, mouseY)) {
-            this.currentOption = (currentOption + 1) % optionNames.length;
+            //Allow only the RMB to reverse cycle
+            if(button == 1) {
```
LAGIdiot, line 136: Please make this one constant for readability sake

src/main/java/gregtech/api/gui/widgets/CycleButtonWidget.java
```diff
@@ -132,7 +132,14 @@ public void readUpdateInfo(int id, PacketBuffer buffer) {
...
+                //Wrap from the first option to the last if needed
+                this.currentOption = currentOption - 1 < 0 ? currentOption + optionNames.length - 1 : currentOption - 1;
```
LAGIdiot, line 138: This seems bit overcomplicated to me, so there is simplification
```suggestion
                this.currentOption = currentOption == 0 ? optionNames.length - 1 : currentOption - 1;
```

LAGIdiot (changes requested):
Good idea behind this PR. I have found few places that needs adjustments

src/main/java/gregtech/api/gui/widgets/CycleButtonWidget.java
ALongStringOfNumbers: I have moved this to a constant

src/main/java/gregtech/api/gui/widgets/CycleButtonWidget.java
ALongStringOfNumbers: I have applied this change.

LAGIdiot (approved):
Thanks for finishing this up

Issue #1425, opened 2021-01-25
[Suggestion] Rotor Holder is Dependent on Controller to Function


DStrand1:
**Is your feature request related to a problem? Please describe.**
Since the current implementation of the rotor holder is only for use in turbines, the code for it is heavily dependent on the turbine controller, rather than the controller simply interfacing with the rotor holder to get and set its speed values.

**Describe the solution you'd like**
Open up an interface for the rotor holder so that the speed up/speed down values aren't hardcoded and allow for it to be used in other multiblocks.

**Additional context**
This would allow us to do much more flexible things with the rotor holder, since right now, it is hard coded to increment by 1 and decrement by 3, and depends on the turbine controller. By opening this up, we could have secondary conditions that effect rotor speed, can allow rotor holders to be used in other multiblocks (including MBT), and maybe even use the rotor speed as a secondary condition to recipes like how EBF coils are (though this is a big step).

Relevant section of code: https://github.com/GregTechCE/GregTech/blob/4dd77f6d1f9d8a848a1836f3c7a2fb812880a748/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityRotorHolder.java#L69-L87

This code is changed slightly with PR #1422, but the same problems described here are still present.

LAGIdiot:
Thank you for your suggestion. This seems like good idea which could help GTCE in future and addons maybe even right now. We should be able to implement it.

DStrand1:
I plan on addressing this issue, but since PR #1422 is modifying the same code, I either want to wait to implement this until that PR is merged, or include it as part of that PR. Either way works for me, if you would like I can include it in that PR, otherwise I will wait.

LAGIdiot:
I would rather see this as separate PR so let's leave this for future release

DStrand1:
Adding this here mostly so that I do not forget when it comes time to work on this issue.

Turbine Rotor's themselves seem to only take durability damage when the rotor speed is changing. When it is at maximum speed, no damage will ever be taken.

ALongStringOfNumbers:
This should have been mostly address in #1527. Was there anything in this issue that was not addressed in that PR?

DStrand1:
Closing since this was addressed in PR #1527 but this issue was not linked to it

PR #1527, opened 2021-03-12
made Rotor holder independant from large turbine controller

galyfray:
**What:**
Implement #1425 

**How solved:**
Create new abstract type `TurbineMultiblockController` which extend the `FueledMultiblockController` and refactor code to make the Rotor holder depending on this class.
added two methods `getRotorIncrementSpeed()` and `getRotorDecrementSpeed()` allowing the variation of the rotor speed to not be hard coded.

**Outcome:**
make `MetaTileEntityRotorHolder` depend on `TurbineMultiblockController` instead of `MetaTileEntityLargeTurbine`

**Additional info:**
Multiblocks using turbines are assumed to be energy producer changing this mean recoding some part  of `FueledMultiblockController` in `MetaTileEntityLargeTurbine` or adding another abstract type in between. 
As both options (implemented and presented ) seems good to me tell me if you found that one way is better than another.

None

### DIFF
--- a/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityRotorHolder.java
+++ b/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityRotorHolder.java
@@ -15,7 +15,7 @@ import gregtech.api.render.Textures;
 import gregtech.api.unification.material.type.IngotMaterial;
 import gregtech.common.items.behaviors.TurbineRotorBehavior;
 import gregtech.common.metatileentities.multi.electric.generator.MetaTileEntityLargeTurbine;
-import gregtech.common.metatileentities.multi.electric.generator.TurbineMultiblockController;
+import gregtech.common.metatileentities.multi.electric.generator.RotorHolderMultiblockController;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.item.ItemStack;
@@ -77,7 +77,7 @@ public class MetaTileEntityRotorHolder extends MetaTileEntityMultiblockPart impl
             this.frontFaceFree = checkTurbineFaceFree();
         }
 
-        TurbineMultiblockController controller = (TurbineMultiblockController) getController();
+        RotorHolderMultiblockController controller = (RotorHolderMultiblockController) getController();
         boolean isControllerActive = controller != null && controller.isActive();
 
         if (!isHasRotor()) {
--- a/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityRotorHolder.java
+++ b/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityRotorHolder.java
@@ -83,13 +83,9 @@ public class MetaTileEntityRotorHolder extends MetaTileEntityMultiblockPart impl
         if (!isHasRotor()) {
             resetRotorSpeed();
         } else if (currentRotorSpeed < maxRotorSpeed && isControllerActive) {
-            incrementSpeed(controller.getRotorIncrementSpeed());
+            incrementSpeed(controller.getRotorSpeedIncrement());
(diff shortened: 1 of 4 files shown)


DStrand1:
I think that the new parent class should be named `RotorHolderMultiblockController`, or something like it, since as it is, it may not technically be connected to the Turbine anymore, it still looks like it is from class names alone. This way it would be clearer to addons that this class is free game to use for new multiblocks.

galyfray:
yeah the current name is not really good I'll change it

src/main/java/gregtech/common/metatileentities/multi/electric/generator/TurbineMultiblockController.java
```diff
@@ -0,0 +1,57 @@
+package gregtech.common.metatileentities.multi.electric.generator;
+
+import gregtech.api.metatileentity.multiblock.MultiblockAbility;
+import gregtech.api.recipes.machines.FuelRecipeMap;
+import gregtech.common.metatileentities.electric.multiblockpart.MetaTileEntityRotorHolder;
+import net.minecraft.util.ResourceLocation;
+
+import java.util.List;
+
+public abstract class TurbineMultiblockController extends FueledMultiblockController {
```
LAGIdiot, line 10: In this case more universal name would be better as it does not have to be tied with Turbine anymore.

src/main/java/gregtech/common/metatileentities/multi/electric/generator/TurbineMultiblockController.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * @return true if structure formed, workable is active and front face is free
+     */
+    public boolean isActive() {
+        return isTurbineFaceFree() && workableHandler.isActive() && workableHandler.isWorkingEnabled();
+    }
+
+    public abstract int getRotorIncrementSpeed();
```
LAGIdiot, line 54: Here it would make more sense to me to use wording `getRotorSpeedIncrement` similarly for decrement

src/main/java/gregtech/common/metatileentities/multi/electric/generator/TurbineMultiblockController.java
```diff
@@ -0,0 +1,57 @@
...
+        super.invalidateStructure();
+    }
+
+    /**
+     * @return true if turbine is formed and it's face is free and contains
+     * only air blocks in front of rotor holder
+     */
+    public boolean isTurbineFaceFree() {
+        return isStructureFormed() && getAbilities(ABILITY_ROTOR_HOLDER).get(0).isFrontFaceFree();
+    }
```
LAGIdiot, line 45: Naming here should be more universal too. But there is one implication, we need to provide function with this name on `MetaTileEntityLargeTurbine` level (with just call to this and marked as obsolete) for compatibility reasons.

LAGIdiot (changes requested):
Good work on this PR. That extraction looks solid there is just few comments mostly regarding naming.

src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityRotorHolder.java
```diff
@@ -76,15 +77,19 @@ public void update() {
...
 
         if (!isHasRotor()) {
             resetRotorSpeed();
         } else if (currentRotorSpeed < maxRotorSpeed && isControllerActive) {
-            incrementSpeed(1);
+            incrementSpeed(controller.getRotorIncrementSpeed());
         } else if (currentRotorSpeed > 0 && !isControllerActive) {
-            incrementSpeed(-3);
+            if(controller == null){
+                incrementSpeed( maxRotorSpeed/10);
```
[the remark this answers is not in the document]
galyfray: Fixed this even if I'm not comfortable with potential NPE that could be caused by interaction with other mods like Mekanism cardboard boxes

LAGIdiot (approved):
Thanks for finishing this up

Issue #1141, opened 2020-07-12
[Suggestion] Fluid voiding


LAGIdiot:
**Idea:**
There is no way how to void fluids via GTCE alone and ways from other mods has to be utilized
Thanks @Saereth for pointing it out in #1131 

**Concept:**
1. Cover which will void fluids coming from machine output slot (with slot for fluid filter)
   - easiest way but not versatile enough as there is only one slot on machine
   - in this case slot for fluid filter will only provide whitelist/blacklist features
2. Pump will get slot for "fluid voiding item" (similar implementation as filter slot)
   - this will extend use of pumps and bring more possibilities of automation
   - working with fluid filter in tandem or as opposite to it would be option
   - option regarding if to void "all the time" or only if no fluid was moved in current cycle
3. New machine which will void any fluid coming in for little bit of energy
   - no instant voiding
   - speed of voiding and inventory size scalable by tier
   - recipe will use pump and "fluid voiding item"

I am proposing to go all the way and implement all three points. As they are connected and implementing one will save work on another. All of them should also provide full functionality regardless play style.

If this idea is interesting enough I will provide additional specifications and implementation notes.

PrincessOfEvil:
Fluid cells into a dropper pointing at lava. That being said, easier
voiding would definitely help with some designs.

Saereth:
This seems great, #2 would definitely be the most useful for me. Use case is generally something like, I overproduce oxygen or some other random byproduct fluid, I'd like to keep x amount for crafting operations and void the rest so the system doesnt get clogged, #2 seems perfect for that or alternatively #2 in conjunction with #3 either way. looks good

Shattered2909:
Maybe a setting for Quantum Tanks to void Fluid when they are full and you insert more?

ALongStringOfNumbers:
Both the first and second options could be solved by adding a fluid voiding mode to the Fluid Filter, which means that a new item does not have to be created. This voiding mode for the fluid filter could simply be a toggle on the existing fluid filter, meaning that anything specified by the filter, and when the filter is in whitelist mode, will be voided, while the fluids not specified by the filter, in whitelist mode, will not be voided.

Additionally, and empty filter on whitelist mode and with fluid voiding enabled would void all fluids, while an empty filter on blacklist mode, with voiding enabled, would not void anything.

Similarly, this fluid voiding mode on the fluid filter would be able to be activated while placed inside of a pump cover, which would cover the second option presented in the original issue.

[...]

PR #1529, opened 2021-03-14
Create the Fluid Voiding cover

ALongStringOfNumbers:
**What:**
As mentioned in #1141, there is a desire for a fluid voiding implementation, including a fluid voiding cover. This PR attempts to implement such a Fluid Voiding cover and by doing so addresses the first point brought up in #1141.

**How solved:**
The Fluid Voiding cover does not actively pull fluids into itself to void, but instead will void any fluids pushed through it or into it that match the filter.

To accomplish this: 
- The fluid transfer method was changed to allow for multiple target `IFluidHandler`s for the target inventories and `Predicate`s used for the fluid filter. This way, the fluid can go through the cover, be voided if it matches the filter, and into a target fluid handler inventory.
- The logic for `pushFluidsIntoNearbyHandlers` in `MetaTileEntity` was adjusted so that the Voiding Cover is a valid destination for the target push, even if there is no fluid handler in the blockspace that the fluid output is facing.
- A new Capability was introduced for blocks/items used for fluid voiding.

The model for this item was taken from one of the unused assets in the resources folder.

**Outcome:**
Introduce a Fluid Voiding Cover.

**Additional info:**
_This section is for screenshots which are required for any GUI, rendering or recipe changes. Or any additional info that reviewers should be aware of._

The changes to the fluid transfer should be fine, as the original method was redirected to the new method. The changes to MTE should also be fine; any addons that were overriding the push method though will not have compatibility for that machine for the new cover


Archengius:
I don't really see it's usability as a cover if it just voids everything. I would recommend reworking it to allow "void on overflow" kind of functionality additionally, which would make it much more useful.

ALongStringOfNumbers:
There are some recipes where there is an output that I would rather void every time the recipe runs, like recipes that have water, hydrogen, diluted x, etc, as an output in addition to another fluid. I would rather void these fluids and never have to move them into fluid storage.

An overflow cover would also have its usages, but could be implemented in addition to this cover to allow for voiding fluids once the internal tank limit is reached, instead of voiding fluids all the time.

src/main/java/gregtech/common/covers/CoverFluidVoiding.java
```diff
@@ -0,0 +1,148 @@
...
+        if (!coverHolder.getWorld().isRemote) {
+            openUI((EntityPlayerMP) playerIn);
+        }
+        return EnumActionResult.SUCCESS;
+    }
+
+    @Override
+    public boolean shouldCoverInteractWithOutputside() {
+        return true;
+    }
```
LAGIdiot, line 80: This should be false as it is only for covers which could input from output side and automatically allow that input. But name of function is not descriptive...

src/main/java/gregtech/api/capability/IFluidVoiding.java
```diff
@@ -0,0 +1,12 @@
+package gregtech.api.capability;
+
+import net.minecraftforge.fluids.FluidStack;
+
+import java.util.function.Predicate;
+
+public interface IFluidVoiding {
+
+    int voidingAmount();
```
LAGIdiot, line 9: This would be needed only if there were tiers.

LAGIdiot (commented):
I have partially reviewed this but then I got stuck. And some questions from given implementation arose. Most important is why  `IFluidVoiding` is not just extending `IFluidHandler`? In that case you would not need to change `GTFluidUtils` probably at all and it would simplify this quite a lot. Or is there something which prevents you to do that and I missed it?

Other comments bellow can be for now ignored.

ALongStringOfNumbers:
I will look into addressing the first comment about extending `IFluidHandler`, but the change to `GTFluidUtils` was done to allow transfer of fluid through multiple handlers easily. 
For example, having the fluid voiding filter on the output side of a chemical reactor, and then a quantum tank in the blockspace where the output side is pointing. With the change to `GTFluidUtils`, now the auto fluid export will attempt to pass fluids into the voiding cover, voiding whichever match the filter, and then into the quantum tank. Whereas previously, it would only attempt to export into one Fluid Handler.

LAGIdiot:
I definitely need to take another look to this as describe functionality regarding how void works seems strange to me.

ALongStringOfNumbers:
I need to review what I have done in this PR, as I am not quite sure it is implemented how it was expected to be, or as cleanly as possible.

PR #1531, opened 2021-03-16
Fix chests rendering invisibly in some cases

### LINKED ISSUE #853: Machines/Blocks going invisible
Hey! We're having an issue where any machine or block from GregTech will go invisible randomly. This is not easily replicatable and I'm wondering if anybody else has experienced this. Thanks!

**Modpack:** Omnifactory
**Mod Version:** 1.12.2-1.8.1.399

### LINKED ISSUE #1117: Invisible chests caused by Emerging Technology CO2 scrubbers
When placing an Emerging Technology CO2 scrubber near GTCE chests, the chests turn invisible. Looking away from the scrubber makes everything go visible again. See attached images for an illustration of the issue.

Relevant info:
GTCE version: 1.9.1.507
Emerging Technology version: 1.3.25
Forge version: 14.23.5.2847

### LINKED ISSUE #1235: [BUG] Thaumic Augmentation Warded Chest Causes GTCE Chests To Go Transparent, Mayhaps Other Mods As Well.
**Describe the bug**
Having Thaumic Augmentation included in the pack and at other times, some other mods (Unknown, can't seem to catch the others yet), causes texture bugs with some GTCE blocks, such as chests to go entirely transparent. Hitbox and all functionality remain, including HUD tools like Waila.

**Versions**
Forge: 14.23.5.2854
GTCE: 1.12.2-1.10.1.557
CodeChicken Lib 1.8.+: 1.12.2-3.2.3.358-universal
**Modpack:** None: Minimum to run GTCE and attempt to catch this bug.
**Addons:**
Baubles 1.12-1.5.2
JEI 1.12.2-4.16.1.301
Thaumcraft 1.12.2-6.1.BETA26
Thaumic Augmentation 1.12.2-1.1.19

**Setup**
Playing Solo Super Flat World. New World as a testing environment.

**Steps To Reproduce**
1) Place down one of each of GTCE chests.
2) Place down Thaumic Augmentation's "Warded Chest"
3) Done. Immediate results.

**Expected behavior**
Expect All blocks to still be fully visible.

**Screenshots**
1. Screenshot Immediately after the described steps (Shadows for chests remain)

2. Hitboxes and functionality remain.

**Additional context**
Removing Thaumic Augmentation Immediately resolves the issue. Placing it in GTCE bug tracker as this issue merely seems to be bugging out with other mods installed, not that GTCE is directly the cause. I have seen this happen with other blocks from other mods (placing lots of things, to come back and find this issue, even without placing a Warded Chest) So I assume, it may be on GTCE's end. It's sometimes hard to narrow things down with hundreds of mods. 

Edit: It seems that Thaumcraft's Hungry Chest may also be causing lighting glitches on CTCE chests as well, they all change to be about 15% brighter than they normally appear. Unfortunately Thaumcraft 6 on 1.12.2, is likely forever stuck in beta to my understanding with Azanor no longer actively developing it.

ALongStringOfNumbers:
**What:**
This PR fixes chests rendering invisible in some cases.

**How solved:**
In `MetaTileEntityRenderer#renderBlock`, all block rendering was getting called using `renderMetaTileEntity`. However, the chests override this method and have an empty method body, so nothing was getting rendered for the chests.

To fix this, I first check if the MTE is an instance of an `IFastRenderMetaTileEntity` and if it is I call the specific render method for those blocks, and then `renderMTE`

**Outcome:**
Fixes chests rendering invisibly in some cases. Closes #1235, Closes #1117, Closes #853 (Some parts)

**Additional info:**

https://user-images.githubusercontent.com/31759736/111386493-9128f280-8669-11eb-8cef-a67174dc00bd.mp4

With this most recent set of changes, there would be issues if someone was doing anything with the `hasFastRenderer` method

### DIFF
--- a/src/main/java/gregtech/api/render/MetaTileEntityTESR.java
+++ b/src/main/java/gregtech/api/render/MetaTileEntityTESR.java
@@ -46,36 +46,44 @@ public class MetaTileEntityTESR extends TileEntitySpecialRenderer<MetaTileEntity
                     ((IRenderMetaTileEntity) cover).renderMetaTileEntityDynamic(x, y, z, partialTicks);
                 }
             }
-            if (coverFast.size() > 0) {
-                Matrix4 translation = (new Matrix4()).translate(x, y, z);
-                Tessellator tessellator = Tessellator.getInstance();
-                BufferBuilder buffer = tessellator.getBuffer();
-                this.bindTexture(TextureMap.LOCATION_BLOCKS_TEXTURE);
-                RenderHelper.disableStandardItemLighting();
-                GlStateManager.blendFunc(GL11.GL_SRC_ALPHA, GL11.GL_ONE_MINUS_SRC_ALPHA);
-                GlStateManager.enableBlend();
-
-                if (Minecraft.isAmbientOcclusionEnabled()) {
-                    GlStateManager.shadeModel(GL11.GL_SMOOTH);
-                }
-                else {
-                    GlStateManager.shadeModel(GL11.GL_FLAT);
-                }
-                buffer.begin(GL11.GL_QUADS, DefaultVertexFormats.BLOCK);
-
-                for (Tuple<IFastRenderMetaTileEntity, EnumFacing> tuple : coverFast) {
-                    CCRenderState renderState = CCRenderState.instance();
-                    renderState.reset();
-                    renderState.bind(buffer);
-                    renderState.setBrightness(te.getWorld(), te.getPos().offset(tuple.getSecond()));
-                    tuple.getFirst().renderMetaTileEntityFast(renderState, translation, partialTicks);
-                }
-
-                buffer.setTranslation(0, 0, 0);
-                tessellator.draw();
-                RenderHelper.enableStandardItemLighting();
+            if (!coverFast.isEmpty()) {
+                renderCoverFast(te, x, y, x, partialTicks, destroyStage, alpha, coverFast);
             }
         }
+    }
+
+    /** Used Specifically to render covers that are IFastRenderMetaTileEntity
+     * Adapted from renderTileEntityFastPart
+     * **/
+    private void renderCoverFast(MetaTileEntityHolder te, double x, double y, double z, float partialTicks, int destroyStage, float alpha, List<Tuple<IFastRenderMetaTileEntity, EnumFacing>> coverList) {
+
+        Matrix4 translation = (new Matrix4()).translate(x, y, z);
+        Tessellator tessellator = Tessellator.getInstance();
+        BufferBuilder buffer = tessellator.getBuffer();
+        this.bindTexture(TextureMap.LOCATION_BLOCKS_TEXTURE);
+        RenderHelper.disableStandardItemLighting();
+        GlStateManager.blendFunc(GL11.GL_SRC_ALPHA, GL11.GL_ONE_MINUS_SRC_ALPHA);
+        GlStateManager.enableBlend();
+
+        if (Minecraft.isAmbientOcclusionEnabled()) {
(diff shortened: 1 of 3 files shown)


detav:
I think `MetaTileEntityRenderer#renderBlock` is just called on block updates, is not suitable for rendering a chest.

Problem is in the TESR, either CCL doesn't know how to render fastTESR or we are not using it right.

Easiest solution would be to make these entities not fast by just deleting [the override](https://github.com/GregTechCE/GregTech/blob/e812e4669feaf220d8eb7b5fbb6ea1ebe0f805bc/src/main/java/gregtech/api/metatileentity/MetaTileEntityHolder.java#L270), and then fixing [`MetaTileEntityTESR`](https://github.com/GregTechCE/GregTech/blob/e812e4669feaf220d8eb7b5fbb6ea1ebe0f805bc/src/main/java/gregtech/api/render/MetaTileEntityTESR.java#L26)  (`te` -> `te.getMetaTileEntity()`)

ALongStringOfNumbers:
Well, even if `MTERenderer#renderBlock` is only called on block updates, the chest still overrides the method it calls with an empty method, meaning that it will not get rendered, https://github.com/GregTechCE/GregTech/blob/e812e4669feaf220d8eb7b5fbb6ea1ebe0f805bc/src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityChest.java#L209-L211

meaning that on block updates, this empty method would still be called

detav:
It was left out on purpose, [the TESR](https://github.com/GregTechCE/GregTech/blob/e812e4669feaf220d8eb7b5fbb6ea1ebe0f805bc/src/main/java/gregtech/api/render/MetaTileEntityTESR.java#L70) is called every frame anyway.

ALongStringOfNumbers:
Tested with your suggested changes and it works as expected. I will be changing the PR to use those changes

ALongStringOfNumbers:
@detav The most recent set of changes was what you had in mind, correct?

ALongStringOfNumbers:
I have marked this for review, although if anyone has any other solutions or alternatives, please feel free to share.

ALongStringOfNumbers:
With the most recent commit, the ability for Covers to be `IFastRenderMTE`s was added. This was done for compatibility with addon mods, such as Gregicality, that were depending on the now removed `hasFastRenderer` override.

The logic for the cover rendering is adapted from `MetaTileEntityTESR#renderTileEntityFastPart` for use with rendering covers.

src/main/java/gregtech/api/render/MetaTileEntityTESR.java
```diff
@@ -14,22 +15,65 @@
...
+                    renderState.reset();
+                    renderState.bind(buffer);
+                    renderState.setBrightness(te.getWorld(), te.getPos().offset(tuple.getSecond()));
+                    tuple.getFirst().renderMetaTileEntityFast(renderState, translation, partialTicks);
+                }
+
+                buffer.setTranslation(0, 0, 0);
+                tessellator.draw();
+                RenderHelper.enableStandardItemLighting();
             }
```
LAGIdiot, line 77: Please extract this to two new methods. One for getting information if we need to render fast cover and second to do rendering.

LAGIdiot (changes requested):
Good job finding out where issue with rendering was and fixing it.

I have commented one place which needs adjustment for readability sake.

src/main/java/gregtech/api/render/MetaTileEntityTESR.java
ALongStringOfNumbers: I have now moved the section for cover rendering into its own method.

LAGIdiot (approved):
Thanks for finishing this up

PR #1537, opened 2021-03-20
Give Lossless Cables Loss as Wires (with Config)

serenibyss:
**What:**
This PR simply makes lossless cables now have loss as wires with a config value default to false. This incentivizes the player to cover their cables.

**How solved:**
I changed the calculation for cable loss to now check for if the cable is set to be lossless, and will now have the wires with loss at the given rates:
1x: 1V/block
2x: 1V/block
4x: 2V/block
8x: 2V/block
16x: 2V/block

**Outcome:**
- Lossless cables now have lossless wires if configured

**Additional info:**
This has been a concern of mine when creating lossless cables through CraftTweaker, as I wanted lossless cables, but also wanted them to still respect the loss multipliers specified in the `Insulation` enum.

None expected.

### DIFF
--- a/src/main/java/gregtech/common/ConfigHolder.java
+++ b/src/main/java/gregtech/common/ConfigHolder.java
@@ -105,6 +105,10 @@ public class ConfigHolder {
     @Config.Comment("If true, powered zero loss wires will damage the player. Default: false")
     public static boolean doLosslessWiresDamage = false;
 
+    @Config.Comment("If true, lossless cables will have lossy wires. Default: false")
+    @Config.RequiresMcRestart
+    public static boolean doLosslessWiresMakeLossyCables = false;
+
     public static class VanillaRecipes {
 
         @Config.Comment("Whether to nerf the paper crafting recipe. Default: true")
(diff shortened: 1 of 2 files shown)


Exaxxion:
While I am constantly annoyed by players finding every way possible to avoid covering their wires, I would urge caution about forcing lossy behaviors on them. If people want lossy wires, don't they already have the ability to add them that way? If packs want to add custom superconductive materials, I don't think it's a good idea to make it so they can't.

I'm old enough to remember playing during the 1.6.4 MJ/RF war and one of the major points of contention during it was that MJ started forcing machines to lose power during idle if modders set the property to zero; it would override with a default value basically saying "no, f*** you, it can't be lossless."

You probably know how all that turned out.

DStrand1:
I understand what you are saying, though I feel like there should at least be an option to generate lossless cables with lossy wires. Maybe even a general config option "Lossless cables have lossy wires" or something to get around this.

Exaxxion:
I think it's fine to give players a configurable option if they want to make use of it. I'm just recommending against forcing such a change.

DStrand1:
I updated this to use a config value and updated the Title and Opening Comment.

LAGIdiot:
I need to thing this PR through so I will not include it in this release.

LAGIdiot (approved):
Seems reasonable to me, after some though. Event though I would not use this myself there is place for it.

Issue #1532, opened 2021-03-19
[Suggestion] Pump (Machine) fluid filtering


Andrakax:
Currently, in order pump out oil that is touching water source blocks (most notably in the oceans), you must put in a wall between the water and oil, otherwise the pump will attempt to pull the water, which is infinite, and the water will cover the oil at some point making the oil un-pumpable.   Creating this wall is tedious, to say the least, in the ocean.

What I'd like to see is a way to put a filter in/on the pump to whitelist/blacklist fluids so that we can control what it pulls to mitigate some of this issue.

While this would help, I don't expect it to solve the issue of the oceanic oil being covered over by water once the oil is removed.  Maybe we can generate some more discussion on this.

PR #1546, opened 2021-03-23
Filter in pumps

galyfray:
**What:**
Added support for liquid filters in pumps (machines) 

**How solved:**
Added a fluid filter widget to the UI, check for liquid type when adding them to the queue and just before pumping the block (to avoid bug in case of block replacement or filter update). Invalidate the queue when switching filter from black/whitelist to white/blacklist and removing/adding the filter.

**Outcome:**
Implement #1532

**Additional info:**
Screenshots of UI : 
screenshot
screenshot2
None Expected.

### DIFF
--- a/src/main/java/gregtech/common/metatileentities/electric/MetaTileEntityPump.java
+++ b/src/main/java/gregtech/common/metatileentities/electric/MetaTileEntityPump.java
@@ -134,26 +134,26 @@ public class MetaTileEntityPump extends TieredMetaTileEntity implements IDirtyNo
 
     @Override
     protected ModularUI createUI(EntityPlayer entityPlayer) {
-        Builder builder = ModularUI.defaultBuilder();
+        Builder builder = ModularUI.builder(GuiTextures.BACKGROUND, 176, 147 + 82);
         builder.image(7, 16, 81, 55, GuiTextures.DISPLAY);
         TankWidget tankWidget = new TankWidget(exportFluids.getTankAt(0), 69, 52, 18, 18)
-            .setHideTooltip(true).setAlwaysShowFull(true);
+                .setHideTooltip(true).setAlwaysShowFull(true);
 
         builder.widget(tankWidget);
         builder.label(11, 20, "gregtech.gui.fluid_amount", 0xFFFFFF);
         builder.dynamicLabel(11, 30, tankWidget::getFormattedFluidAmount, 0xFFFFFF);
         builder.dynamicLabel(11, 40, tankWidget::getFluidLocalizedName, 0xFFFFFF);
 
-        this.fluidFilter.initUI(88, builder::widget);
+        this.fluidFilter.initUI(75, builder::widget);
 
         return builder.label(6, 6, getMetaFullName())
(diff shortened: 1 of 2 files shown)


LAGIdiot (changes requested):
Thanks for your hard work. There are few areas which needs adjustment and one big question regarding performance

src/main/java/gregtech/common/metatileentities/electric/MetaTileEntityPump.java
```diff
@@ -212,10 +221,10 @@ private void checkFluidBlockAt(BlockPos pumpHeadPos, BlockPos checkPos) {
...
 
         if (blockHere.getBlock() instanceof BlockLiquid ||
-            blockHere.getBlock() instanceof IFluidBlock) {
+                blockHere.getBlock() instanceof IFluidBlock) {
             IFluidHandler fluidHandler = FluidUtil.getFluidHandler(getWorld(), checkPos, null);
             FluidStack drainStack = fluidHandler.drain(Integer.MAX_VALUE, false);
-            if (drainStack != null && drainStack.amount > 0) {
+            if (drainStack != null && drainStack.amount > 0 && this.fluidFilter.testFluidStack(drainStack)) {
                 this.fluidSourceBlocks.add(checkPos);
             }
```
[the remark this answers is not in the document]
galyfray: To solve this the pump now has two dequeues : 
- `blockToCheck`  all neighbors that are further from the pump than the currently checked block are added to the dequeue
- `blockToReCheck` all neighbors closer to the pump than the checked block are added to this queue.
when checking block the first dequeue is emptied first using the older checking behavior, then the new dequeue get emptied but neighbors are added to the correct dequeue only if the current block is fully valid.

this way allow to infinitly pump fluids like water and ensure dequeues get empty over time on fluids like lava.
I didn't really find  a better way to achieve this but this way has the advantage of keeping older behavior as is in case some add-on makers rely on some of those functions.

ALongStringOfNumbers:
I have edited the opening comment to link the related issue, so that it will be closed when this is merged.

galyfray:
So what do you think of the current state of this PR @LAGIdiot  ? is there any other changes needed or anything else ?

src/main/java/gregtech/common/metatileentities/electric/MetaTileEntityPump.java
```diff
@@ -154,30 +171,40 @@ private int getMaxPumpRange() {
...
-        for(int i = 0; i < blocksToCheckAmount; i++) {
-            BlockPos checkPos = null;
-            int amountIterated = 0;
-            do {
-                if (checkPos != null) {
-                    blocksToCheck.push(checkPos);
-                    amountIterated++;
-                }
-                checkPos = blocksToCheck.poll();
+    private BlockPos getValidPos(Deque<BlockPos> posDeque) {
```
ALongStringOfNumbers, line 179: The passed argument is unused in this method. Any reason for not removing it?

galyfray: That's a copy paste fail the arg should get used

PR #1550, opened 2021-03-24
Add Zoom Functionality to Multiblock Previews

serenibyss:
**What:**
This PR allows for Multiblock Info Pages to override a new method `getZoom()`, which defaults to 1.0f if not overridden as it was before.

**How solved:**
I added a method to `MultiblockInfoPage` that can now be overridden by a specific multi info to set its zoom. It defaults to 1.0f if not overridden, as was the behavior before.

Example of the new zoom functionality:
Set to default (1.0f):
oldExample

Set to a new custom zoom (0.7f in this example):
newExample

It can be any float, recommended in the range (0.0f, 1.0f] though I think values above 1.0f would still work fine.

**Outcome:**
Allow Multiblock Previews to set a custom zoom

None expected.


warjort:
> I created a new interface named IMultiblockInfoPage so that I could have the getZoom() override not be required, as the old MultiblockInfoPage was an abstract class which lacks default methods.

Why not just?
````
public abstract class MultiblockInfoPage {

+public float getZoom() {
+      return +1.0f;
+}
````

DStrand1:
Yes that is a lot cleaner, I changed it to that

LAGIdiot (changes requested):
Good idea implementing this. Please just adjust naming there

DStrand1:
Name was updated to `getDefaultZoom()`

LAGIdiot (approved):
Thanks for finishing this up.

Issue #1374, opened 2020-12-27
[BUG] Infinite recursion when connecting GTCE cables to Compact Machine tunnels


md5i:
**Describe the bug**
As a test, I created a compact machine and placed it next to an active Gregtech Community Edition cable.  I added a tunnel to in the compact machine to the side with the cable.  When I placed a GTCE cable next to the tunnel, the game crashed.  The crash is due to infinite recursion in `CableEnergyContainer` between `dispatchEnergyToNode()` and `acceptEnergyFromNetwork()`.  The crash is persistent between reloads, so I have had to restore from backup.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.1.557
_Modpack:_ FTB Interactions 2.0.10 found (from FTB discord) at https://drive.google.com/file/d/1maGmng2twpUpZ-ZTBEBHPtEJPTmRN_d4/view
_Addons:_ none

**Setup**
Playing Solo

**Steps To Reproduce**
1) Place a Gregtech Advanced Diesel Generator down.
2) Fill the generator with fuel (Diesel).
3) Connect a copper cable to the generator's output.
4) Place the compact machine next to the cable.
5) Enter the machine and place a tunnel on the wall that connects to the external cable.
6) Place a copper cable next to the tunnel.
7) Crash.

**Expected behavior**
Since GTCE seems to be using Forge capabilities, I was expecting the cable to transmit energy into the compact machine through the tunnel.  In the worst case, I expected the cables not to connect at all through the tunnel.

Relevant portion of crash log:
```
[19:58:35] [Server thread/ERROR] [minecraft/MinecraftServer]: Encountered an unexpected exception
net.minecraft.util.ReportedException: Ticking block entity
	at net.minecraft.server.MinecraftServer.updateTimeLightAndEntities(MinecraftServer.java:773) ~[MinecraftServer.class:?]
	at net.minecraft.server.MinecraftServer.tick(MinecraftServer.java:668) ~[MinecraftServer.class:?]
	at net.minecraft.server.integrated.IntegratedServer.tick(IntegratedServer.java:185) ~[chd.class:?]
	at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:526) [MinecraftServer.class:?]
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_275]
Caused by: java.lang.StackOverflowError
	at net.minecraft.world.World.getChunk(World.java:305) ~[amu.class:?]
[...]

PR #1552, opened 2021-03-26
Fix NPE when something proxies a fluid pipe, e.g. a compact machine

### LINKED ISSUE #1551: Games crashes when placing bronze tube inside compact machine
https://github.com/GregTechCE/GregTech/blob/v1.10.0/src/main/java/gregtech/common/pipelike/fluidpipe/BlockFluidPipe.java#L134

Time: 2021-03-26 00:56:40 GMT
Description: Exception while ticking a block

```
java.lang.NullPointerException
    at gregtech.common.pipelike.fluidpipe.BlockFluidPipe.canPushIntoFluidHandler(BlockFluidPipe.java:134)
    at gregtech.common.pipelike.fluidpipe.BlockFluidPipe.getActiveNodeConnections(BlockFluidPipe.java:121)
    at gregtech.api.pipenet.block.BlockPipe.updateTick(BlockPipe.java:110)
    at net.minecraft.world.WorldServer.tickUpdates(WorldServer.java:737)
    at net.minecraft.world.WorldServer.tick(WorldServer.java:223)
    at net.minecraft.server.MinecraftServer.updateTimeLightAndEntities(MinecraftServer.java:756)
    at net.minecraft.server.MinecraftServer.tick(MinecraftServer.java:668)
    at net.minecraft.server.integrated.IntegratedServer.tick(IntegratedServer.java:185)
    at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:526)
    at java.lang.Thread.run(Thread.java:745)
```

Mod pack: FTB Interactions 2.0.9

The crash occured after I put the medium bronze tube on the wall of compact machine (inside), which was connected to another bronze tube with steam on outside (see https://www.curseforge.com/minecraft/mc-mods/compact-machines,  installed version compactmachines3-1.12.2-3.0.18-b278.jar)

Looks like Compact Machine mode creates virtual connections between pipes, but BlockFluidPipe tries to find TileEntity from inside of compact machine and fails to do so.

warjort:
**What:**
There is an NPE when something proxies capabilities of a fluid pipe, e.g. a compact machine.
This is in the code that checks the size of the pipes to see if it can push fluid. 
The size is not obtained through the capability and is obviously unavailable from the proxy tile entity.

**How solved:**
When something proxies the fluid pipe capability, simply disallow pushing  fluids across the proxy from one pipe to another.
In the case of the compact machine tunnel, allowing it would likely lead to recursive death as fluid is pushed back and forth both ways.

**Outcome:**
Fixes #1551

**Additional info:**
I checked the other places that call getPipeTileEntity() and all the others have null checks, except for those that are doing self references which should never return null. e.g. getting the "insulation" of a cable when a mob walks into it to see if damage should be inflicted.

None. This never could have worked before. So disabling the push instead of crashing is the better solution.

### DIFF
--- a/src/main/java/gregtech/common/pipelike/fluidpipe/BlockFluidPipe.java
+++ b/src/main/java/gregtech/common/pipelike/fluidpipe/BlockFluidPipe.java
@@ -131,7 +131,11 @@ public class BlockFluidPipe extends BlockMaterialPipe<FluidPipeType, FluidPipePr
         boolean isDestPipe = destinationHandler instanceof FluidPipeFluidHandler;
         if(isSourcePipe && isDestPipe) {
             float sourceThickness = selfTileEntity.getPipeType().getThickness();
-            float destThickness = getPipeTileEntity(otherTileEntity).getPipeType().getThickness();
+            IPipeTile<FluidPipeType, FluidPipeProperties> otherPipe = getPipeTileEntity(otherTileEntity);
+            if (otherPipe == null) {
+                return false;
+            }
+            float destThickness = otherPipe.getPipeType().getThickness();
             return sourceThickness > destThickness;
         }
         return true;


ALongStringOfNumbers:
Would this recursion you are talking about be similar to what is happening in #1374? In that case, there is a bout of infinite recursion between the energy cables connected via the Compact Machine's proxy handler block.

warjort:
I referenced that issue in the original bug report.

It should already be impossible for energy to be proxied via a compact machine tunnel using cables on both sides as discussed in #1374. A cable should not expose its capability to the tunnel unless the tunnel is already "recipricating".
The bug in #1374 is that this is happening but we can't reproduce the circumstance under which it occurs.

The fluid handler capability of the fluid pipes is different. In that fluid pipe networks can see each other as "tanks". 
There is no equivalent of this for different energy cable networks as the cable networks don't store energy internally.

LAGIdiot (approved):
Good work fixing this issue

PR #1553, opened 2021-03-27
Allow stone variants to output their respective stone dusts

Chaoziken:
What:
The stone variants (granite/diorite/andesite) ores (as well as bedrock ore) all have the same ore dictionary entry (ore<Material>) and thus all output stone dust when macerated, even though their respective stone dusts already exist.

How solved:
Changed the OrePrefix to oreGranite, oreDiorite, and oreAndesite and added their respective stone dusts as secondary materials. Since bedrock ore also has the "ore" OrePrefix, I changed it to oreBedrock but kept the regular stone dust as the secondary material since there is no bedrock dust.

Outcome:
Granite/Diorite/Andesite ores now output Granite/Diorite/Andesite dust, respectively, in the macerator.

Additional info:
Someone else is going to have to update the other language files with the new prefixes.
I have also changed the black/red granite prefixes' names to better differentiate them from regular granite.

Possible compatibility issue:
- Granite/Diorite/Andesite/Bedrock ores can no longer be processed by anything that only accepts ore<Material> ore dictionary entries. Previously, things such as Ender IO's SAG mill could process some granite/diorite/andesite/bedrock ores. If desired, this could be solved by changing the normal stone's OrePrefix to something like oreStone, and then adding a second "ore<Material>" ore dictionary entry to all of them in a similar way to how the dustRegular<Material> entry is handled.
- Any CraftTweaker script that relied on these ores being in the same oreDictEntry will no longer work as intended.

### DIFF
--- a/src/main/java/gregtech/api/unification/ore/OrePrefix.java
+++ b/src/main/java/gregtech/api/unification/ore/OrePrefix.java
@@ -29,6 +29,11 @@ public enum OrePrefix {
 
     ore("Ores", -1, null, MaterialIconType.ore, ENABLE_UNIFICATION | DISALLOW_RECYCLING, (mat) -> mat instanceof DustMaterial && mat.hasFlag(GENERATE_ORE)), // Regular Ore Prefix. Ore -> Material is a Oneway Operation! Introduced by Eloraam
 
+    oreGranite("Granite Ores", -1, null, MaterialIconType.ore, ENABLE_UNIFICATION | DISALLOW_RECYCLING, (mat) -> mat instanceof DustMaterial && mat.hasFlag(GENERATE_ORE)), // Granite Ore Prefix. Ore -> Material is a Oneway Operation!
+    oreDiorite("Diorite Ores", -1, null, MaterialIconType.ore, ENABLE_UNIFICATION | DISALLOW_RECYCLING, (mat) -> mat instanceof DustMaterial && mat.hasFlag(GENERATE_ORE)), // Diorite Ore Prefix. Ore -> Material is a Oneway Operation!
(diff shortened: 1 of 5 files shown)


warjort:
I don't think you want to break these ore dictionaries entries to achieve this.

Ore dictionaries are how mods communicate with each other to say they have the same item.
This is why your sag mill recipes no longer work.

Or look at how your change would affect this processing in an addon mod:
https://github.com/Gregicality/gregicality/blob/master/src/main/java/gregicadditions/recipes/OreRecipeHandler.java

There a probably many other similar things, as you say, some will be in people's CT scripts.

At the very least, a breaking change like this should be a config option (disabled by default).

But I would suggest you look at a way to keep the old oreXXX entries while still achieving what you want. An item can have multiple ore dictionary entries, like the dustRegular you mentioned.

Chaoziken:
Looking at things further, I do not know if there is a clean and easy way to have both the old ore<Material> entry and the new oreGranite<Material>, etc. entries on the same item.

The problem is the macerator auto-generates recipes that accept any ore<Material>, so having both ore dictionary entries would generate conflicting macerator recipes. Changing the recipe so that it accepts a different entry, say oreStone<Material> instead of ore<Material>, would mean that compatibility would break with every mod that adds ore<Material> entries, not to mention addon processing that relies on using OrePreifx.ore. I don't think there is a clean and easy way around that.

I think having a disabled-by-default config option for granite/diorite/andesite/bedrock ores to have their own unique OrePrefix is the only reasonable way to go forward, assuming it is possible and assuming it is not okay to just go forward as is.

LAGIdiot:
Thanks for your hard work.

I have already once looked at this problem and I come to conclusion that we don't have any way how to implement this without severally breaking compatibility. As was already mentioned we and our addons are heavily depending on these prefixes and we are generating tons of stuff base on them. So changing something to new one is not possible.

Also we don't gate breaking changes by config option.

In the end I won't accept this PR.

PR #1554, opened 2021-03-30
Better idling for metatileentities and multiblocks

PrototypeTrousers:
**What:**
This PR is my attempt on the issue of gregtech machines being "laggy" when actually idling with full inputs and outputs.
I've iterated this concept a some times and i think its good so here it is.

**How solved:**
implemented checks in the Abstract recipe logic for inputs (if a valid recipe can be made out of them) and outputs (if its full, and we cant accept a recipe because of it, wait for the output to make some space before looping unsuccessfully)
implemented fluid and item handlers that can change those settings by notifying the metatileentity that its inventory changed.

**Outcome:**
GTCE single and multiblock machines will idle with a input of partial ingredients and when its outputs are full.

**Additional info:**
By avoiding recipe lookups when the inputs dont make a valid recipe, and by avoiding simulating merging items that we cant push into the output, we can save some tick time for other stuff

Included tests that match the expected behaviour of a chemical reactor ( made by @Exa ) and a blast furnace.
the outputfull is set on the `addItemsToItemHandler` method, and checked for in the new method `shouldSearchForRecipes`, so will work immediatelly.
the inputsinvalidation flag should be set o the method that actually returns the Recipe object, and may need implementation on their part, but should not be too much work.


src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityFluidHatch.java
```diff
@@ -101,13 +102,21 @@ private int getInventorySize() {
...
         this.fluidTankHandler = new FluidTankList(false, fluidTank);
         this.exportFluids = fluidTankHandler;
         this.importFluids = isExportHatch ? new FluidTankList(false) : fluidTankHandler;
         this.fluidInventory = fluidTankHandler;
     }
 
+    @Override
+    public void addEntityToDirtyableHandlers(MetaTileEntity metaTileEntity) {
+        NotifiableFluidTank handler;
+        handler = (NotifiableFluidTank) getExportFluids().getFluidTanks().get(0);
```
LAGIdiot, line 115: There should be if with `instanceof NotifiableFluidTank ` as it can be inherited and `initializeInventory` overrided

LAGIdiot (changes requested):
Thanks for your hard work. From functional point it looks solid. I have some questions regarding implementation and other comments to places which needs adjustments.

Also there is problem with code formatting. As we are providing formatting in `.editorconfig` I would suggest you to apply it or I may ignore additional PRs without proper formatting (and yes I am that type of person).

src/main/java/gregtech/api/metatileentity/MetaTileEntity.java
```diff
@@ -89,7 +88,9 @@
     private int cachedLightValue;
     protected boolean isFragile = false;
 
-    private final CoverBehavior[] coverBehaviors = new CoverBehavior[6];
+    private CoverBehavior[] coverBehaviors = new CoverBehavior[6];
```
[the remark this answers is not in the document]
PrototypeTrousers: Im gonna blame a bad merge as i didnt intend to touch that at all

src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
```diff
@@ -143,22 +163,27 @@ protected void trySearchNewRecipe() {
...
+            currentRecipe = this.previousRecipe;
+        // If there is no active recipe, then we need to find one.
+        else {
+            metaTileEntity.setInputsDirty(false);
+            currentRecipe = findRecipe(maxVoltage, importInventory, importFluids);
         }
+        // If a recipe was found, then inputs were valid.
+        // recipe multiplying machines may not be able to fit
+        // the multiplied recipes. in that case the inputs are valid
+        if (!(this.invalidInputsForRecipes = (currentRecipe == null && !this.isOutputsFull)))
```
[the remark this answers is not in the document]
LAGIdiot: Thanks for explanation and rewriting that code

src/main/java/gregtech/api/capability/INotifiableHandler.java
```diff
@@ -0,0 +1,46 @@
...
+import gregtech.api.metatileentity.MetaTileEntity;
+
+import java.util.HashSet;
+import java.util.Iterator;
+
+/**
+ * For Item and Fluid handlers capable of notifying entities when
+ * their contents change
+ */
+public interface INotifiableHandler {
```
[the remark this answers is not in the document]
LAGIdiot: On one hand I don't want to restrict someone's usage of this by constraining it only to one MetaTileEntity to notify but on other hand it does not makes sense to me to hold HashSet in current implementation. Also there is problem that if someone incorrectly overrides `getNotifiableMetaTileEntities` as it generates new hashset every time as I suggested in comment bellow it will prevent correct function of `addNotifiableMetaTileEntity`... This needs to be addressed too

src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
```diff
@@ -106,12 +108,14 @@ public int getNetworkID() {
 
     @Override
     public void update() {
-        if (!getMetaTileEntity().getWorld().isRemote) {
+        World world = getMetaTileEntity().getWorld();
+        if (world == null || !world.isRemote) {
```
ALongStringOfNumbers, line 112: Should this be checking for a not null world?

src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
```diff
@@ -149,22 +176,30 @@ protected void trySearchNewRecipe() {
...
-                    this.previousRecipe = currentRecipe;
-                }
-            }
+
+        // see if the last recipe we used still works
+        if (this.previousRecipe != null && this.previousRecipe.matches(false, importInventory, importFluids))
+            currentRecipe = this.previousRecipe;
+        // If there is no active recipe, then we need to find one.
+        else {
+            metaTileEntity.setInputsDirty(false);
```
ALongStringOfNumbers, line 185: Is setting the inputs as not dirty here correct? We will be finding a new recipe, thus causing the inputs to change, meaning that unless the output is full, or the recipe lookup fails, the inputs will always be dirty.

src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
```diff
@@ -149,22 +176,30 @@ protected void trySearchNewRecipe() {
...
+        // they may return a null recipe. Since we only check for items and fluid here, having
+        // findRecipe return a null recipe with isOutputsFull being true, means we have a valid
+        // recipe in the input waiting for space in the output.
+        } else this.invalidInputsForRecipes = !this.isOutputsFull;
+
+        // proceed if we have a usable recipe.
         if (currentRecipe != null && setupAndConsumeRecipeInputs(currentRecipe)) {
             setupRecipe(currentRecipe);
+            //avoid new recipe lookup caused by item consumption from input
+            metaTileEntity.setInputsDirty(false);
```
ALongStringOfNumbers, line 202: Could this be combined with the previous `currentRecipe != null check`

src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
```diff
@@ -54,6 +54,8 @@
     protected boolean workingEnabled = true;
     protected boolean hasNotEnoughEnergy;
     protected boolean wasActiveAndNeedsUpdate;
+    protected boolean isOutputsFull;
```
ALongStringOfNumbers, line 57: Would there ever be a case where we want to differentiate between if the Fluid Outputs are full and if the item outputs are full? Maybe for something like the Electrolyzer?

src/main/java/gregtech/api/metatileentity/MetaTileEntity.java
```diff
@@ -973,8 +990,14 @@ public static boolean addItemsToItemHandler(final IItemHandler handler,
...
-        final boolean canMerge = simulateItemStackMerge(items, handler);
-
+        boolean canMerge;
+        if (items.size() == 1) {
+            canMerge = ItemHandlerHelper.insertItemStacked(handler, items.get(0), true).isEmpty();
+        } else {
+            // merging 2 or more stacks may change the slots available for merging,
+            // needing full simulation of the state of the inventory after every merge
+            canMerge = InventoryUtils.simulateItemStackMerge(items, handler);
+        }
```
ALongStringOfNumbers, line 1000: Not quite sure about the purpose of this change. Merging 1 stack of items can change the available slots for merging, if the input stack merges into a partial stack, filling it up, and then creating a new partial stack that has to be placed in a new slot.

src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java
```diff
@@ -98,27 +98,35 @@ protected void trySearchNewRecipe() {
...
                 currentRecipe = findRecipe(maxVoltage, importInventory, importFluids);
-                if (currentRecipe != null) {
-                    this.previousRecipe = currentRecipe;
-                }
             } else if (previousRecipe != null && previousRecipe.matches(false, importInventory, importFluids)) {
                 //if previous recipe still matches inputs, try to use it
                 currentRecipe = previousRecipe;
+            } else {
+                metaTileEntity.setInputsDirty(false);
+                currentRecipe = findRecipe(maxVoltage, importInventory, importFluids);
```
ALongStringOfNumbers, line 115: This logic is essentially duplicated on either side of the check of the previous recipe. It should most likely be on one side or another.

ALongStringOfNumbers (changes requested):
Spent a brief time looking this over, and left a small number of comments.

In addition to the comments left on the PR, I noticed some weird behavior while testing in game.

When testing on multiblocks, the front face was not lighting up green as should when the multiblock becomes active. This occurred for both the Multismelter and the EBF, so I expect somewhere that `isActive` is not being properly updated.

src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
PrototypeTrousers: the same effect would be achieved by testing for ex.: ( isItemOutpuFull || isFluidOutputFull ) for recipe purposes.
setupAndConsumeRecipeInputs requires both to merge successfully before returning true

src/main/java/gregtech/api/metatileentity/MetaTileEntity.java
PrototypeTrousers: if we only have 1 stack to merge, the simulated insertion will mirror the actual result of the real insertion.
with 2 or more, the insertion of stack 1 can change the space available for stack 2.

src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java
PrototypeTrousers: solved by initializing inputsDirty to true, otherwise multiblock would get stuck if saved with output full until you changed something in the inputs

src/main/java/gregtech/api/metatileentity/MetaTileEntity.java
PrototypeTrousers: you're right

LAGIdiot:
Thanks for updating this. Can you please fix formatting of all edited files as it does not conform our style. IntelliJ IDEA should be able to do it perfectly as we are providing `.editorconfig` file.

LAGIdiot:
This looks quite solid to me. Anyone from Addons do you see any glaring problems that this could cause to you?
Also this is candidate for early sprint merging as it will need lot of testing and would be great to get it finished soon because with every release there are conflicts which needs to be fixed which I believe is quite excruciating.

src/main/java/gregtech/common/metatileentities/electric/MetaTileEntityMacerator.java
```diff
@@ -41,11 +41,11 @@ protected int getMachineTierForRecipe(Recipe recipe) {
 
     @Override
     protected IItemHandlerModifiable createExportItemHandler() {
-        return new ItemStackHandler(outputAmount);
+        return new NotifiableItemStackHandler(outputAmount,this,true);
```
ALongStringOfNumbers, line 44: need some spaces between arguments

src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java
```diff
@@ -198,9 +205,12 @@ protected Recipe findRecipe(long maxVoltage,
             }
 
             // If there were no accepted ingredients, then there is no recipe to process.
-            if(recipeInputs.isEmpty()) {
+            // the output may be filled up
+            if (recipeInputs.isEmpty() && !invalidInputsForRecipes) {
```
ALongStringOfNumbers, line 209: The only case where `!invalidInputsForRecipes` would be true is when a matching recipe exists, and so the recipe inputs would not be empty, so this logic doesn't quite make sense.

ALongStringOfNumbers (changes requested):
Some more things brought up in looking this over.

ALongStringOfNumbers:
For addons, this might cause some issues with @DStrand1 's Distinct Input bus implementation found in Gregicality (Initial PR done here <https://github.com/Gregicality/gregicality/pull/451>), although I will leave it up to him to review this and see if he can rework the distinct bus implementation based on the changes in this PR, and if he has any other concerns.

It may be interesting to implement distinct buses directly in GTCE, as although GTCE multiblocks could not make much use of it, especially since there are no Large Machine Multiblocks as there are in Gregicality, having a unified implementation may make it easier for addons to implement it, and integrate it with the changes done to GTCE logic.

src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java
PrototypeTrousers: if the output is full ( so canFit is false ) for that recipe, we exit the loop before adding to recipeInputs.
in that case, an empty recipeInputs does not mean that there are no valid input, it means we could not add anything because the output is full.

ALongStringOfNumbers: However, we could have iterated through the for loop once first, which would set the state of `invalidInputsForRecipes`. It might be a good idea to reset the value of `invalidInputsForRecipes` if the check for `canFitOutputs` passes, but before the break out of the loop.

warjort:
Sorry to be late to the party on this.

I have a few comments, the first is a longish diff.
It basically makes 2 changes.

* It replaces the lists of modified handlers with booleans in MetaTileEntity
The only thing that was being done with these lists was x.add(), x.clear() and x.size() > 0 which is isomorphic to x = true, x = false, x == true

* It makes the passing of the notification handler around more implicit.
The handler is only ever "this" currently. Is there a use case where it would be something else?
I also think isExport could be implicit as well if isExport() was part of the INotifiableHandler interface and implemented on the 3 handlers from their internal fields. I didn't change this though.
Is there any use case where isExportHatch on the multipart bus/hatch doesn't match isExport on the handler?

Here is the diff:
````
diff --git a/src/main/java/gregtech/api/capability/INotifiableHandler.java b/src/main/java/gregtech/api/capability/INotifiableHandler.java
index 34cf1c36..f6badc9b 100644
--- a/src/main/java/gregtech/api/capability/INotifiableHandler.java
+++ b/src/main/java/gregtech/api/capability/INotifiableHandler.java
@@ -14,12 +14,12 @@ public interface INotifiableHandler {
      * @param isExport boolean specifying if a handler is an output handler
      */
 
-    default <T> void addToNotifiedList(MetaTileEntity metaTileEntity, T handler, boolean isExport) {
+    default void addToNotifiedList(MetaTileEntity metaTileEntity, boolean isExport) {
         if (metaTileEntity != null && metaTileEntity.isValid()) {
             if (isExport) {
-                metaTileEntity.addNotifiedOutput(handler);
+                metaTileEntity.addNotifiedOutput(this);
             } else {
-                metaTileEntity.addNotifiedInput(handler);
+                metaTileEntity.addNotifiedInput(this);
             }
         }
     }
diff --git a/src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java b/src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
index c8504522..1048b34b 100755
--- a/src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/AbstractRecipeLogic.java
@@ -131,13 +131,13 @@ public abstract class AbstractRecipeLogic extends MTETrait implements IWorkable
     }
 
     protected boolean hasNotifiedInputs() {
-        return (metaTileEntity.getNotifiedItemInputList().size() > 0 ||
-                metaTileEntity.getNotifiedFluidInputList().size() > 0);
+        return (metaTileEntity.isNotifiedItemInput() ||
+                metaTileEntity.isNotifiedFluidInput());
     }
 
     protected boolean hasNotifiedOutputs() {
-        return (metaTileEntity.getNotifiedItemOutputList().size() > 0 ||
-                metaTileEntity.getNotifiedFluidOutputList().size() > 0);
+        return (metaTileEntity.isNotifiedItemOutput() ||
+                metaTileEntity.isNotifiedFluidOutput());
     }
 
     protected boolean canFitNewOutputs() {
@@ -145,8 +145,8 @@ public abstract class AbstractRecipeLogic extends MTETrait implements IWorkable
         if (this.isOutputsFull && !hasNotifiedOutputs()) return false;
         else {
             this.isOutputsFull = false;
-            metaTileEntity.getNotifiedItemOutputList().clear();
-            metaTileEntity.getNotifiedFluidOutputList().clear();
+            metaTileEntity.resetNotifiedItemOutput();
+            metaTileEntity.resetNotifiedFluidOutput();
         }
         return true;
     }
@@ -205,8 +205,8 @@ public abstract class AbstractRecipeLogic extends MTETrait implements IWorkable
         if (currentRecipe != null && setupAndConsumeRecipeInputs(currentRecipe))
             setupRecipe(currentRecipe);
         // Inputs have been inspected.
-        metaTileEntity.getNotifiedItemInputList().clear();
-        metaTileEntity.getNotifiedFluidInputList().clear();
+        metaTileEntity.resetNotifiedItemInput();
+        metaTileEntity.resetNotifiedFluidInput();
     }
 
     public void forceRecipeRecheck() {
@@ -235,8 +235,8 @@ public abstract class AbstractRecipeLogic extends MTETrait implements IWorkable
     @Deprecated
     protected boolean checkRecipeInputsDirty(IItemHandler inputs, IMultipleTankHandler fluidInputs) {
         boolean isDirty = this.hasNotifiedInputs();
-        metaTileEntity.getNotifiedItemInputList().clear();
-        metaTileEntity.getNotifiedFluidInputList().clear();
+        metaTileEntity.resetNotifiedItemInput();
+        metaTileEntity.resetNotifiedFluidInput();
         return isDirty;
     }
 
diff --git a/src/main/java/gregtech/api/capability/impl/NotifiableFilteredFluidHandler.java b/src/main/java/gregtech/api/capability/impl/NotifiableFilteredFluidHandler.java
index 7e5d3ab4..db4ee451 100644
--- a/src/main/java/gregtech/api/capability/impl/NotifiableFilteredFluidHandler.java
+++ b/src/main/java/gregtech/api/capability/impl/NotifiableFilteredFluidHandler.java
@@ -17,7 +17,7 @@ public class NotifiableFilteredFluidHandler extends FilteredFluidHandler impleme
     @Override
     protected void onContentsChanged() {
         super.onContentsChanged();
-        addToNotifiedList(notifiableEntity, this, isExport);
+        addToNotifiedList(notifiableEntity, isExport);
     }
 
     @Override
diff --git a/src/main/java/gregtech/api/capability/impl/NotifiableFluidTank.java b/src/main/java/gregtech/api/capability/impl/NotifiableFluidTank.java
index 2c54e096..9aa141e4 100644
--- a/src/main/java/gregtech/api/capability/impl/NotifiableFluidTank.java
+++ b/src/main/java/gregtech/api/capability/impl/NotifiableFluidTank.java
@@ -18,7 +18,7 @@ public class NotifiableFluidTank extends FluidTank implements INotifiableHandler
     @Override
     protected void onContentsChanged() {
         super.onContentsChanged();
-        addToNotifiedList(notifiableEntity, this, isExport);
+        addToNotifiedList(notifiableEntity, isExport);
     }
 
     @Override
diff --git a/src/main/java/gregtech/api/capability/impl/NotifiableItemStackHandler.java b/src/main/java/gregtech/api/capability/impl/NotifiableItemStackHandler.java
index 88e3c58e..4858287e 100644
--- a/src/main/java/gregtech/api/capability/impl/NotifiableItemStackHandler.java
+++ b/src/main/java/gregtech/api/capability/impl/NotifiableItemStackHandler.java
@@ -17,9 +17,9 @@ public class NotifiableItemStackHandler extends ItemStackHandler implements IIte
     }
 
     @Override
-    public void onContentsChanged(int slot) {
+    protected void onContentsChanged(int slot) {
         super.onContentsChanged(slot);
-        addToNotifiedList(notifiableEntity, this, isExport);
+        addToNotifiedList(notifiableEntity, isExport);
     }
 
     @Override
diff --git a/src/main/java/gregtech/api/metatileentity/MetaTileEntity.java b/src/main/java/gregtech/api/metatileentity/MetaTileEntity.java
index 2c457529..d94462d4 100644
--- a/src/main/java/gregtech/api/metatileentity/MetaTileEntity.java
+++ b/src/main/java/gregtech/api/metatileentity/MetaTileEntity.java
@@ -88,10 +88,10 @@ public abstract class MetaTileEntity implements ICoverable {
     protected boolean isFragile = false;
 
     private final CoverBehavior[] coverBehaviors = new CoverBehavior[6];
-    protected List<IItemHandlerModifiable> notifiedItemOutputList = new ArrayList<>();
-    protected List<IItemHandlerModifiable> notifiedItemInputList = new ArrayList<>();
-    protected List<IFluidHandler> notifiedFluidInputList = new ArrayList<>();
-    protected List<IFluidHandler> notifiedFluidOutputList = new ArrayList<>();
+    protected boolean notifiedItemOutput = false;
+    protected boolean notifiedItemInput = false;
+    protected boolean notifiedFluidInput = false;
+    protected boolean notifiedFluidOutput = false;
 
     public MetaTileEntity(ResourceLocation metaTileEntityId) {
         this.metaTileEntityId = metaTileEntityId;
@@ -261,25 +261,17 @@ public abstract class MetaTileEntity implements ICoverable {
 
     public <T> void addNotifiedInput(T input) {
         if (input instanceof IItemHandlerModifiable) {
-            if (!notifiedItemInputList.contains(input)) {
-                this.notifiedItemInputList.add((IItemHandlerModifiable) input);
-            }
+            this.notifiedItemInput = true;
         } else if (input instanceof FluidTank) {
-            if (!notifiedFluidInputList.contains(input)) {
-                this.notifiedFluidInputList.add((FluidTank) input);
-            }
+            this.notifiedFluidInput = true;
         }
     }
 
     public <T> void addNotifiedOutput(T output) {
         if (output instanceof IItemHandlerModifiable) {
-            if (!notifiedItemOutputList.contains(output)) {
-                this.notifiedItemOutputList.add((IItemHandlerModifiable) output);
-            }
+            this.notifiedItemOutput = true;
         } else if (output instanceof NotifiableFluidTank) {
-            if (!notifiedFluidOutputList.contains(output)) {
-                this.notifiedFluidOutputList.add((NotifiableFluidTank) output);
-            }
+            this.notifiedFluidOutput = true;
         }
     }
 
@@ -1239,20 +1231,36 @@ public abstract class MetaTileEntity implements ICoverable {
         return exportFluids;
     }
 
-    public List<IItemHandlerModifiable> getNotifiedItemOutputList() {
-        return notifiedItemOutputList;
+    public boolean isNotifiedItemOutput() {
+        return this.notifiedItemOutput;
+    }
+
+    public boolean isNotifiedItemInput() {
+        return this.notifiedItemInput;
+    }
+
+    public boolean isNotifiedFluidInput() {
+        return this.notifiedFluidInput;
+    }
+
+    public boolean isNotifiedFluidOutput() {
+        return this.notifiedFluidOutput;
+    }
+
+    public void resetNotifiedItemOutput() {
+        this.notifiedItemOutput = false;
     }
 
-    public List<IItemHandlerModifiable> getNotifiedItemInputList() {
-        return notifiedItemInputList;
+    public void resetNotifiedItemInput() {
+        this.notifiedItemInput = false;
     }
 
-    public List<IFluidHandler> getNotifiedFluidInputList() {
-        return notifiedFluidInputList;
+    public void resetNotifiedFluidInput() {
+        this.notifiedFluidInput = false;
     }
 
-    public List<IFluidHandler> getNotifiedFluidOutputList() {
-        return notifiedFluidOutputList;
+    public void resetNotifiedFluidOutput() {
+        this.notifiedFluidOutput = false;
     }
 
     public boolean isFragile() {
diff --git a/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityFluidHatch.java b/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityFluidHatch.java
index 1664c39e..526ce7aa 100644
--- a/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityFluidHatch.java
+++ b/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityFluidHatch.java
@@ -130,7 +130,7 @@ public class MetaTileEntityFluidHatch extends MetaTileEntityMultiblockPart imple
         }
         if (handler != null) {
             handler.setNotifiableMetaTileEntity(metaTileEntity);
-            handler.addToNotifiedList(this, handler, isExportHatch);
+            handler.addToNotifiedList(this, this.isExportHatch);
         }
     }
 
diff --git a/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityItemBus.java b/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityItemBus.java
index 39783c00..fe52029d 100644
--- a/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityItemBus.java
+++ b/src/main/java/gregtech/common/metatileentities/electric/multiblockpart/MetaTileEntityItemBus.java
@@ -99,7 +99,7 @@ public class MetaTileEntityItemBus extends MetaTileEntityMultiblockPart implemen
         }
         if (handler != null) {
             handler.setNotifiableMetaTileEntity(metaTileEntity);
-            handler.addToNotifiedList(this, handler, isExportHatch);
+            handler.addToNotifiedList(this, this.isExportHatch);
         }
     }
 
diff --git a/src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java b/src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java
index 1651bc56..c6e18997 100644
--- a/src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java
+++ b/src/main/java/gregtech/common/metatileentities/multi/electric/MetaTileEntityMultiFurnace.java
@@ -126,7 +126,7 @@ public class MetaTileEntityMultiFurnace extends RecipeMapMultiblockController {
             if (currentRecipe != null && setupAndConsumeRecipeInputs(currentRecipe))
                 setupRecipe(currentRecipe);
             // Inputs have been inspected.
-            metaTileEntity.getNotifiedItemInputList().clear();
+            metaTileEntity.resetNotifiedItemInput();
         }
 
         @Override
````

File: src/main/java/gregtech/api/metatileentity/MetaTileEntity.java [hunk 13]
```diff
@@ -257,6 +259,30 @@ public final String getMetaFullName() {
         return getMetaName() + ".name";
     }
 
+    public <T> void addNotifiedInput(T input) {
```
warjort, line 262: I am not understanding the logic here?

Why do items use the top level class IItemHandlerModifiable
while the fluids use the one subclass FluidTank in one place and another subclass NotifiableFluidTank in another.

Why don't the fluids use the IFluidHandler top level class?

src/main/java/gregtech/api/metatileentity/MetaTileEntity.java
```diff
@@ -257,6 +259,30 @@ public final String getMetaFullName() {
...
+                this.notifiedItemInputList.add((IItemHandlerModifiable) input);
+            }
+        } else if (input instanceof FluidTank) {
+            if (!notifiedFluidInputList.contains(input)) {
+                this.notifiedFluidInputList.add((FluidTank) input);
+            }
+        }
+    }
+
+    public <T> void addNotifiedOutput(T output) {
```
warjort, line 274: The parameterized type <T> on this and the addNotifiedInput() does nothing?
It could just be replaced with Object.
You are using instanceof on the parameter anyway.

src/main/java/gregtech/api/capability/INotifiableHandler.java
```diff
@@ -0,0 +1,33 @@
...
+            } else {
+                metaTileEntity.addNotifiedInput(handler);
+            }
+        }
+    }
+
+    /**
+     * @param metaTileEntity MetaTileEntity to be notified
+     */
+    default void setNotifiableMetaTileEntity(MetaTileEntity metaTileEntity) {
```
warjort, line 30: Should this really be a default method?
Is there a case where an implementing class doesn't implement this method?

warjort:
Unfortunately, I don't know much about the AbstractRecipeLogic class which is the important change in this PR.
So I can't really offer any insights there.

PR #1555, opened 2021-03-31
Allow Quantum Chests/Tanks to Keep Inventory on Break

### LINKED ISSUE #1530: [Suggestion] Allow Quantum Containers to Keep Inventory on Break
Historically (and currently), both Quantum Tanks and Chests destroy their contents when broken. I believe the primary reason for this being the case in GT5 was because at the time, these storage options were *very* strong, having significantly more storage than other options available. However, in modern modded, this just does not seem to be the case anymore. Storage Drawers are cheaper than Quantum Chests, and can easily hold more than enough items in them. And similarly, ExtraCells 1k Fluid Drives can hold even more than the early tiers of Quantum Tanks.

I would like to open the discussion on changing this, and allow for these storage options to keep their inventory on break. As far as implementation goes, this would be trivial. Though, this is a relatively significant change, so I want to know some opinions before writing the change.

**Additional context**
A suggested idea to balance it somewhere in-between was to not allow Quantum Tanks and Chests to have any more items/fluids input into them after moved until they are fully emptied first. This way, the player can still move the container safely if they put it in a bad spot, but they cannot do it for free.

DStrand1:
This PR is simple in execution, but complex in design. Quantum Tanks/Chests now simply write their storage data to NBT when broken, so that they can keep their inventory and be moved. However, this is a big change from previous behavior, so I would like to discuss my reasonings for this change.

## Compared to Other Options
**Quantum Chest:**
|                       | **Total Size**                   | **Portability**                      | **Ease of use** | **Number of unique types**                    | **Cost to produce** | **Other notes**                                                                   |
|-----------------------|----------------------------------|--------------------------------------|-----------------|-----------------------------------------------|---------------------|-----------------------------------------------------------------------------------|
| **Quantum Tank**      | Extremely large                  | Cannot be moved or in-place upgraded | Not very easy   | Exactly 1                                     | Very high           | No in-place upgrade, no way to move at all, MUST be emptied or voids all on break |
| **Storage Drawer**    | 64 stacks, up to 30x that        | Can be moved with an item            | Very easy       | Only 1 type, up to 4 but cuts storage of each | Very low            | Can be easily upgraded in place, can be sorted with a Drawer Controller           |
| **AA Storage Crate**  | Much larger than standard chests | Can be moved with an item            | Very easy       | Any number of types                           | Very low            | Nothing of note                                                                   |
| **GT Chest/GA Crate** | Fairly large at highest tier     | Cannot be moved or in-place upgraded | Quite easy      | Any number of types                           | Low                 | Can support covers                                                                |
| **AE2 Cell**          | Extremely large                  | Very portable                        | Very easy       | Up to 63 different types                      | Fairly high         | Extremely compact                                                                 |
| **Iron Chests**       | Slightly larger than chests      | Not very portable                    | Very easy       | Any number of types                           | Very low            | Can be easily upgraded in place                                                   |

I think that this can be summarized quite easily. GT Quantum Chests are by far the least user-friendly option, being the only storage option to void all of its contents on break, where others simply drop them. It also has no way of moving it, and no real benefits to using it other than being very large. Additionally, it is arguably the most expensive option aside from maybe AE2 cells, but most of that cost is in infrastructure and not the cell itself. With this conclusion, I feel that this change is justified for the Quantum Chest.

**Quantum Tank:**
|                         | **Total Size**  | **Portability**                      | **Ease of Use** | **Number of Unique Types** | **Cost to Produce** | **Other Notes**                                                                               |
|-------------------------|-----------------|--------------------------------------|-----------------|----------------------------|---------------------|-----------------------------------------------------------------------------------------------|
| **Quantum Tank**        | Quite large     | Cannot be moved or in-place upgraded | Not very easy   | Exactly 1                  | Very high           | Cannot be moved, only tank in this list that voids on break                                   |
| **AE2 Fluid Cell**      | Extremely large | Very portable                        | Very easy       | Up to 5                    | Fairly high         | Holds massive amounts of fluid very conveniently, and multiple types                          |
| **GT Multiblock Tank**  | Fairly small    | Can be moved and keeps contents      | Fairly easy     | Exactly 1                  | Fairly low          | Requires many blocks to hold at its max capacity                                              |
| **ExtraUtilities Drum** | Quite large     | Can be moved and keeps contents      | Very easy       | Exactly 1                  | Fairly low          | Nothing of note                                                                               |
| **EnderIO Tank**        | Very small      | Can be moved and keeps contents      | Extremely easy  | Exactly 1                  | Very low            | Very configurable, can push/pull from adjacent inventories, can be crafted to clear inventory |
| **GA Drum**             | Very small      | Can be moved and keeps contents      | Extremely easy  | Exactly 1                  | Very low            | Can support covers, can be crafted to clear inventory                                         |

In this table, you can see that not only is the Quantum Tank similarly flawed like the Quantum Chest, it no longer has the benefit of being the largest storage option. AE2 Fluid Cells are significantly larger than quantum tanks, can hold multiple types, and are even cheaper at the lowest tier. I think our Quantum Tank is closest to the Extra Utilities Drum, which does hold its inventory on break. So with this, I feel this change is justified for the Quantum Tank.

## Results
Both the Quantum Tank and Chest now have a new tooltip to show what they are carrying inside

Additionally, similarly to the EnderIO tank, I added recipes to strip the NBT off of the Tank/Chest, to make it easier for players to use them in recipes, if that is needed.

None expected.

Closes #1530


src/main/java/gregtech/api/metatileentity/MetaTileEntity.java
```diff
@@ -1211,4 +1211,14 @@ public float getBlockHardness() {
     public float getBlockResistance() {
         return 6.0f;
     }
-}
\ No newline at end of file
+
+    /**
+     * Override this if the MTE will keep its Item inventory on-break
+     * If this is carelessly overridden and set to true, then Items will be voided.
```
LAGIdiot, line 1217: I think this should be more informative that by setting this to true author is responsible for handling inventory storing and restoring on break placement respectively

src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityQuantumChest.java
```diff
@@ -134,6 +135,25 @@ protected void addDisplayInformation(List<ITextComponent> textList) {
...
+
+        NBTTagCompound compound = stack.getTagCompound();
+        if (compound != null) {
+            String translationKey = null;
+            long count = 0;
+            if (compound.hasKey("ItemStack")) {
+                translationKey = new ItemStack(compound.getCompoundTag("ItemStack")).getUnlocalizedName() + ".name";
+                count = compound.getLong("ItemAmount");
+            } else if (compound.hasKey("PartialStack")) {
+                ItemStack tempStack = new ItemStack(compound.getCompoundTag("PartialStack"));
```
LAGIdiot, line 147: You should use constants instead of duplicating string literals

src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityQuantumChest.java
```diff
@@ -178,6 +198,37 @@ public void readFromNBT(NBTTagCompound data) {
...
+            if (!this.itemStack.isEmpty()) {
+                this.itemsStoredInside = itemStack.getLong("ItemAmount");
+            }
+        } else if (itemStack.hasKey("PartialStack", NBT.TAG_COMPOUND)) {
+            importItems.setStackInSlot(0, new ItemStack(itemStack.getCompoundTag("PartialStack")));
+        }
+    }
+
+    @Override
+    public void writeItemStackData(NBTTagCompound itemStack) {
```
LAGIdiot, line 215: There is problem with calculating number of items on stacks when full stack is not 64 items. As on pictures below I have put in 2 full stacks (but capacity is 16) and got back 5 full stacks.

src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityQuantumChest.java
```diff
@@ -134,6 +135,25 @@ protected void addDisplayInformation(List<ITextComponent> textList) {
...
+                count = compound.getLong("ItemAmount");
+            } else if (compound.hasKey("PartialStack")) {
+                ItemStack tempStack = new ItemStack(compound.getCompoundTag("PartialStack"));
+                translationKey = tempStack.getUnlocalizedName() + ".name";
+                count = tempStack.getCount();
+            }
+            if (translationKey != null) {
+                tooltip.add(I18n.format("gregtech.machine.quantum_chest.tooltip.item",
+                        I18n.format(translationKey)));
+                tooltip.add(I18n.format("gregtech.machine.quantum_chest.tooltip.count", count));
```
LAGIdiot, line 154: Tooltip regarding cleaning content by placing in crafting grid would be nice too.

src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityQuantumChest.java
```diff
@@ -134,6 +135,25 @@ protected void addDisplayInformation(List<ITextComponent> textList) {
     public void addInformation(ItemStack stack, @Nullable World player, List<String> tooltip, boolean advanced) {
         super.addInformation(stack, player, tooltip, advanced);
         tooltip.add(I18n.format("gregtech.machine.quantum_chest.capacity", maxStoredItems));
+
+        NBTTagCompound compound = stack.getTagCompound();
+        if (compound != null) {
+            String translationKey = null;
+            long count = 0;
+            if (compound.hasKey("ItemStack")) {
+                translationKey = new ItemStack(compound.getCompoundTag("ItemStack")).getUnlocalizedName() + ".name";
```
LAGIdiot, line 144: There is problem with showing names for some of our items as on picture below:

LAGIdiot (changes requested):
I like this idea of making our Quantum Chests/Tanks more user friendly and. And I must say that provided tables with comparisons were really cool.

I have commented on some glaring bugs and place which needs adjustments.

DStrand1:
> There is problem with calculating number of items on stacks when full stack is not 64 items. As on pictures below I have put in 2 full stacks (but capacity is 16) and got back 5 full stacks.

I addressed this by adding `this.itemStack.getMaxStackSize()` instead of just assuming 64.

> There is problem with showing names for some of our items as on picture below:

Changed to use `getDisplayName()` which works perfectly for our meta items.

> Tooltip regarding cleaning content by placing in crafting grid would be nice too.

I decided to do this just like how EnderIO displays clearing its tank's contents. Its implementation is also taken from EnderIO.
destroycontents

LAGIdiot (dismissed):
Thanks for fixing mentioned problems.

New implementation for tooltip is incorrect as it is show not only in expected cases:



So please take another look at it.

DStrand1:
I have addressed that issue, though I am quite unhappy with the required solution. Currently, I do not feel it is up to our standards, so if anyone can come up with something better, that would be great. While it does work (and even excludes the normal crafting recipe for Quantum Tank/Chest), I feel that it probably should be left out of this PR if the code cannot be cleaned up more.

src/main/java/gregtech/common/ClientProxy.java
```diff
@@ -182,6 +190,37 @@ public static void addMaterialFormulaHandler(ItemTooltipEvent event) {
...
+            if (player.openContainer instanceof ContainerWorkbench)
+                inv = ((ContainerWorkbench) player.openContainer).craftMatrix;
+            else if (player.openContainer instanceof ContainerPlayer)
+                inv = ((ContainerPlayer) player.openContainer).craftMatrix;
+
+            if (inv != null) {
+                boolean foundSelf = false;
+                for (int i = 0; i < inv.getSizeInventory(); i++) {
+                    if (ItemStack.areItemsEqual(inv.getStackInSlot(i), event.getItemStack())) {
+                        foundSelf = true;
```
detav, line 208: btw, before in the method `ItemStack itemStack = event.getItemStack()` was declared so doing `event.getItemStack()` everytime is not necessary.
```suggestion
                        if (itemStack.getItem() instanceof MachineItemBlock) {
                            if (MachineItemBlock.getMetaTileEntity(itemStack).keepsInventory()) {
                                event.getToolTip().add(I18n.format("gregtech.universal.clear_nbt_recipe.tooltip"));
                            }
                        }
```

Also, IMO the tooltip should only be in the resulting item, you did this before with `craftResult.getStackInSlot(0) == itemStack`. But as I said personal opinion.

LAGIdiot: This suggestions is not entirely correct as not every `MachineItemBlock` with `keepsInventory() ==true` has recipe to remove NBT data

detav: If they don't have the recipe, the checks should fail before, part of the reason I suggested the tooltip to only be in the resulting item.

LAGIdiot: There is nothing which prevent to have recipe with same input as output which does not clear NBT so in case of `keepsInventory() == true` it would assume that it is NBT clearing recipe which is not correct. Sure that chance is not big but someone may have these recipes generated automatically by other mod...I would be rather more defensive regarding this.

LAGIdiot:
Please rebase this against master as it is using old mappings and Gradle dependencies.

src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityQuantumChest.java
```diff
@@ -161,23 +184,53 @@ public boolean hasFrontFacing() {
...
+            importItems.setStackInSlot(0, new ItemStack(itemStack.getCompoundTag(NBT_PARTIALSTACK)));
         }
     }
 
+    @Override
+    public void writeItemStackData(NBTTagCompound itemStack) {
+        super.writeItemStackData(itemStack);
+        if (!this.itemStack.isEmpty()) {
+            itemStack.setTag(NBT_ITEMSTACK, this.itemStack.writeToNBT(new NBTTagCompound()));
+            itemStack.setLong(NBT_ITEMCOUNT, itemsStoredInside + this.itemStack.getMaxStackSize());
```
ALongStringOfNumbers, line 222: This line combined with the different virtualization is causing the duplicating chemical reactors and chemical baths issue that I pointed out in the body of the review. We should either prevent virtualizing different items than are in the output slot, or not make assumptions about having the max stack size of items in the output slot, because this might not be the case if the virtualized item differs from the item in the output slot.

src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityQuantumChest.java
```diff
@@ -161,23 +184,53 @@ public boolean hasFrontFacing() {
...
+        } else {
+            ItemStack partialStack = exportItems.extractItem(0, 64, false);
+            if (!partialStack.isEmpty()) {
+                itemStack.setTag(NBT_PARTIALSTACK, partialStack.writeToNBT(new NBTTagCompound()));
+            }
+        }
```
ALongStringOfNumbers, line 228: The different virtualization issue could be solved by serializing both the partial stack and any virtualized items, so that you don't have to make the assumption that the virtualized item and the partial stack are the same item.

src/main/resources/assets/gregtech/lang/en_us.lang
```diff
@@ -2273,6 +2275,8 @@ gregtech.machine.quantum_chest.iv.name=Black Hole-Quantum Chest III
 #Quantum tanks
 gregtech.machine.quantum_tank.tooltip=Compact place to store all your fluids.
 gregtech.machine.quantum_tank.capacity=Fluid Capacity: %,d mb
+gregtech.machine.quantum_tank.tooltip.name=Fluid Stored: §b%s
+gregtech.machine.quantum_tank.tooltip.count=Fluid Amount: §e%,d
```
ALongStringOfNumbers, line 2279: This localization should have `mb` after the number, so the tooltip would look like `64,000,000 mb` rather than `64,000,000`

ALongStringOfNumbers (changes requested):
I rebased this locally and took a stab at reviewing it. Other than my review comments, I found several other issues, mainly stemming from an (I think) missed implementation feature that the quantum chest can virtualize a different item than is in its output slot.

Currently, when you have a quantum chest in the crafting grid, the destruction tooltip will be displayed on other quatum chests in the inventory:
java_2021-06-13_23-35-17


In addition, the tooltip will be displayed, even when the recipe cannot be accomplished, such as when another item is in the crafting grid:
java_2021-06-13_23-44-34


Another Issue I saw (which is just general quantum chest behavior), The quantum chest can virtualize a different item than in its "output" slot. For example, I shift clicked a gold pickaxe into the quantum chest, and it went into the slot in the GUI, however, I was then able to shift click a diamond pickaxe into the quantum chest and it was virtualized. This left a golden pickaxe in the "output" slot and a diamond pickaxe virtualized, however when broke and replaced, a diamond pickaxe was in the "output" slot, and was also virtualized, meaning I transmuted the gold pickaxe to diamond.

Another issue, I placed a damaged diamond pickaxe into the "output" slot, and virtualized a full durability one, and then after breaking and replacing, both pickaxes had full durability.

Another issue stemming from being able to virtualize an item different from the "output" slot, I placed a chemical bath into the input slot, and virtualized a chemical reactor, and after breaking it and replacing it, I got 65 chemical reactors, which is a combination of the different virtualization and also maybe a combination of not setting the count correctly.

Another issue is that it voids items in the input slot of the quantum chest. I had placed 1 stack of rubies in the "output" slot, virtualized a stack of rubies, and placed a stack of ruby dust into the input slot. After breaking and replacing, the ruby dust was voided. In this case, I don't think the Ruby Dust should be stored in the chest when broken, but rather dropped into the world.

src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityQuantumChest.java
DStrand1: Instead of doing this, since I felt like an item getting virtualized "under" another (i.e., one item in output slot, none virtualized, and a different item is put into input and gets virtualized) was a confusing mechanic, I instead stopped this from happening. Now if the item in the output stack does not match the item in the input stack, the input item cannot be virtualized.

src/main/java/gregtech/common/metatileentities/storage/MetaTileEntityQuantumChest.java
```diff
@@ -161,21 +188,51 @@ public boolean hasFrontFacing() {
...
+    @Override
+    public void initFromItemStackData(NBTTagCompound itemStack) {
+        super.initFromItemStackData(itemStack);
+        if (itemStack.hasKey(NBT_ITEMSTACK, NBT.TAG_COMPOUND)) {
+            this.itemStack = new ItemStack(itemStack.getCompoundTag(NBT_ITEMSTACK));
+            if (!this.itemStack.isEmpty()) {
+                this.itemsStoredInside = itemStack.getLong(NBT_ITEMCOUNT);
+            }
+        } else if (itemStack.hasKey(NBT_PARTIALSTACK, NBT.TAG_COMPOUND)) {
+            importItems.setStackInSlot(0, new ItemStack(itemStack.getCompoundTag(NBT_PARTIALSTACK)));
```
ALongStringOfNumbers, line 217: Would it be better to set the itemstack in `exportItems` instead of setting it in the input slot and letting the chest transfer the items?

src/main/java/gregtech/common/ClientProxy.java
```diff
@@ -184,6 +184,53 @@ public static void addMaterialFormulaHandler(ItemTooltipEvent event) {
...
+                result = ((ContainerPlayer) player.openContainer).craftResult;
+            }
+
+            if (inv != null) {
+                ItemStack stackResult = result.getStackInSlot(0);
+
+                if (stackResult == event.getItemStack()) {
+                    if (!stackResult.isEmpty() && ItemStack.areItemsEqual(stackResult, event.getItemStack())) {
+                        String unlocalizedName = stackResult.getTranslationKey();
+                        for (String key : clearRecipes) {
```
ALongStringOfNumbers, line 214: Should this check also include a check the unlocalized name for the mod id, in case there are other mods with the same translation key? The quantum tank/chest have `gregtech` leading off their unlocalized name, so it should be a quick check for that. However, you might want to also include `gtadditions` for Gregicality, depending on if you were planning on adding higher tier tanks/chests.

ALongStringOfNumbers (approved):
Looks good, I just left two minor comments for consideration

ALongStringOfNumbers (approved):

PR #1556, opened 2021-04-02
Fix TE bottom side texture being rotated.

### LINKED ISSUE #1533: [BUG] AO on Bottom Face of MTE is Rotated
**Describe the bug**
For all Tile Entities, the bottom face has its ambient occlusion rotated by 90 degrees counterclockwise when compared to other blocks.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.12.1

**Setup**
_Playing_: Solo
_New world generated_: Yes

**Steps To Reproduce**
1) Place down any TE next to a normal block with AO (like casings).
2) You can observe that the AO is rotated on the bottom face for the TE.

**Expected behavior**
AO should be rendered properly on all faces of TEs.

**Screenshots**
image
You can see in this screenshot that the Tile Entities (including back 2 corners of formed EBF) are rotated poorly. All other faces are proper.

detav:
**What:**
Bottom side of TE had their textures rotated, I could not find within our codebase whether we do something with that side so I assume it is a bug with CCL.

**How solved:**
Applied a new UVTransformation only on that side to fix it.

**Outcome:**
Fixes #1533

**Additional info:**

### DIFF
--- a/src/main/java/gregtech/api/render/Textures.java
+++ b/src/main/java/gregtech/api/render/Textures.java
@@ -8,6 +8,7 @@ import codechicken.lib.vec.Cuboid6;
 import codechicken.lib.vec.Matrix4;
 import codechicken.lib.vec.TransformationList;
 import codechicken.lib.vec.uv.IconTransformation;
+import codechicken.lib.vec.uv.UVTransformationList;
 import gregtech.api.GTValues;
 import gregtech.api.util.GTLog;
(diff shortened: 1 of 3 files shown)


LAGIdiot (changes requested):
Thanks for fixing this. End result looks correct and I will trust that implementation you chose is best available.

There one small naming problem, please see to it.

LAGIdiot (approved):
Thanks for finishing this up

PR #1558, opened 2021-04-05
Fix glitches for SlotWidgets inside ScrollableListWidgets

warjort:
**What:**
Trying to use a slot widget inside a scrollable list widget leads to a number of glitchy behaviours.

The main cause of this is that the slots are NOT drawn inside the normal drawInBackground() handling where the ScrollableListWidget applies a scissor. Instead they are drawn "out of band" in ModularUIGui.

**How solved:**
1) When a ScrollableListWidget determines the positions of its components and its size, it now visits each of its widgets and gives them the option to remember its area that is actually visible.
The SlotWidget does this. It exposes this area through a new interace "gregtech.api.gui.IScissored" on its slot delegate.
When the ModularUIGui displays the slots, it checks for this new interface and uses it as a scissor to determine how much (if any) of the slot should be drawn.
The end result is that ModularUIGui can now do the same scissor handling for slots that would normally be done in drawInBackground if the slots behaved like other widgets.

2) The ScrollableListWidget now checks if a widget is within its visible area when it checks if a widget can be clicked.
Previously it just used the AbstractWidgetGroup handling which assumes all sub widgets were visible if the group is.

3) A number of places use Slot.isEnabled() to check for behaviours, like showing tooltips and some behaviours internal to SlotWidget itself. SlotWidget.isEnabled was not taking into account whether it was actually visible inside a ScrollableListWidget.
SlotWidget.isEnabled() has been changed to use the scissor from (1) when determining if it is enabled.

**Outcome:**
Putting a slot widget inside a ScrollableListWidget is not as glitchy as it was.
I can't guarantee I have spotted all bugs in this area.

**Additional info:**
The workbench (crafting station) has a ScrollableListWidget that contains slots, but notably it doesn't use the SlotWidget.
It has its own implementation. I wonder if that is because of the issues I have fixed here?

All of these changes except one only affect SlotWidgets inside ScrollableListWidgets. I don't think anybody else has done this, otherwise they would have seen these issues.

The one exception is the change to ScrollableListWidget.isWidgetClickable(Widget) which affects all subwidgets.
The fix seems correct even for non slot widgets, in that you shouldn't be able to click widgets outside the bounds of the
visible scroll area.


LAGIdiot:
Thanks for spotting and fixing this issue. I have idea what you did but just be sure that we are on same page can you please provide some pictures or gif of problems which this fixes.

warjort:
These are from my addon mod.
If you want to try yourself, change the CoverKeepInStock to have more than 6 slots (my current workaround is to disable the scrolling).

Here you can see the slots leaking outside the scroll area.



This second image is hard to see (the highlighted white area), but these are clickable slots over the player's inventory.

warjort:
I have been testing it by making a modpack with this PR and a version of my addon with more than 6 slots.

LAGIdiot (approved):
Good I had correct idea what it was. Code seems solid.

PR #1562, opened 2021-04-06
Implement ore weight tooltips

### LINKED ISSUE #1501: [Suggestion] Finding an individual ore's weight in a vein
Now that #1386 is merged, I am making this issue as a reminder of a feature that was desired to be included in that page, but could not be done at the time.

It is desired to have the information of an ore's individual weight in the vein, so that users could, at a glance, see the ore distribution of the vein. This would be displayed as a tooltip on the different ores in the "output" slots of the page.

The issue, as described somewhat here: https://github.com/GregTechCE/GregTech/pull/1386#issuecomment-766642825 is that the instances of `WeightRandomMatcherEntry` which has the desired information is private, and there were no other ways that I could see to get to the desired information

Some way to get the weighting of the individual ore in the vein, which would then be applied as a tooltip to the ores in the "output" slots on the JEI page.

**Additional context**
During the PR, I added TODO stubs for rendering this information as a tooltip, the information extraction just needs to happen first

ALongStringOfNumbers:
**What:**
This PR implements weighting tooltips for the recently added Ore Generation Page. The weighting tooltip is the percentage weight of a specific ore out of the total weight in the vein.

**How solved:**
A `default` method was added to the `FillerEntry` Interface allowing for retrieving the weight and the contents of the `Filler`. This was the only way that I could see this being implemented without changing some methods in the API.

**Outcome:**
Implements display of an ore's weight in the vein. Closes #1501.

**Additional info:**
java_2021-04-06_15-20-19

There should not be any compatibility issues due to the fact that a default method was used

### DIFF
--- a/src/main/java/gregtech/api/worldgen/config/FillerConfigUtils.java
+++ b/src/main/java/gregtech/api/worldgen/config/FillerConfigUtils.java
@@ -221,5 +221,10 @@ public class FillerConfigUtils {
         public Collection<IBlockState> getPossibleResults() {
             return blockStates;
         }
+
+        @Override
+        public List<Pair<Integer, FillerEntry>> getEntries() {
+            return randomList;
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


ALongStringOfNumbers:
Force Push to update onto current master to check for any MCP name issues, and also because I messed it up once.

LAGIdiot (approved):
This seems as reasonable implementation.

PR #1563, opened 2021-04-07
Fixed Global Renderer in MetaTileEntityTESR

Yefancy:
**What:**
1. Fixed an issue where the original logic only checked the MTEHolder but ignored the metatileEntity it held. The final rendering is still performed by a metatileEntity instead of its' holder, so I think this is an oversight. Some additional mods may require this function to be executed correctly when rendering block (e.g. the central monitor of gregicality)

2. Fixed an issue that ItemCollector has a bug which can copy item. #1578 

**How solved:**
1. added checking logic.

2. check whether EntityItem has been killed.

### DIFF
--- a/src/main/java/gregtech/api/render/MetaTileEntityTESR.java
+++ b/src/main/java/gregtech/api/render/MetaTileEntityTESR.java
@@ -125,15 +125,22 @@ public class MetaTileEntityTESR extends TileEntitySpecialRenderer<MetaTileEntity
 
     @Override
     public boolean isGlobalRenderer(MetaTileEntityHolder te) {
-        if(te instanceof IRenderMetaTileEntity && ((IRenderMetaTileEntity) te).isGlobalRenderer()) {
-            return true;
-        } else if(te instanceof IFastRenderMetaTileEntity && ((IFastRenderMetaTileEntity) te).isGlobalRenderer()) {
-            return true;
+        if (te instanceof IRenderMetaTileEntity) {
+            if (((IRenderMetaTileEntity) te).isGlobalRenderer()) {
+                return true;
+            }
+        } else if (te instanceof IFastRenderMetaTileEntity) {
+            if (((IFastRenderMetaTileEntity) te).isGlobalRenderer()) {
+                return true;
+            }
         }
-        if (te.getMetaTileEntity() instanceof IRenderMetaTileEntity && ((IRenderMetaTileEntity) te.getMetaTileEntity()).isGlobalRenderer()) {
-            return true;
-        } else {
-            return te.getMetaTileEntity() instanceof IFastRenderMetaTileEntity && ((IFastRenderMetaTileEntity) te.getMetaTileEntity()).isGlobalRenderer();
+        if (te.getMetaTileEntity() instanceof IRenderMetaTileEntity) {
+            if (((IRenderMetaTileEntity) te.getMetaTileEntity()).isGlobalRenderer()) {
+                return true;
+            }
+        } else if (te.getMetaTileEntity() instanceof IFastRenderMetaTileEntity){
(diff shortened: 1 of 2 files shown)


LAGIdiot (commented):
This seems strange to me I would understand if it was one long if else statement. Where you would check `instanceof` in if and then casted value and returned `isGlobalRenderer()`. And there was one global else return false.

But current implementations seems to be over combined and hard to read. Or am I missing something which prevents my idea from working?

Yefancy:
> This seems strange to me I would understand if it was one long if else statement. Where you would check `instanceof` in if and then casted value and returned `isGlobalRenderer()`. And there was one global else return false.
> 
> But current implementations seems to be over combined and hard to read. Or am I missing something which prevents my idea from working?

well, its easier to understand now. Logically equivalent, but previous code style was less elegant.

src/main/java/gregtech/api/render/MetaTileEntityTESR.java
```diff
@@ -125,10 +125,21 @@ public void renderTileEntityFast(MetaTileEntityHolder te, double x, double y, do
...
+            if (((IFastRenderMetaTileEntity) te).isGlobalRenderer()) {
+                return true;
+            }
+        }
+        if (te.getMetaTileEntity() instanceof IRenderMetaTileEntity) {
+            if (((IRenderMetaTileEntity) te.getMetaTileEntity()).isGlobalRenderer()) {
+                return true;
+            }
+        } else if (te.getMetaTileEntity() instanceof IFastRenderMetaTileEntity){
+            return ((IFastRenderMetaTileEntity) te.getMetaTileEntity()).isGlobalRenderer();
```
LAGIdiot, line 142: It looks still way too complicated to me. I would write it like this:
```suggestion
        if (te instanceof IRenderMetaTileEntity) {
			return ((IRenderMetaTileEntity) te).isGlobalRenderer();
        } else if (te instanceof IFastRenderMetaTileEntity) {
			return ((IFastRenderMetaTileEntity) te).isGlobalRenderer();
        }
		
        if (te.getMetaTileEntity() instanceof IRenderMetaTileEntity) {
			return ((IRenderMetaTileEntity) te.getMetaTileEntity()).isGlobalRenderer();
        } else if (te.getMetaTileEntity() instanceof IFastRenderMetaTileEntity){
            return ((IFastRenderMetaTileEntity) te.getMetaTileEntity()).isGlobalRenderer();
```
But my code is invalid if `te` can be `I*RenderMetaTileEntity` and `MetaTileEntity` it contains can be `I*RenderMetaTileEntity` at same time. So can it?

Yefancy: This change also work well at present, as far as I know no `MetatileEntityHolder` implements these two interfaces now, so it's still essentially a judgment for `MetatileEntity`. However,  `te` and `MetatileEntity` should be cared at same time. Logically, we should ensure that it returns true whenever one of `MetatileEntityHolder` or `MetatileEntity` it contains has a `GlobalRenderer` requirement. In addition, returning true is still work fine for `MetatileEntityHolder` even if `GlobalRenderer` is not required.

Yefancy:
also #1578
I don't think these are very complicated fixes, so I put them together. If I have to separate, I can separate the PR

src/main/java/gregtech/api/render/MetaTileEntityTESR.java
LAGIdiot: I understand let's leave current implementation.

LAGIdiot:
Please don't put unrelated fixes in same PR. I can't accept it now.

Yefancy:
> Please don't put unrelated fixes in same PR. I can't accept it now.

i reverted it. it just work for global renderer now

PR #1565, opened 2021-04-08
Fix git using wrong folder, upgrade jgit and gradle

warjort:
**What:**
build.gradle.kts currently initialises jgit with file(".")
This is wrong, since that is the CWD (current working directory)

**How solved:**
Changed the build to initialise git with "projectDir" which is the project root directory

**Outcome:**
The build now works when your CWD is not the project folder.
Which is pretty much always if you use eclipse or vscode.

**Additional info:**
While I was doing this, I have made 2 other "housekeeping" changes:

* Upgraded jgit to 5.8 from 5.5, this is the latest version before eclipse started mandating java 11
* Upgraded gradle to 4.10.3 from 4.7 which was the last release in the 4.x series


eutro:
Updating the Gradle wrapper could cause issues in itself, the build system can be quite fragile.

warjort:
> Updating the Gradle wrapper could cause issues in itself, the build system can be quite fragile.

I have been hacking the build for a while now to get it to load into eclipse. I always modify it it to use 4.10.3 without issue.
Of course, I don't test the maven or curseforge parts of the build.

LAGIdiot:
I am using only IDEA and I already has some hacks in place for it to work because of environmental issues. So I am probably wrong person to test this. So if any one can try it I would be glad.
I am not against upgraded to latest version with same constrains as current one. Even though build is fragile we should try it as it may fix problems for other (even potential) contributors.

ALongStringOfNumbers:
I personally have not had any issues in projects when I used 4.10.3 for the gradle version, however I am not sure if it entirely necessary. 

However, someone in discord was talking about how the git initialization change fixed an issue that they were having, so that part of the PR is important, and may need some testing.

phantamanta44:
had this same issue with the project cwd in IDEA; if the version updates are too drastic of a change, the change to the git repo dir should at least be merged separately

LAGIdiot:
Base on feedback lets use this. I will merge this in master after today release and we will have whole cycle to test it (and if it does break build then we will have time to fix it).

LAGIdiot (approved):

LAGIdiot (approved):

PR #1568, opened 2021-04-11
Expose Some Fields to CraftTweaker

### LINKED ISSUE #1526: [BUG] Material's blastFurnaceTemperature not exposed to CraftTweaker
**Describe the bug**
blastFurnaceTemperature should be exposed to crafttweaker but it is not.

**Versions**
Forge: 14.23.5.2854
GTCE: 1.11.1.632
_Addons:_ CraftTweaker2 1.12-4.1.20.574

**Setup**
Both solo and multiplayer

**Steps To Reproduce**
1) Try do this in a CraftTweaker script:
if(MaterialCasting.toIngot(mat).blastFurnaceTemperature > 0) {}
2) Get an error as blastFurnaceTemperature is not exposed to CraftTweaker.

**Expected behavior**
returns the corresponding blastFurnaceTemperature field of the material.

Thanks!

serenibyss:
**What:**
This PR simply exposes `IngotMaterial#blastFurnaceTemperature` and `SolidMaterial#toolAttackDamage` to CraftTweaker as ZenProperties.

Blast furnace temperature was requested via issue #1526, and tool attack damage was requested in Discord, and was the only field in `SolidMaterial` to not have a ZenProperty annotation, so it makes sense to apply it.

**Outcome:**
- Blast Furnace Temperature and Tool Attack Damage exposed to CraftTweaker.
- Closes #1526

None expected.

### DIFF
--- a/src/main/java/gregtech/api/unification/material/type/IngotMaterial.java
+++ b/src/main/java/gregtech/api/unification/material/type/IngotMaterial.java
@@ -64,9 +64,11 @@ public class IngotMaterial extends SolidMaterial {
     /**
      * Blast furnace temperature of this material
      * Equal to zero if material doesn't use blast furnace
-     * If below 1000C, primitive blast furnace recipes will be also added
+     * If below 1000C, primitive blast furnace recipes will be also added.
+     * If above 1750C, a Hot Ingot and its Vacuum Freezer recipe will be also added.
      */
-    public final int blastFurnaceTemperature;
+    @ZenProperty("blastFurnaceTemp")
+    public int blastFurnaceTemperature;
 
     /**
      * If set, cable will be generated for this material with base stats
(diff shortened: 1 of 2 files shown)


Archengius:
You could already set attack damage and EBF temperature before through material's constructor.
These fields are final for a good reason: people should *not* mess with already created materials.

DStrand1:
If this is the stance you think we should be taking, then we should remove a lot of other ZenMethods and ZenProperties because existing materials can already be changed significantly. For example:
- setCableProperties
- setFluidPipeProperties
- oreMultiplier
- oreByproducts
- directSmelting
- washedIn
...

I could go on, but is removing this functionality really what we want? All of these may be intended for use with newly generated CraftTweaker materials, but all of them function fully with our existing ones, and our material handlers are set up fully to do this without any issues.

Archengius:
yep, technically we're not really following it, but it doesn't mean we shouldn't even try
I would still like to avoid situations where you can break GTCE native recipes by changing vanilla materials

DStrand1:
I will update this to an accessor of a final variable instead of a non-final field.

LAGIdiot (approved):
Okay for now let's leave it open as read only. Other variant can be discussed later.

PR #1569, opened 2021-04-11
Rework Pipe Rendering

serenibyss:
I am opening this PR as a draft before I have finished everything up to get some opinions on the style these pipes have taken.

**What:**
Currently, our pipes are using some CCL model generation to have a non-square pipe. However, the corner models were never fully implemented, and there are issues with how the texture is mapped onto the pipes themselves. This PR aims to address those issues and complete the rendering that these pipes could have had. Also, the Item model for the pipes has been updated as well, and is now the "straight connection" model instead of the "zero connections" model. I find this model looks a lot more like a pipe, so it is more fitting to the Item.

**Outcome:**
image
image
image
image

**Additional Info:**
With the Inventory Pipes PR, I would like to see our pipes (Fluid and Inventory) have better models on release. However, as this PR is a draft, I still have some issues to tackle.

**Issues Remaining:**
- Pipes have no rendered endpoints, just an empty hole. The texture exists but needs to be mapped onto the model.
    - This issue is clearly visible in the current screenshots.
    - This is the most important issue of these by far. I do not expect it to be too difficult, though it may take me some time to figure out.
- Lighting is being looked into, as it is using flat lighting rather than smooth. Also, shading is being looked into.
    - Smooth lighting, in my opinion, is very necessary for this PR to be finished. I think it should not be difficult, just need to figure out the right place and way to enable it.
    - Shading, I think may be far too difficult to do properly in this PR. I have someone who is much better at rendering looking into this, but it will probably not be implemented.
- Pipes placed initially face a constant direction, and should instead face the player (or perpendicular to the player). In other words, it should not be a static direction.
    - This may require some changes to how the connections are stored in the pipes themselves, but ultimately I will ensure that no behavior is affected outside of rendering.
- Pipes, when all connections are lost, are reverted to a static position. They should have some "memory" of the most recent connection and stay facing that way despite no connections.
    - Similar to above.

None expected.

### DIFF
--- a/src/main/java/gregtech/api/capability/impl/ItemFuelInfo.java
+++ b/src/main/java/gregtech/api/capability/impl/ItemFuelInfo.java
@@ -15,6 +15,6 @@ public class ItemFuelInfo extends AbstractFuelInfo {
     }
 
     public String getFuelName() {
-        return itemStack.getUnlocalizedName();
+        return itemStack.getTranslationKey();
     }
 }
--- a/src/main/java/gregtech/common/pipelike/fluidpipe/FluidPipeType.java
+++ b/src/main/java/gregtech/common/pipelike/fluidpipe/FluidPipeType.java
@@ -6,9 +6,9 @@ import gregtech.api.unification.ore.OrePrefix;
 public enum FluidPipeType implements IMaterialPipeType<FluidPipeProperties> {
 
     TINY_OPAQUE("tiny", 0.20f, 1, OrePrefix.pipeTiny, true),
-    SMALL_OPAQUE("small", 0.30f, 2, OrePrefix.pipeSmall, true),
-    MEDIUM_OPAQUE("medium", 0.35f, 4, OrePrefix.pipeMedium, true),
-    LARGE_OPAQUE("large", 0.40f, 8, OrePrefix.pipeLarge, true);
+    SMALL_OPAQUE("small", 0.35f, 2, OrePrefix.pipeSmall, true),
+    MEDIUM_OPAQUE("medium", 0.50f, 4, OrePrefix.pipeMedium, true),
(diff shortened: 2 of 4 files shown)


src/main/java/gregtech/api/capability/impl/ItemFuelInfo.java
```diff
@@ -15,6 +15,6 @@ public ItemFuelInfo(ItemStack itemStack, int fuelRemaining, int fuelCapacity, in
     }
 
     public String getFuelName() {
-        return itemStack.getUnlocalizedName();
+        return itemStack.getTranslationKey();
     }
```
DStrand1, line 19: Not sure why this change got carried into this PR, but I will make sure it is a non-issue by the time this PR is ready to be merged.

warjort:
I don't know if you aware, but a single pipe doesn't seem to be rendering its far side?

warjort:
Will your first point fix this existing issue when pipes of different sizes are placed next to each other.


It looks more pronounced in your version

DStrand1:
> I don't know if you aware, but a single pipe doesn't seem to be rendering its far side.
> I guess this is your end not rendering issue?

Yes, this is on my end. I have not bothered to disable culling since the endpoint texture still needs to be applied.

> Will your first point fix this existing issue when pipes of different sizes are placed next to each other.

Yes, that should be fixed when endpoint textures are applied, though I will have to be careful to make sure that the “opening” of the endpoint doesn’t poke out around the edges of the endpoint when the pipe is stepped down to a smaller one.

LAGIdiot:
This looks lovely and I would like to see this finished. I see that there are still quite few problems which need to be fixed. I am totally oblivious regarding rendering and aesthetics so I will probably not be any helpful with this.

PR #1570, opened 2021-04-11
Fluid Regulator

### LINKED ISSUE #952: Suggestion: Robot Arm Style Settings for Pumps
Crafting can often require very particular fractional-bucket quantities of fluids. It would be extremely useful to be able to route fluids precisely with pumps like you can items with Robot Arms.

The main use cases would be stocking multiple machines with the same fluid to parallelize crafting without having portions left over in them, and to avoid putting more than necessary into fluid input hatches for multiblock structures like the EBF since it is impossible to extract fluid out of them once inserted.

Opening this ticket in response to https://github.com/GregTechCE/GregTech/issues/793#issuecomment-527245036.

hjae78:
This PR adds a new set of cover items, the Fluid Regulators. These covers are designed to act as the Robot Arm but for fluids.

There are 3 modes, just like the Robot Arm;
Transfer Any: works exactly like the normal Electric Pump.
image
Supply Exact: moves fluids in only the exact quantity that you specify.
image
Keep Exact: tries to keep exactly the quantity specified.
image

The main logic of the cover itself is finished as far as I can tell so I submitted so that there would be more time for people to look over it and comment on it while I work on fixing up the last things.

Closes #952 

TODO: Localization, fix text overspilling when in Keep Exact/Supply Exact mode and the amount becomes too large, proper textures, recipes


idcppl:
P.S. missing lang file stuff.

DStrand1 (changes requested):
I was asked by @hjae78 to review this PR, so here goes

DStrand1:
Additionally, I found an issue where the amount of fluid to keep in stock is not always within an even bucket amount when set to "B/s" mode. For example, when in B/s mode, set to Keep Exact 4:
image

Then when I switch it to mB/s mode, it shows that it is set to Keep Exact 4008:
image

The behavior observed in-game is that internally, this value is 4008, as it transferred 4008mb of fluid rather than the expected 4B.

DStrand1:
I edited the opening comment to state that this PR will close #952

src/main/java/gregtech/common/covers/CoverFluidRegulator.java
```diff
@@ -0,0 +1,245 @@
...
+        blockPos.setPos(coverHolder.getPos()).move(attachedSide);
+        TileEntity tileEntity = coverHolder.getWorld().getTileEntity(blockPos);
+        blockPos.release();
+        IFluidHandler fluidHandler = tileEntity == null ? null : tileEntity.getCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, attachedSide.getOpposite());
+        IFluidHandler myFluidHandler = coverHolder.getCapability(CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY, attachedSide);
+        if (fluidHandler == null || myFluidHandler == null) {
+            return 0;
+        }
+        IFluidHandler sourceHandler;
+        IFluidHandler destHandler;
```
LAGIdiot, line 54: This is already handled by `doTransferFluids` on Pump you should just override `doTransferFluidsInternal` and continue form there.

src/main/java/gregtech/common/covers/CoverFluidRegulator.java
```diff
@@ -0,0 +1,245 @@
...
+                if (sourceFluid != null && sourceFluid.amount > 0) {
+                    destHandler.fill(sourceFluid, true);
+
+                    fluidLeftToTransfer -= sourceFluid.amount;
+                    if (fluidLeftToTransfer == 0) break;
+                }
+            }
+        }
+        return transferLimit - fluidLeftToTransfer;
+    }
```
LAGIdiot, line 131: These are both quite similar would be possible to extract some parts to methods and pass in just values. Possible place them to GTFluidUtils?

src/main/java/gregtech/common/covers/CoverBehaviors.java
```diff
@@ -61,6 +61,16 @@ public static void init() {
         registerBehavior(37, new ResourceLocation(GTValues.MODID, "machine_controller"), MetaItems.COVER_MACHINE_CONTROLLER, CoverMachineController::new);
         registerBehavior(38, new ResourceLocation(GTValues.MODID, "smart_filter"), MetaItems.SMART_FILTER, (tile, side) -> new CoverItemFilter(tile, side, "cover.smart_item_filter.title", Textures.SMART_FILTER_FILTER_OVERLAY, new SmartItemFilter()));
         registerBehavior(39, new ResourceLocation(GTValues.MODID, "facade"), MetaItems.COVER_FACADE, CoverFacade::new);
+
+        registerBehavior(40, new ResourceLocation(GTValues.MODID, "fluid.regulator.lv"), MetaItems.FLUID_REGULATOR_LV, (tile, side) -> new CoverFluidRegulator(tile, side, GTValues.LV, 1280));
+        registerBehavior(41, new ResourceLocation(GTValues.MODID, "fluid.regulator.mv"), MetaItems.FLUID_REGULATOR_MV, (tile, side) -> new CoverFluidRegulator(tile, side, GTValues.MV, 5120));
```
LAGIdiot, line 66: Maybe we could invest to system which will create these for us more easily as we can take names from array we could have all of it in arrays and just run it on correct set of them.

Given that you need these arrays in MachineRecipeLoader anyway.

What you think?

LAGIdiot (changes requested):
Thanks for your hard work. I have commented on few places I feel needs adjustments.

Also I have question regarding textures are they same as electric pumps or is it just me?

hjae78:
> Thanks for your hard work. I have commented on few places I feel needs adjustments.
> 
> Also I have question regarding textures are they same as electric pumps or is it just me?

Thanks for your review, I'll start addressing your comments later today

The textures are the same as pumps just flipped. This is how it was in GT5 so I felt it was appropriate.

LAGIdiot:
> The textures are the same as pumps just flipped. This is how it was in GT5 so I felt it was appropriate.

They are flipped in the code and I missed it? Because when I open LV variant it looks exactly same as pump.

hjae78:
> They are flipped in the code and I missed it? Because when I open LV variant it looks exactly same as pump.

image

src/main/java/gregtech/api/util/GTUtility.java
```diff
@@ -716,4 +726,23 @@ public static int getRedstonePower(World world, BlockPos blockPos, EnumFacing si
     public static int getRandomIntXSTR(int bound) {
         return random.nextInt(bound);
     }
+
+    public static String formatLongToCompactString(long value, int precision) {
```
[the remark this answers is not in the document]
hjae78: Okay, I don't see a util class that I immediately think this is correct. Do you have any suggestions?

LAGIdiot:
Hm OK, but when it is done in code do we need separate textures for them? I mean files are identical to pumps right? Or am I missing something?

hjae78:
> Hm OK, but when it is done in code do we need separate textures for them? I mean files are identical to pumps right? Or am I missing something?

I don't know how to flip textures using jsons so I did it using textures

LAGIdiot:
Ah my bad I was looking at `fluidregulator.lv` which are already in GTCE not ones you are adding. And these old does look exactly like pump so I was assuming you are flipping them in code.

Maybe you should replace them with yours so we don't have wrong unused files in jar, also this will not break binary compatibility if someone is using them and I am also not concern regarding naming as it is all over the place. What you think?

hjae78:
Ah okay, I noticed these lines in `en_us.lang` which I thought that maybe I had added at some point. Everything called `fluidregulator.xx` is not added by me and as far as I can tell ununsed. I will remove them in an upcoming push

src/main/java/gregtech/common/covers/CoverFluidRegulator.java
hjae78: It might be possible to extract the common parts of these but after spending some time looking at the code no obvious, easy to read solution comes to me. I could move them to GTFluidUtils. Currently I followed the same workflow as in CoverRoboticArm which keeps all these methods in the same class.

src/main/java/gregtech/common/covers/CoverBehaviors.java
hjae78: Yes, this sounds like a good idea. I'm not sure where I should place this array, what do you think?

LAGIdiot:
I would rather not remove them as someone outside may use them (binary compatibility break). I thing we should replace them with yours and change references to them in your PR.

src/main/java/gregtech/common/covers/CoverBehaviors.java
LAGIdiot: I don't feel like this has API potential so we should leave it in common. Maybe something like `GTValuesCommon` as we may store there later more of these.

src/main/java/gregtech/common/covers/CoverFluidRegulator.java
```diff
@@ -0,0 +1,245 @@
...
+            int canInsertAmount = destHandler.fill(sourceFluid, false);
+            if (canInsertAmount == amountToDrainAndFill) {
+                sourceFluid = sourceHandler.drain(sourceFluid, true);
+                if (sourceFluid != null && sourceFluid.amount > 0) {
+                    destHandler.fill(sourceFluid, true);
+
+                    fluidLeftToTransfer -= sourceFluid.amount;
+                    if (fluidLeftToTransfer == 0) break;
+                }
+            }
```
LAGIdiot, line 128: This is the part I feel can be extracted as it is quite similar to one in `doTransferExact` and also `GTFluidUtils.transferFluids` with just few params

src/main/java/gregtech/common/covers/CoverFluidRegulator.java
LAGIdiot: I have marked place which I feel is extractable via another comment. Also I feel we should leverage `GTFluidUtils` as much as is reasonable because we are manipulating fluids from quite few places

src/main/java/gregtech/api/util/GTUtility.java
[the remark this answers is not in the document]
LAGIdiot: Just create new one for it. Maybe `TextFormattingUtils` (not sure with last word as we have here Utils\Util\Utilities). Best place will be under `api/util` as we may place more similar ones in it.

src/main/java/gregtech/common/covers/CoverFluidRegulator.java
```diff
@@ -0,0 +1,248 @@
...
+            FluidStack sourceFluid = tankProperties.getContents();
+            if (sourceFluid == null || sourceFluid.amount == 0 || !fluidFilter.test(sourceFluid)) continue;
+            sourceFluid.amount = keepAmount;
+            FluidStack destFluid = destHandler.drain(sourceFluid, false);
+            int amountToDrainAndFill;
+            //no fluid in destination
+            if (destFluid == null) {
+                amountToDrainAndFill = Math.min(keepAmount, fluidLeftToTransfer);
+            //if the amount of fluid in the destination is sufficient or the destinations fluid isnt equal to the sources
+            //how to check if destHandler is full?
```
ALongStringOfNumbers, line 119: To check if the destination tank is full, you could check the tank properties to the destination handler and get the capacity and contents and compare them.

ALongStringOfNumbers (changes requested):
Currently when placing a Fluid Pump onto a machine, it will show the Fluid Regulator configuration screen.

When you break the cover with a crowbar, it will turn back into the Fluid Pump.

Additionally (minor nitpick), the Keep Exact and Supply Exact amount is not centered in the GUI, like it is for Robot Arms:

java_2021-05-28_14-04-51

java_2021-05-28_14-05-32

src/main/java/gregtech/common/GTValuesCommon.java
```diff
@@ -0,0 +1,16 @@
...
+public class GTValuesCommon {
+    
+    public static final MetaItem<?>.MetaValueItem[] PUMPS = {ELECTRIC_PUMP_LV, ELECTRIC_PUMP_MV, ELECTRIC_PUMP_HV, 
+            ELECTRIC_PUMP_EV, ELECTRIC_PUMP_IV, ELECTRIC_PUMP_LUV, ELECTRIC_PUMP_ZPM, ELECTRIC_PUMP_UV};
+
+    public static final MetaItem<?>.MetaValueItem[] FLUID_REGULATORS = {FLUID_REGULATOR_LV, FLUID_REGULATOR_MV, 
+            FLUID_REGULATOR_HV, FLUID_REGULATOR_EV, FLUID_REGULATOR_IV, FLUID_REGULATOR_LUV, FLUID_REGULATOR_ZPM, 
+            FLUID_REGULATOR_UV};
+    
+}
```
LAGIdiot, line 16: I was expecting here bit different values like pump through put, but you were able to solve these with math.

For values which are currently here it would be better to place them in `MetaItems` file as there are already some arrays of other items

LAGIdiot (commented):
Also I am still not sure how to correctly handle fact that there are already files for `fluidregulator`. What you (or any dev) think about my proposition to update them and reuse them?

hjae78:
> Also I am still not sure how to correctly handle fact that there are already files for `fluidregulator`. What you (or any dev) think about my proposition to update them and reuse them?

I think they should be left unused. I think these are simply a leftover from GT5 and I know of no addons that make any use of it. If there are non public addons that use them they could update to use the new resources. Primarily I think this is the best solution since the current resources don't follow the current naming convention, e.g. `fluidregulator.lv` instead of `fluid.regulator.lv`.

src/main/java/gregtech/common/covers/CoverFluidRegulator.java
hjae78: I think that was an old comment that I made for myself. In my testing I have run into no issues without including this check. I resolved the issue of the pumps accidentally being fluid regulators. Regarding the text not being centered, this stems from `AdvancedTextWidget::drawInBackground`. The other fields are `SimpleTextWidget` which centers the text in that method. I don't feel this falls in the scope of this PR and would be better suited in a separate one.

ALongStringOfNumbers: I agree that fixing this widget is not in the scope of this PR. Would you mind making an issue for it through, so it is not forgotten?

src/main/java/gregtech/api/util/GTUtility.java
```diff
@@ -716,4 +716,5 @@ public static int getRedstonePower(World world, BlockPos blockPos, EnumFacing si
     public static int getRandomIntXSTR(int bound) {
         return random.nextInt(bound);
     }
+
```
ALongStringOfNumbers, line 719: This space could be removed

src/main/java/gregtech/loaders/recipe/MachineRecipeLoader.java
```diff
@@ -689,6 +690,17 @@ private static void registerAssemblerRecipes() {
...
+
+        for (int i = 0; i < circuitTiers.length; i++) {
+            RecipeMaps.ASSEMBLER_RECIPES.recipeBuilder()
+                    .inputs(PUMPS[i].getStackForm())
+                    .input(OrePrefix.circuit, circuitTiers[i], 2)
+                    .outputs(FLUID_REGULATORS[i].getStackForm())
+                    .EUt((int) (GTValues.V[i] / 2))
+                    .duration(400)
+                    .buildAndRegister();
+        }
```
ALongStringOfNumbers, line 703: These recipes could be moved to the file added in #1633

src/main/java/gregtech/common/covers/CoverFluidRegulator.java
```diff
@@ -0,0 +1,248 @@
...
+    private static final String supplyKey = "SupplyAmount";
+    private static final String keepKey = "KeepAmount";
+
+    public CoverFluidRegulator(ICoverable coverHolder, EnumFacing attachedSide, int tier, int mbPerTick) {
+        super(coverHolder, attachedSide, tier, mbPerTick);
+        this.transferMode = TransferMode.TRANSFER_ANY;
+    }
+
+/*    @Override
+    protected int doTransferFluids(int transferLimit) {
```
ALongStringOfNumbers, line 39: If this commented out section was handled in a different method, this commented out section could be deleted

ALongStringOfNumbers (commented):
Some formatting/cleanup suggestions

src/main/java/gregtech/common/covers/CoverFluidRegulator.java
hjae78: Done

LAGIdiot (changes requested):
As currently we don't know of anyone using old Fluid regulator textures then it should be save to remove them right? Am in favor of removing them so we take less unused baggage.

In game it works as expected.

I feel we may want to create issue for updating pump GUI as it is not aligned to right side of inventory and given the max throughput 100 buttons for updating it are quite inefficient. But that should not be concern for this.

src/main/java/gregtech/common/covers/CoverFluidRegulator.java
```diff
@@ -0,0 +1,217 @@
...
+        return 0;
+    }
+
+    protected int doTransferExact(int transferLimit, IFluidHandler sourceHandler, IFluidHandler destHandler, Predicate<FluidStack> fluidFilter, int supplyAmount) {
+        int fluidLeftToTransfer = transferLimit;
+        for (IFluidTankProperties tankProperties : sourceHandler.getTankProperties()) {
+            if (fluidLeftToTransfer < supplyAmount)
+                break;
+            FluidStack sourceFluid = tankProperties.getContents();
+            if (sourceFluid == null || sourceFluid.amount == 0 || !fluidFilter.test(sourceFluid)) continue;
```
ALongStringOfNumbers, line 66: Do we want to add another check here to check if the fluid being transferred matches the one in the destination tank, to short circuit some more logic if the fluids don't match?

src/main/java/gregtech/loaders/recipe/ComponentRecipes.java
```diff
@@ -379,5 +383,17 @@ public static void register() {
                     .outputs(MetaItems.ELECTRIC_PUMP_IV.getStackForm())
                     .duration(100).EUt(7680).buildAndRegister();
         }
+
+        Material[] circuitTiers = new Material[] {MarkerMaterials.Tier.Basic, MarkerMaterials.Tier.Good, MarkerMaterials.Tier.Advanced, MarkerMaterials.Tier.Extreme};
```
ALongStringOfNumbers, line 387: This should go up to IV tier, `MarkerMaterials.Tier.Elite` to match all the other component recipes in GTCE being craftable up to IV.

ALongStringOfNumbers (commented):
This looks good to me, just a couple minor comments and requests to address and it will be good for merging in my opinion. Others are welcome to review as well.

src/main/java/gregtech/common/covers/CoverFluidRegulator.java
hjae78: We could but this is already handled [here](https://github.com/GregTechCE/GregTech/blob/ba57b3f7f7dc548cf0abd8d65bca173a1977868e/src/main/java/gregtech/api/util/GTFluidUtils.java#L56). I used the same logic flow as the Pumps use to remain consistent and they don't do this.

ALongStringOfNumbers (approved):

LAGIdiot (approved):
Thanks for your hard work. Now it feels complete to me.

PR #1572, opened 2021-04-12
Fix 12th layer for DT

serenibyss:
**What:**
This PR fixes an issue where the Distillation Tower would not form with 12 layers for fluid output hatches, despite the RecipeMap supporting up to 12 outputs. This was not an issue for us, since none of our recipes use all 12 fluid outputs, but this could easily be an issue for addons and modpacks.

**Outcome:**
- DT can now properly have 12 layers

None expected.


LAGIdiot (approved):
Thanks for spotting this and fixing it.

PR #1573, opened 2021-04-16
Open Up Machine Base Texture API Method

serenibyss:
**What:**
This PR is extremely simple; all it does is set the method `getBaseRenderer` to protected instead of private in:
- `TieredMetaTileEntity`
- `SteamMetaTileEntity`
- `SteamBoiler`

Addons and other machines we choose to add can now override the base texture of these types of MetaTileEntity if they so choose.

**Outcome:**
- Open up machine base texture method to API

None expected.


LAGIdiot (approved):
This change seems reasonable.

PR #1574, opened 2021-04-18
Rework Some GUI Aspects

### LINKED ISSUE #1524: [BUG] Slots in MAX tier buses exceed GUI
**Describe the bug**
The slots in the MAX tier input and output buses exceed the GUI

**Versions**
Forge: 14.23.5.2847
GTCE: 1.12.0.662
_Modpack:_ None
_Addons:_ None

**Setup**
Playing Solo
_New world generated_ Yes

**Steps To Reproduce**
1) Place a MAX import/export bus and open then to show their GUI

**Expected behavior**
For the slots to not exceed the GUI of the bus

**Screenshots**
java_2021-03-08_19-17-31

javaw_2021-03-08_19-29-00

DStrand1:
**What:**
This PR fixes a lot of very small GUI issues, as well as redesigning some of the more outdated machine UI buttons, and adding a logo.
electrolyzer

**How solved:**
Here I will enumerate a list of GUI aspects that were changed:
- Player Inventory was offset by 1px to the right, this has been fixed
- SimpleMachine title was offset by 2px to the left, this has been fixed. It is now in-line with the auto-output buttons
- Multiblock GUI title and text was offset 2px up and 1px to the left, causing some of the text to overlap with the border. This has been fixed
- Overclocking button has been redesigned to look more like the other buttons*, and rearranged to fit better
- `RecipeMapGroupOutput` button swapping between item and fluid outputs has been redesigned to look more like the other buttons, and rearranged to look better
- MAX Input/Output Bus GUI has been fixed. Previously had slots overflowing outside of the background

*: I would like to go into some more depth on this button. We needed the button to be wider than 18px like other buttons for 2 reasons:
- ULV Tier setting
- OFF setting
I removed the ULV-tier setting, because it was redundant. The behavior is identical if set to ULV or to OFF. I also changed the OFF button to instead be `X` saving space. I realize that Gregicality will have issues with its 3 letter tier names, and SoG similarly. However, since the button's edges are a few pixels slimmer, we are left with this:
zpmbutton
which I personally think is sufficient for 3-letter tiers past what GTCE offers natively. As a developer of Gregicality, I can say I am happy with this text. If SoG developers feel differently, I can work on a fix for them to address it.

Lastly, I added a logo in several places, as you can see above in the Electrolyzer images. I included it in a few key places:
- Singleblock Electric Machines and Generators
- Steam/High-Pressure Steam Machines
- Quantum Tanks/Chests
- GTCE JEI recipes

Notably, I did not include it for
- Multiblock parts (buses, hatches)
- Multiblock Controllers
- Multiblock Controller JEI layout
- GTCE Chests
- A few other random ones, like the Locked Safe

As was suggested in Discord, I left some hooks for addons to override or use to apply their own logo in place of GTCE's.

**Outcome:**
- Many minor GUI issues fixed
- Logo added to machines and JEI recipes
- Closes #1524

**Additional info:**
Screenshot gallery:

Basic machine:
alloysmelter

Standard JEI page:
jeiEBF

JEI page with many outputs:
jeiElectrolyzer

Steam machine (which actually previously had the GT5 logo on it):
steam

New MAX Bus GUI:
maxBus

I tested the logo placement, ensuring that it does not overlap with any other GUI element, for all GTCE Tile Entities and all JEI page structures, and have seen no issues. However, another review over them would be great, as it is possible I missed something minor.

None expected


warjort:
> I removed the ULV-tier setting, because it was redundant. The behavior is identical if set to ULV or to OFF.

It's not. I made the same mistake.

ULV is special, you can overclock *within* the ULV tier. It's the only tier that does this.
A ULV recipe that takes 1 or 2 EU will overclock to 4 or 8 EU which is still ULV.

>  I also changed the OFF button to instead be X saving space.

I originally suggested using blank or "--", but @LAGIdiot wanted to use OFF. 
I think XX would make more sense than X since everything else is 2 or more chars?
It's just a language issue anyway so people can change it to whatever they like with a resource pack.

warjort:
To me that recipe group button looks odd. 

It doesn't look like its aligned with anything, even though I can force myself to see it is actually aligned with the 3rd to last column of the player inventory.

I would also suggest that we find a "standard" place for the logo. 
It looks weird with it moving to different places on those JEI screens.

DStrand1:
The logo is actually in a fixed position in all GUIs, save for the Electrolyzer/Centrifuge JEI page, where it is simply shifted down slightly to not overlap with the slots. Though maybe leaving it in that bottom left corner for all JEI recipes makes more sense. As for the group output button, it is symmetrical with the overclock button around the battery charger. There really isnt a great place for this button period, but this is at least much better than before in my opinion.

For the overclock button, personally I think setting overclocking to ULV even with very low EU/t recipes existing is unnecessary and confusing to players, especially since we do not have ULV-tier machines.

warjort:
Finally, I don't know what the policy should be about branding other people's machines?
Or other people using the GTCE brand.

I would guess you might want load the texture based on who registered the machine?

Something like:

mod-id:textures/logo.png
mod-id = MetaTileEntity.metaTileEntityId.getNamespace()

warjort:
> As for the group output button, it is symmetrical with the overclock button around the battery charger. There really isnt a great place for this button period, but this is at least much better than before in my opinion.

To me it would make more sense if it were aligned with the output slots above it since it is related to them?
I guess that's why it looks odd to me. It looks disjointed from what it controls.

DStrand1:
This was my initial idea as well, but feedback from @LAGIdiot was to try and align to player inventory. Here is the picture of it aligned to the outputs:

D48BEFAA-C6BA-4D5E-888A-D0CA53E626B2

warjort:
> I guess that's why it looks odd to me. It looks disjointed from what it controls.

I wonder if a better UI design would be to place some small tabs above or to the right of the output slots to switch modes 
and get rid of the button altogether?

warjort:
> This was my initial idea as well

It looks better to me. But it prbably makes it more difficult to add new buttons next to it in future.

LAGIdiot:
I think that aligning that item/fluid button to inventory slots is more reasonable for future additions of other buttons (if we need it).

As was already mentioned ULV overclocking setting is valid one. So please return it back.

It think we should extract logo to separate PR so it does not block this one as I feel it needs more refinement mostly regarding usability for our addons.

ALongStringOfNumbers (approved):

PR #1576, opened 2021-04-19
Fix Tooltips in JEI Recipes

serenibyss:
**What:**
This PR fixes a very minor visual issue in JEI where, if a recipe has a not consumed input, and that same ingredient as an output, the output will also have the "notConsumed" tooltip applied to it. This could also occur if a chanced output was also an input.

**How solved:**
Added checks to make sure chanced and not consumed tooltips are only applied to the correct side ingredient side.

**Outcome:**
- Fix not consumed and chance tooltips on wrong ingredients in JEI

None expected.

### DIFF
--- a/src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
+++ b/src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java
@@ -119,7 +119,7 @@ public class GTRecipeWrapper implements IRecipeWrapper {
             throw new IllegalArgumentException("Unknown ingredient type: " + ingredient.getClass());
         }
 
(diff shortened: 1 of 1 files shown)


TechLord22 (approved):

LAGIdiot (approved):
Thanks for spotting this issue and fixing it.

For this type of fixes pictures should be provided in opening comment and template mentions it! But you don't have do it for this one.

DStrand1:
My bad. I had actually taken a photo and forgot to put it:
image

LAGIdiot:
Picture before is alone good only for issue. Here would be better to show before and after pictures.

PR #1577, opened 2021-04-19
Updated Changelog and version bump to 1.14.1

LAGIdiot:


LAGIdiot:
Currently there are problems with Maven and resolving dependencies so release 1.14.1 is being postponed

LAGIdiot:
Currently there are problems with Maven and resolving dependencies so release 1.14.1 is being postponed

Issue #1342, opened 2020-12-10
[Suggestion] Coil tier indication in JEI


Ar4093:
**Is your feature request related to a problem? Please describe.**
For Blast Furnace recipes, it would be nice to quickly see what coil tier (at the minimum) is needed

**Describe the solution you'd like**
Similar to the power tier indicator, add the coil tier after the temperature, e.g. **Temperature: 2900K (Nichrome)**

LAGIdiot:
Thank you for suggesting this. I think it is cool idea which would help other players.

Rendering of this is currently handled here: https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/integration/jei/recipe/GTRecipeWrapper.java#L118-L121 . And it supports only showing of string values from recipe.

To support more than that it would require full reimplementation. But we could make it more versatile so it supports any similar concept as coils for temperature. There is probably not much more usage for GTCE alone right now but addons could make killing on this.

PR #1580, opened 2021-04-22
Rework of Recipe Properties

LAGIdiot:
**What:**
This PR is full rework of Recipe Properties. Now they provide information how to format and render their value. They are able to hold additional information that can be shown in JEI. And usage should be easier to use while being strongly typed.

Blast furnace recipes now shows minimal Coil tier needed for them.

This is based on #1371 

**How solved:**
Recipe property system now consists of 2 parts. First part is  `RecipeProperty` which works as template. Here is defined how it should be formatted for rendering purposes or what should be shown. Second part is value which does this property render. System is strongly typed and allows usage of default values if needed. It provides level of validation.

`Recipe` now has `recipePropertyStorage()` which will handle all Recipe Property related things.

Old way is now marked as deprecated.

**Outcome:**
Closes: #1342 
Supersedes: #1371 

**Additional info:**
Coils

This will created technical debt in `AmplifierRecipeBuilder` and `FusionRecipeBuilder` as they are still using old system but we can't switch them to new one as they currently don't have any (useful) usage.

Credits:
@detav for first iteration of this system
@ALongStringOfNumbers for refreshing my memory how it works


ALongStringOfNumbers:
Even though GTCE currently does not implement anything for the Fusion Reactor or the Amplifier, maybe they should still be transferred to the new system, in minimal implementation form, so that addon mods (Which I know implement the Fusion Reactor, not sure about the Amplifier) have a simple, single place where they could implement these features.

In my opinion, it would be cleaner for addons to override various things in a FusionReactorProperty instead of in FusionRecipeBuilder.

LAGIdiot:
Yeah that was my point too even if I implemented in new way I would not know if it works or not as there are no recipes for it.

DStrand1 (approved):
This looks well implemented to me. I wonder if there is any other machines we have that could take advantage of this though. Maybe it is time we give some coil functionality to the Pyrolyse Oven or Cracker Unit, though obviously that is not part of this PR, just food for thought.

Additionally, I think this shouldn't be any trouble for addons to take advantage of, and shouldn't break current builds or anything either

Issue #1089, opened 2020-05-29
[Bug] Full 3-Dimensional Axis Viewing And Zoom Support In JEI's Multi-block Pattern Category / Text Overlap


DarkAmbientNoise:
I love that we can now see how to build the multi-block structures through JEI now, as I don't remember that being a feature around 2 years ago. However for some machines, it can be confusing, especially for newer users of the mod to know exactly how to build the machines.

Problem: The view-field that we get tends to be too zoomed in and only supports the X and Y axis for rotation, which is nice, but makes it impossible to see internal blocks of the structure for something like the Fusion Reaction Computers. Rotating around this block is also a little confusing to find where you left off at.

![FRC Full](https://user-images.githubusercontent.com/12950685/83263185-d1bf8a00-a183-11ea-81e9-2d7604601be8.png)

Layering the multi-block can help on some of these structures, but only where it is a taller one as it forces the rendered layer to have a somewhat top down view, closer to isometric rather than on one direct face, but on others like this one, it doesn't.

![FRC Layer 2](https://user-images.githubusercontent.com/12950685/83263310-003d6500-a184-11ea-9cdf-cc48912f37e1.png)

Expected Feature: Able to see a full 3-D version of the model, accessing the Z axis simply by clicking and dragging up or down. Zooming out with the scroll wheel while the mouse is in the 3-D view-port would severely help for larger structures along side the Z axis improvement. Removing text overlapping would also be convenient if possible.

Note: I checked the closed issues and there was one that was mentioning a few things about this and was primarily about missing the rendering of specific blocks, though it didn't seem to have much info aside from it being an issue with newer versions. (Issue #945). This is clearly a back-burner bug fix, as a stable mod is more important, but it would be nice.

ALongStringOfNumbers:
Everything in this issue should now be done as of #1583. Please feel free to open a new issue if this is not the case.

PR #1583, opened 2021-04-25
Transfer multiblock descriptions to an info tab

ALongStringOfNumbers:
**What:**
This PR transfers the overlapping multiblock description on the JEI display page into a specific JEI info page, so that the description does not make it harder to see the structure.

In addition, information has been added about the various ways that the multiblock preview can be manipulated.

Also, this should be the final remaining point from #1089, so that might be able to be closed after.

**How solved:**
Removes the description drawing from the `drawText` method in `MultiblockInfoRecipeWrapper` and moves it to `GTJEIPlugin`, making an accessor for `InfoPage` in the process.

**Outcome:**
Moves multiblock descriptions to a JEI info page and adds an informational hover.

**Additional info:**
Current:
javaw_2021-04-25_12-17-44

In PR:
java_2021-05-22_13-00-01

Hover:
java_2021-05-22_13-54-28

java_2021-04-22_00-04-31

Also still works for localization:
java_2021-04-22_00-14-38

Addons that don't already override MultiblockInfoPage#getDescription will need to do so now.


warjort:
````
 GTUtility.drawCenteredSizedText(recipeWidth / 2, -FONT_HEIGHT, localizedName, 0x333333, 1.3);
````

Its not a good idea to use these negative coords. This is what it looks like for me with "auto" gui scale.

**Comment by warjort** [comment]
Maybe you could add a dummy "I" (for information) widget next to the L:A button that has the mouse controls as a tooltip?

**Comment by LAGIdiot** [comment]
I think it would be better to add some widget which will on hover show these preview manipulation choices instead putting them up there.

**Comment by ALongStringOfNumbers** [comment]
I am working on a widget/button that will be able to display the information.

**Comment by ALongStringOfNumbers** [comment]
With my most recent commit I have removed the information from the top, which was put there by scaling in the negative y, and moved instead to an informational hover icon, which will display information in a tooltip when hovering over the icon. I have updated my original post with new images

**Comment by ALongStringOfNumbers** [comment]
As a note, I fixed the somewhat dim look on the information hover, as seen here:
java_2021-05-05_21-18-51

This was done by changing some calls in `WorldSceneRenderer` to avoid calling some methods that Forge had marked as do not use, because they did not accept the PR to fix the issue. However, this change made the button highlighting bleed into the information hover, as mentioned in #1452. I am looking into how to fix the button highlight bleed now.

**Comment by warjort** [comment]
> This was done by changing some calls in WorldSceneRenderer to avoid calling some methods that Forge had marked as do not use, because they did not accept the PR to fix the issue

Did you commit those changes? I don't see them.

**Comment by ALongStringOfNumbers** [comment]
> Did you commit those changes? I don't see them.

I have not committed the changes because I have not found out what is causing the button bleed that comes with the changes. The specific changes though were moving away from GLSTM.popAttributes and GLSM.pushAttributes to call GL11.glPopAttributes and GL11.glPushAttributes

**Comment by LAGIdiot** [comment]
I don't have knowledge regarding fixing rendering issues so I can't offer my help here. But if anyone has any idea we would appreciate it because I don't want to put it in with render issue.

**Comment by ALongStringOfNumbers** [comment]
Just as clarification, the button highlight bleed does not occur in the current version of the PR, it is just the somewhat dim information hover.

The highlight occurs again after I made the changes as stated above, which is also what makes the information hover not as dim. I have not committed these changes though because I did not find out what was causing the button highlight issue.

**Comment by ALongStringOfNumbers** [comment]
The last set of commits have fixed the button highlight bleeding issue that I was seeing and demonstrated above. That should finalize this PR and make it ready for review again. I would be happy for feedback, especially on my previous two commits fixing the button highlighting.

**Comment by LAGIdiot** [comment]
Maybe we mention all possible keys to manipulate that scene. Like: RMB to drag, Mouse wheel to reset. I don't see any rendering issues expect for that dim problem witch is making it unusable in middle of night. And we need definitely to fix it.

Strange thing is it somehow follow day/night cycle and it is not correctly shown on screenshots.

**Comment by LAGIdiot** [comment]
I did some quick testing and these two calls to `minecraft.entityRenderer.enableLightmap()` is what **F**s whole GUI. When I removed them GUI behaved correctly, except for info button that was dim and highlighting problem on button hover. But that one can be mitigated by rendering them last as @warjort found out.

**Comment by LAGIdiot** [comment]
Eureka! Info button is not dim anymore

File: src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoCategory.java [hunk 1]
```diff
@@ -84,4 +88,10 @@ public void setRecipe(IRecipeLayout recipeLayout, MultiblockInfoRecipeWrapper re
         recipeWrapper.setRecipeLayout((RecipeLayout) recipeLayout, this.guiHelper);
     }
 
+    @Override
+    public void drawExtras(Minecraft minecraft) {
+        for (int i = 0; i < 20; ++i) {
+            this.slot.draw(minecraft, 18*i-180*(i/10), background.getHeight()-36+18*(i/10));
+        }
+    }
```
LAGIdiot, line 96: I don't think this should be extracted here. As it does not bring us any advantage. Also if we need to change number of rendered slots we would need to do it in two files

LAGIdiot:
I have made PR to fix rendering problems: https://github.com/ALongStringOfNumbers/GregTech/pull/3

LAGIdiot:
This needs to be released as minor because we will stop showing description for our addons till they put it in info page.

ALongStringOfNumbers:
I have pulled in the PR to fix the rendering dimming and to cleanup minor things, and have updated the first post to display everything at the current state of the PR.

src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoCategory.java
ALongStringOfNumbers: This was fixed when the rendering fix/cleanup was merged in

ALongStringOfNumbers:
With that last commit, I added a bit more information to the hovering text box, as mentioned in a previous comment, and have updated the image in the original post

LAGIdiot:
Thanks for your hard work!

src/main/java/gregtech/integration/jei/GTJeiPlugin.java
```diff
@@ -197,5 +199,10 @@ public void register(IModRegistry registry) {
             registry.addIngredientInfo(machine.getStackForm(), VanillaTypes.ITEM,
                 "gregtech.machine.fluid_canner.jei_description");
         }
+
```
Exaxxion, line 202: Probably more succinct to do something like this here:
```java
MultiblockInfoCategory.multiblockRecipes.values().forEach(v -> {
    TypeGoesHere infoPage = v.getInfoPage();
    registry.addIngredientInfo(infoPage.getController().getStackForm(),
                               VanillaTypes.ITEM,
                               infoPage.getDescription();
});
```

src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoCategory.java
```diff
@@ -26,8 +26,7 @@
     public MultiblockInfoCategory(IJeiHelpers helpers) {
         this.guiHelper = helpers.getGuiHelper();
         this.background = this.guiHelper.createBlankDrawable(176, 166);
-        ResourceLocation iconLocation = new ResourceLocation(GTValues.MODID, "textures/gui/icon/coke_oven.png");
-        this.icon = this.guiHelper.createDrawable(iconLocation, 0, 0, 16, 16, 16, 16);
+        this.icon = guiHelper.drawableBuilder(GuiTextures.MULTIBLOCK_CATEGORY.imageLocation, 0,0,18,18).setTextureSize(18,18).build();
```
Exaxxion, line 29: It looks like the coding convention is to put space between lists of arguments, so you might want to adjust that here.

src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoPage.java
```diff
@@ -15,4 +17,15 @@
     public float getDefaultZoom() {
         return 1.0f;
     }
+
+    public List<String> informationText() {
+
+        return new ArrayList<String>() {{
```
Exaxxion, line 23: This generates an anonymous subclass of `ArrayList<String>` with statically-initialized members. Perhaps it would make more sense to use something like:
```java
return Stream.of("gregtech.multiblock.preview.tilt", [...])
             .map(I18n::format)
             .collect(Collectors.toList());
```

src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoRecipeWrapper.java
```diff
@@ -162,6 +175,13 @@ private void switchRenderPage(int amount) {
         }
     }
 
+    private void preparePlaceForParts(int recipeHeight) {
+        IGuiItemStackGroup itemStackGroup = recipeLayout.getItemStacks();
+
+        for (int i = 0; i < MAX_PARTS; ++i)
+            itemStackGroup.init(i, true, 18 * i - 180 * (i / 10), recipeHeight - PARTS_HEIGHT + 18 * (i / 10));
```
Exaxxion, line 182: There's quite a few magic numbers here. Maybe declare some constants so the reader can understand what is being computed here and why.

src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoRecipeWrapper.java
```diff
@@ -210,24 +230,32 @@ public void preRenderScene(WorldSceneRenderer renderer) {
...
+        drawMultiblockName(recipeWidth);
+
+        //reset colors (so any elements render after this point are not dark)
+        GlStateManager.color(1.0F, 1.0F, 1.0F, 1.0F);
+
+        this.infoIcon.draw(minecraft, recipeWidth - 25, 49);
+
         for (int i = 0; i < MAX_PARTS; ++i) {
-            this.slot.draw(minecraft, 18*i-180*(i/10), recipeHeight-36+18*(i/10));
+            this.slot.draw(minecraft, 18 * i - 180 * (i / 10), sceneHeight + 18 * (i / 10));
```
Exaxxion, line 244: More magic numbers

src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoRecipeWrapper.java
```diff
@@ -266,20 +294,21 @@ public void drawInfo(Minecraft minecraft, int recipeWidth, int recipeHeight, int
         this.lastMouseY = mouseY;
     }
 
-    private void drawText(Minecraft minecraft, int recipeWidth) {
+    @SideOnly(Side.CLIENT)
+    protected void drawHoveringInformationText(Minecraft minecraft, List<String> tooltip, int mouseX, int mouseY) {
+        int minX = recipeLayout.getRecipeCategory().getBackground().getWidth();
+        int[] yRange = new int[]{49, 69};
```
Exaxxion, line 300: More magic numbers type stuff going on here, and generally it's better to replace a complicated conditional expression with a function call so the intent of the expression is communicated. It's easy for logic errors to crop up in places like that because there's so much noise.

src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoRecipeWrapper.java
warjort: There's really only 2 magic numbers missing, one is a standard, but I would suggest something like:

````
private static final int MAX_PARTS = 20;
+private static final int NO_PARTS_ROWS = 2; // 2 rows of parts
+private static final int PARTS_PER_ROW = MAX_PARTS / NO_PARTS_ROWS; // This is the 10
+private static final int SLOT_DIMENSIONS = 18; // This is the standard width/height of an item slot
-private static final int PARTS_HEIGHT = 36;
+private static final int PARTS_HEIGHT = SLOT_DIMENSIONS * NO_PARTS_ROWS; // This is the 36
+private static final int PARTS_WIDTH = SLOT_DIMENSIONS * PARTS_PER_ROW; // This is the 180
````

I am not sure it makes it anymore transparent that it is drawing a grid of item slots?
````
itemStackGroup.init(i, true, SLOT_DIMENSIONS * i - PARTS_WIDTH * (i / PARTS_PER_ROW), recipeHeight - PARTS_HEIGHT + SLOT_DIMENSIONS * (i / PARTS_PER_ROW));
````

You need to recognise the trick that (i / PARTS_PER_ROW) returns 0 for 0 to 9, 1 for 10 to 19, etc.

Maybe the following is clearer?
````
int row = i / PARTS_PER_ROW;
itemStackGroup.init(i, true, SLOT_DIMENSIONS * i - PARTS_WIDTH * row, recipeHeight - PARTS_HEIGHT + SLOT_DIMENSIONS * row);
````

Exaxxion (approved):
Code looks good to me with the latest revisions. Testing in game I'm not seeing any graphical quirks like highlight bleed that were previously occurring. The description is showing up on the JEI info category for uses of the multiblock controller like one would expect for each multiblock.

It certainly seems usable in its current state, though if possible it would be great if we could eventually figure out how to not have that large empty gap at the top of the category page.

As an aside, I think there are some typos or outdated information in the description strings of some of these multiblocks but those are outside the purview of this PR.

warjort:
> though if possible it would be great if we could eventually figure out how to not have that large empty gap at the top of the category page.

I had a look at it before and concluded it would be pretty hacky. 

JEI does a calculation based on the screen height (depends on gui scale) and a config parameter called maxRecipeGuiHeight.
And then it basically pads the top and bottom so your recipe layout is centred within that common height for all recipes.

If you read the config parameter from JEI and got the screen height from minecraft you could repeat the calculation and then choose a background height for your recipe layout that fills it.

But the gui scale and JEI parameter can be changed at runtime, breaking your calculation.

LAGIdiot:
As discussed I will merge this to master now so Addons have more time to adjust changes, later this will be part of next release which will be Minor.

PR #1584, opened 2021-04-26
Add clarification tooltips to blocks in the multiblock preview

ALongStringOfNumbers:
**What:**
Adds a clarification tooltip to item buses and fluid hatches in the multiblock preview and the component list specifying that the multiblock structure can be formed with any tier bus/hatch.

**How solved:**
Added a new default tooltip to `MultiblockInfoPage`. It was done this way for addon mods, which implement the assembly line that only uses ULV tier buses.

The implementation of the tooltips themselves was tacked onto the existing tooltips for the preview, and a new method was created for the components list

**Outcome:**
Adds a clarification tooltip to buses/hatches in the multiblock preview

**Additional info:**
java_2021-04-25_23-10-11

java_2021-04-25_23-41-45

None expected


warjort:
This type of code doesn't look correct to me.
The hatches and buses and not necessarily implementations from gregtech/common
````
          MetaTileEntity mte = GregTechAPI.META_TILE_ENTITY_REGISTRY.getObjectById(tooltipBlockStack.getItemDamage());
            if(mte instanceof MetaTileEntityItemBus || mte instanceof MetaTileEntityFluidHatch) {
                tooltip.add(1, TextFormatting.WHITE + I18n.format(infoPage.getBusTooltip()));
            }
````
You should instead be checking something like:
````
if (mte instanceof IMultiBlockAbilityPart) {
    MultiBlockAbility ability = ((IMultiBlockAbilityPart) mte)).getAbility();
    if (ability == MultiBlockAbility.EXPORT_ITEMS || etc.) 
}
````

**Comment by warjort** [comment]
I don't want to introduce feature creep, but it would be good if you implement this in a way thinks about a possible future PR 
that introduces tool tips from the real structure info/recipe map, 

e.g. how many hatches/buses you really need.
e.g.2 the minimum casings

pseudo code:
````
MultiblockAbility ability = ...;
if (controller instance of RecipeMultiBlockController)
   addMinTooltip(controller.recipeMap, ability));
   addMaxTooltip(controller.recipeMap, ability);
}
````

I don't think you can actually implement it that simplistically since these recipemap constraints are actually hidden behind predicates and might not be the real constraint?

**Comment by warjort** [comment]
> I don't think you can actually implement it that simplistically since these recipemap constraints are actually hidden behind predicates and might not be the real constraint?

Just spitballing;

Absent some additional plumbing to expose the real constraints, you could hack it by adding something like
````
public MultiblockControllerBase {

    public boolean isSimpleConstraints() {
        return false; // Override on multiblocks that don't have complicated constraints
    }
}
````
This would give the JEI plugin a hint that it can bypass the full constraint system and use the simpler source constraints.

File: src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoRecipeWrapper.java [hunk 1]
```diff
@@ -307,11 +310,23 @@ public boolean handleClick(Minecraft minecraft, int mouseX, int mouseY, int mous
                     tooltip.set(k, TextFormatting.GRAY + tooltip.get(k));
                 }
             }
+            MetaTileEntity mte = GregTechAPI.META_TILE_ENTITY_REGISTRY.getObjectById(tooltipBlockStack.getItemDamage());
+            if(mte instanceof MetaTileEntityItemBus || mte instanceof MetaTileEntityFluidHatch) {
```
Archengius, line 314: It shouldn't be really checking for hatchets or buses, better check multiblock capability type instead

ALongStringOfNumbers:
I have changed this PR to instead accept a Map<MultiblockAbility, List<String>>` so that tooltips can instead be added to any `MultiblockAbilityPart`, and in addition, multiple tooltips can be added to the same part.

This is an example of multiple tooltips being added to the same part:
java_2021-04-30_15-41-35

The "Invalid Structure" in this image was just a test of the multiple tooltips.

src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoPage.java
```diff
@@ -15,4 +16,14 @@
     public float getDefaultZoom() {
         return 1.0f;
     }
+
+    public Map<MultiblockAbility, List<String>> getAbilityTooltipMap() {
+        Map<MultiblockAbility, List<String>> abilityTooltipMap = new HashMap<>();
+        abilityTooltipMap.put(MultiblockAbility.EXPORT_ITEMS, Collections.singletonList("gregtech.multiblock.preview.any_hatch"));
```
warjort, line 22: I wouldn't use Collections.singletonList here.

You would expect this code to work
````
@Override
public Map<MultiblockAbility, List<String>> getAbilityTooltipMap() {
    Map<MultiblockAbility, List<String>> result = super.getAbilityTooltipMap();
    List<String> exportItems = result.putIfAbsent(MultiBlockAbility.EXPORT_ITEMS, Lists.newArrayList());
    exportItems.add("my.tooltip");  // <--- Doesn't work?!
    return result;
}
````

It will actually throw an UnsupportedOperationException because that singleton list is immutable.

**Comment by warjort** [comment]
I also think that tooltip map could be cached in the recipe setup. It does not need to be dynamic.

I believe the way you are doing it, it will be creating that map on every rendering tick?
Probably not super important though.

File: src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoPage.java [hunk 3]
```diff
@@ -19,10 +19,12 @@ public float getDefaultZoom() {
...
         Map<MultiblockAbility, List<String>> abilityTooltipMap = new HashMap<>();
-        abilityTooltipMap.put(MultiblockAbility.EXPORT_ITEMS, Collections.singletonList("gregtech.multiblock.preview.any_hatch"));
-        abilityTooltipMap.put(MultiblockAbility.IMPORT_ITEMS, Collections.singletonList("gregtech.multiblock.preview.any_hatch"));
-        abilityTooltipMap.put(MultiblockAbility.EXPORT_FLUIDS, Collections.singletonList("gregtech.multiblock.preview.any_hatch"));
-        abilityTooltipMap.put(MultiblockAbility.IMPORT_FLUIDS, Collections.singletonList("gregtech.multiblock.preview.any_hatch"));
+        List<String> tooltipList = new ArrayList<>();
+        tooltipList.add("gregtech.multiblock.preview.any_hatch");
+        abilityTooltipMap.put(MultiblockAbility.EXPORT_ITEMS, tooltipList);
+        abilityTooltipMap.put(MultiblockAbility.IMPORT_ITEMS, tooltipList);
+        abilityTooltipMap.put(MultiblockAbility.EXPORT_FLUIDS, tooltipList);
```
warjort, line 26: You are sharing the same list between the abilities?

So my previous attempt to add a tooltip to the export items would also be added to the others.

How about implementing it this way?

````
// Cached version
private Map<MultiblockAbility, List<String>> abilityTooltips;

public Map<MultiblockAbility, List<String>> getAbilityTooltipMap() {
    // Generate on first use
    if (this.abilityTooltips == null) {
        this.abilityTooltips = Maps.newHashMap();
        generateAbilityTooltips();
    }
    return this.abilityToolTips;
}

// Helper method
public void addAbilityTooltip(MulitblockAbility ability, String tooltip) {
    List<String> tooltips = this.abilityToolTips.putIfAbsent(ability, Lists.newArrayList());
    tooltips.add(tooltip);
}

// The default implementation
public void generateAbilityTooltips() {
    addAbilityToolTip(MultiblockAbility.EXPORT_ITEMS, "gregtech.multiblock.preview.any_hatch");
    addAbilityToolTip(MultiblockAbility.IMPOTY_ITEMS, "gregtech.multiblock.preview.any_hatch");
    addAbilityToolTip(MultiblockAbility.EXPORT_FLUIDS, "gregtech.multiblock.preview.any_hatch");
    addAbilityToolTip(MultiblockAbility.IMPORT_FLUIDS, "gregtech.multiblock.preview.any_hatch");
}
````

This is easier to use (harder to get wrong) and would make my subclassing example less verbose
````
@Override
public void generateAbilityTooltips() {
    super.generateAbilityTooltips();
    addAbilityToolTip(MultiblockAbility.EXPORT_ITEMS, "my.tooltip");
}
````

It also adds caching on first use.

**Comment by LAGIdiot** [comment]
This looks like good idea but I would like to open here discussion how we can make this to show minimal tier for tier dependent MultiblockParts like rotors, energy outputs and so on. As that would significantly enchant this.

File: src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoPage.java (hunk 3 above)
Comment (galyfray) [reply src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoPage.java:26]: There is still a problem with the way you are proposing warjot : you can't override existing tooltip without rewriting the main par of  `generateAbilityTooltips()`
I think a `removeTooltip(MulitblockAbility ability, String tooltip)` method might be useful.

File: src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoRecipeWrapper.java [hunk 4]
```diff
@@ -307,11 +310,36 @@ public boolean handleClick(Minecraft minecraft, int mouseX, int mouseY, int mous
...
 
+    public void addAbilityPartTooltip(int slotIndex, boolean input, ItemStack itemStack, List<String> tooltip) {
+        MetaTileEntity mte = GregTechAPI.META_TILE_ENTITY_REGISTRY.getObjectById(itemStack.getItemDamage());
+        if(mte instanceof IMultiblockAbilityPart) {
+            MultiblockAbility ability = ((IMultiblockAbilityPart) mte).getAbility();
+            Map<MultiblockAbility, List<String>> abilityTooltipMap = infoPage.getAbilityTooltipMap();
+            if(abilityTooltipMap.containsKey(ability)) {
+                List<String> tooltips = abilityTooltipMap.get(ability);
+                for(int i = 0; i < tooltips.size(); i++) {
+                    tooltip.add(i + 1, TextFormatting.GREEN + I18n.format(tooltips.get(i)));
```
galyfray, line 337: All tooltips added this way will be green tooltips it could be interesting to allow other colors and formating or at least not enforce green color.

src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoPage.java
warjort: I don't think that remove is very useful? 
You can remove tooltips you know about, but later additions to the default processing that you don't know about  at development time might still be wrong.

If you really are overridding behaviour such that default tooltips are wrong, you probably shouldn't be calling super.

warjort:
> but I would like to open here discussion how we can make this to show minimal tier for tier dependent MultiblockParts like rotors, energy outputs and so on

The way I see it, is that each JEI MB definition can override the generateToolTips() to add tips that are specific to its usage.

In future, if the opaque predicate problem is solved, these tool tips could instead be generated from the real MB pattern and recipe map automatically instead of having to define them in the JEI pattern.

src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoPage.java
galyfray: well i think the multiblock which require a minimal or very specific tier for some hatches may want to use super for other hatches or in case the default tooltips added changes. I think the steam multiblock from gregicality could be an exemple.

LAGIdiot:
I really like @warjort suggestion regarding implementation and I think that is the way to go forth on this. With possibility to override (and possible not call at all) default values we could make this work for anyone.
@ALongStringOfNumbers can you please try implementing that suggestion?

ALongStringOfNumbers:
I have done the refactoring of how the tooltips were created and stored as suggested above. However, now I am thinking that I should store tooltips for `ItemStack`s instead of `MultiblockAbility`s. This is so that tooltips can be added to non-multiblock parts to more easily declare part limitations for the multiblocks.

For example, when using `MultiblockAbility`s, tooltips would not be able to be applied to the Heat proof casings on the EBF, which would be something that we want to do to convey the minimum part requirement of 10 for the casings. 

Switching to Itemstacks would also make it easier for tooltips to be applied to more blocks in custom multiblocks created by MultiblockTweaker, as blocks in those structures have less of a chance of being `MultiblockAbility`s.

I am putting this comment here for some feedback.

ALongStringOfNumbers:
Does anyone know if there is a way to apply `TextFormatting` to `TextComponentTranslatable` without having the `TextFormatting` codes in the lang entry? I wanted to switch the `Tuple<String, TextFormatting>` to a `TextComponentTranslatable` to take advantage of the substitutions in the lang file entry (So that I could create a universal tooltip for minimum casing limit, instead of having one per multiblock), however, I was not able to find a way to apply formatting when the formatting is not in the lang entry in format codes.

Switching the formatting to being in lang codes instead of in the Tuple is not that big of a deal, I just wanted to see if anyone knew how to apply formatting outside of the localization entry.

Archengius:
You need to use setStyle and then apply color inside of the Style object

warjort:
> Does anyone know if there is a way to apply TextFormatting to TextComponentTranslatable without having the TextFormatting codes in the lang entry?

Do you mean?
````
new TextComponentTranslatable("translation.key").getStyle().setColor(TextFormatting.GREEN);
````

Using ITextComponent instead of that Tuple looks like a much better api.

File: src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoPage.java [hunk 5]
```diff
@@ -1,14 +1,26 @@
...
-import java.util.List;
+import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public abstract class MultiblockInfoPage {
 
+    private final Hash.Strategy<ItemStack> strategy = ItemStackHashStrategy.comparingAllButCount();
+
+    private Map<ItemStack, List<Tuple<String, TextFormatting>>> abilityTooltips = new Object2ObjectOpenCustomHashMap<>(strategy);
```
warjort, line 22: This is no longer an ability keyed tooltip.
You need to update the name and some javadocs.

src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoPage.java
```diff
@@ -26,4 +38,51 @@ public float getDefaultZoom() {
...
+        return this.abilityTooltips;
+    }
+
+    /**
+     * Contains the default tooltips that will be applied to all MultiblockAbilities
+     */
+    protected void generateAbilityTooltips() {
+
+        for(int i  = 0; i < GTValues.V.length; i++) {
+            addAbilityTooltip(MetaTileEntities.ITEM_EXPORT_BUS[i].getStackForm(), new Tuple<>("gregtech.multiblock.preview.any_hatch", TextFormatting.GREEN));
```
warjort, line 60: These tuple objects (or ITextComponent if you change it) are immutable.
No need to create multiple copies, use a static field.

src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoPage.java
```diff
@@ -26,4 +38,51 @@ public float getDefaultZoom() {
...
+
+    /**
+     * A Helper method for adding tooltips to MultiblockAbilities. Can be called if {@link MultiblockInfoPage#generateAbilityTooltips()} is not overridden
+     * Will add tooltips to MultiblockAbilities with existing tooltips
+     * @param itemStack - The MultiblockAbility to add a Tooltip too
+     * @param tooltip - The tooltip String and TextFormatting to be added
+     */
+    protected void addAbilityTooltip(ItemStack itemStack, Tuple<String, TextFormatting> tooltip) {
+
+        List<Tuple<String, TextFormatting>> tooltipList = this.abilityTooltips.getOrDefault(itemStack, null);
```
warjort, line 75: Why default "null" ? Looks weird :-)

src/main/java/gregtech/integration/jei/multiblock/MultiblockInfoPage.java
```diff
@@ -26,4 +38,51 @@ public float getDefaultZoom() {
...
                 .collect(Collectors.toList());
     }
+
+    /**
+     * Gets the Map containing all tooltips for MultiblockAbilities. Will generate the map if it does not exist
+     * @return - The Map containing Tooltips and Formatting for specific MultiblockAbilities
+     */
+    public Map<ItemStack, List<Tuple<String, TextFormatting>>> getAbilityTooltipMap() {
+        // Generate on first use
+        if (this.abilityTooltips.size() == 0) {
```
warjort, line 48: isEmpty() is usually better - but since you are using an array list it won't make much difference here.

e.g. somebody could change it to use a LinkedList or something where size() is not a trivial operation

warjort: Actually, this is an Object2ObjectOpenCustomHashMap which ironically does:
````
	public boolean isEmpty() {
		return size() == 0;
	}
````

ALongStringOfNumbers:
I have switched to using an ITextComponent for the second map parameter and right now am working on some default tooltips for multiblocks in GTCE.

ALongStringOfNumbers:
I have now added some tooltips for GTCE multiblocks, beyond the default tooltips saying that any tier hatch/bus can be used.

Some examples are:
java_knx7kLTW6u

java_jikpEZNYEN

As well as tooltips for the minimum number of casings required to form the structure:
java_AX8QqovUro

As this information is not currently obtainable from the Pattern Builder, the limits are set manually for each structure.

DStrand1 (approved):

PR #1586, opened 2021-05-02
Fix EV charger not having a recipe

ALongStringOfNumbers:
**What:**
This PR fixes the EV charger not having a recipe in game.

**How solved:**
Replaces the designated battery, which did not work before, with an actual battery, choosing the EV tier battery

**Outcome:**
Fix EV tier charger not having a recipe

**Additional info:**
Working recipe:
java_2021-05-02_00-50-37

None expected


Exaxxion (approved):
Based on this file history this appears to have been broken since the upstream merge of GregicAdditions functionality in 2019. Confirmed that the recipe is missing without these changes, and that it is working again with them.

Code looks fine, the EV charger recipe was an outlier trying to use a nonexisting oreDict for the battery when the rest of them specify precisely which battery to use. This change brings it in line with the other entries to use the Lapotron Crystal specifically.

LAGIdiot (approved):
Looks reasonable, thanks for spotting and fixing this.

PR #1587, opened 2021-05-02
Fix Steel Steam furnace not using steel pipes

ALongStringOfNumbers:
**What:**
Fixes the Steel tier Steam furnace not using Steel Pipes in its recipe like all other Steel tier Steam machines

**How solved:**
Changed bronze to steel.

**Outcome:**
Changes the Steel tier Steam furnace to use steel pipes

**Additional info:**
Before:
java_2021-05-02_01-11-02

After:
java_2021-05-02_01-05-02

None expected for mods, modpacks might have to minorly update a CT script


LAGIdiot (approved):
This makes sense to me.

PR #1590, opened 2021-05-07
Do not spawn rubber trees in water, looks ugly

alfuken:
**What:**
Prevents rubber trees from spawning in water. Looks fugly otherwise: (see screenshot below)

**How solved:**
Prevent spawning by checking if the above block is a liquid. We wouldn't want them to spawn in lava or oil, too.

**Outcome:**
Changed rubber tree spawning to not spawn in water (or any other liquid)

**Additional info:**

By the way, the so called `randomPos` is used in `WorldGenRubberTree` is not, well, really random... but that's another story.

I don't see any.


warjort:
Interesting, I never realised the rubber trees were actually generated on a grid.

Your fix looks like it works, but I have raised a new bug report #1591 what I think is the real problem.

LAGIdiot (approved):
This makes sense to me. Thanks for fixing this.

LAGIdiot (approved):

PR #1594, opened 2021-05-10
Remove Forge Multipart legacy support

LAGIdiot:
**What:**
I have removed Forge Multipart from dependencies and all code which was using it. GTCE is not supporting Forge Multipart at least from 2019-08. There was only mechanism for legacy block conversion. But given how long this code was here all legacy worlds were already converted or owners are not playing anymore.

**Outcome:**
Forge Multipart is no longer dependency which means quicker and easier builds.
No legacy block conversion will be done.

**Additional info:**
I strongly believe (and Arch too) that we don't need this anymore. Also it prevents one possible vector of problems which we were facing recently. And given how little this bring us I don't see reason to keep it.

It won't be possible to convert old worlds.


warjort:
It looks to me that all the code you removed could be reconstituted into an addon mod if somebody really needs to convert an old world?

Archengius:
Why can't they just install older version which supports conversion before though?

DStrand1 (approved):
Tested this myself and works just fine without FMP in the dev environment. I think we should merge this soon since we need to fetch these commits to get the environment to work currently

PR #1595, opened 2021-05-10
Using MetaItems using FoodStats with a Non-Null containerItem now return that ItemStack to the Player

bruberu:
**What:**
Previously, although containerItem was one of the fields within FoodStats, it was not used in any way, thus causing the player to not receive the ItemStack once it was eaten. This PR intends to fix this issue.

**How solved:**
After looking through vanilla ItemPotion code, I found the suitable portion of logic required to correctly return the container to the player, which indicated that the resulting stack would need to be passed through onItemUseFinish, which unfortunately made it a requirement that onEaten would need to return an ItemStack, thus causing a minor API change.

**Outcome:**
Using FoodStats with Non-Null containerItems return the container to the player.
Most notably, this now means that Purple Drink will now return a bottle after being consumed.
I do not believe there are any previous issues mentioning this problem previously, nor could I find any.

**Additional info:**
https://user-images.githubusercontent.com/80226372/117707562-5092b480-b194-11eb-81d6-7739e98503dd.mp4
This showcases this PR working upon Purple Drink.

Any addons that do implement their own onEaten methods will require them to return an ItemStack now, although I do not believe there are any that do so, considering that the method is implemented already through the common implementation of IFoodBehavior.

### DIFF
--- a/src/main/java/gregtech/api/items/metaitem/FoodStats.java
+++ b/src/main/java/gregtech/api/items/metaitem/FoodStats.java
@@ -67,22 +67,32 @@ public class FoodStats implements IFoodBehavior {
     }
 
     @Override
-    public void onEaten(ItemStack itemStack, EntityPlayer player) {
+    public ItemStack onEaten(ItemStack itemStack, EntityPlayer player) {
         if (!player.world.isRemote) {
             for (RandomPotionEffect potionEffect : potionEffects) {
                 if (Math.random() * 100 > potionEffect.chance) {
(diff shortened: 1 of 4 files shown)


bruberu:
I just implemented those changes with the most recent commit. I should note that `player.world.isRemote() ? null : stack` should instead just return `stack`, as the original function outputs that variable regardless of side as long as it doesn't return something else. As such, with this commit, those compatibility issues should now be resolved.

PR #1596, opened 2021-05-11
Fix Issues with Dust Uncrafting

serenibyss:
**What:**
GTCE registers 2x2 recipes for Tiny and Small Dusts so that the player can do these recipes in their internal inventory without needing a Crafting Table. GTCE registers:
- `(Dust -> 9 Tiny Dust)` in the top left of the 2x2
- `(Dust -> 4 Small Dust)` in the bottom right of the 2x2

Now, when Minecraft tries to expand this out for a 3x3, it will cause weird issues making JEI display a wrong recipe for a 3x3.
image
On top of this, SoG/Gregicality tries to fix the problem by removing GTCE's `(Dust -> 4 Small Dust)` recipe and register its own in the top right of the 2x2. This only partially fixes the problem, however.

**How solved:**
Minecraft will search for a Crafting Table recipe in a first-encounter basis. Because these recipes are handled through the `OrePrefix` Material Handlers, the recipes are registered in the order of the `OrePrefixes` themselves. This leads to Tiny Dust recipes being done before Small Dust, taking priority and overriding the recipe that JEI shows.

This PR changes the order of the `OrePrefixes` to resolve this. I am not totally sure of the implications this has, but from my (admittedly brief) testing, this does not display any issues and also resolves the JEI issue.

**Outcome:**
- Fix Small Dust recipe showing wrong in JEI

After images:
JEI:
image

Crafting Table:
image
image

**Additional info:**
Additionally, I changed our Small Dust recipe to be the top right of the 2x2 rather than bottom right. I did this because both major addons are doing this, so I think it makes sense to just pull this change up into GTCE to save them the hassle of removing and re-registering recipes. Considering GTCE is almost always used with one of these two major addons, this should not be particularly disruptive. However, if anyone disagrees that we should make this change, I will happily reverse the recipe change.

None expected

### DIFF
--- a/src/main/java/gregtech/api/unification/ore/OrePrefix.java
+++ b/src/main/java/gregtech/api/unification/ore/OrePrefix.java
@@ -62,8 +62,8 @@ public enum OrePrefix {
     gemFlawless("Flawless Gemstones", M * 2, null, MaterialIconType.gemFlawless, ENABLE_UNIFICATION, mat -> mat instanceof GemMaterial), // A regular Gem worth two Dusts. Introduced by TerraFirmaCraft
     gemExquisite("Exquisite Gemstones", M * 4, null, MaterialIconType.gemExquisite, ENABLE_UNIFICATION, mat -> mat instanceof GemMaterial), // A regular Gem worth four Dusts. Introduced by TerraFirmaCraft
 
-    dustTiny("Tiny Dusts", M / 9, null, MaterialIconType.dustTiny, ENABLE_UNIFICATION | DISALLOW_RECYCLING, mat -> mat instanceof DustMaterial), // 1/9th of a Dust.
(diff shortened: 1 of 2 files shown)


Exaxxion:
I don't know that it would be appropriate for me to review this since I worked with you on it, but the code certainly looks fine to me. Very small change with a nice benefit.

I am not aware of a use case that would be tripped up by the ordinal reordering, nor do I think we have to worry about (de)serialization or other such issues that could arise from it (we're not writing a SOAP service or JavaBean type thing, after all).

I agree with the recipe change but that's a matter for the council to hash out.

warjort:
The only other use I can see for OrePrefix.oridinal/values is in the RecyclingRecipeHandler
But that does not process dust prefixes so its not going to change those recipes.

LAGIdiot (approved):
This seems reasonable to me. And I am too not aware of any ill effects of this change. Just to be sure I marked it as rsr:Minor

PR #1600, opened 2021-05-14
fix TOP showing wring burn time

galyfray:
**What:**
Just found while playing that the one prob doesn't show correct burn time remaining on steam lava boilers, it appears that the same problem occur with the coals one. 100mb of Lava produce 100 burn time but burn time is consumed every 12 ticks and not every ticks has the first person who did this code thought. How much burn time is consume depends on if the machine is high pressure.

**How solved:**
Simply add a  factor depending on if the boiler is high pressure

**Outcome:**
Fixes: Wrong remaining burn time showed by the one prob on small steam boilers

None


LAGIdiot (approved):
Thank you for spotting and fixing this issue.

Issue #1344, opened 2020-12-12
[Suggestion][Texture] Make differing input/output overlays on buses/hatches


ChromaPIE:
**Is your feature request related to a problem? Please describe.**
When building, and, of course, working with multiblock structures, it's sometimes annoying pointing at buses and hatches, taking a look at The One Probe at the corner of the screen, to finally tell one from another. Not a big problem but it's undoubtedly doing a minus to the gaming experience.

**Describe the solution you'd like**
So, I'm hoping for a texture change to solve the annoying little issue. That is to say, make four overlay textures (input/output bus/hatches), each differs a liitle bit from others, to replace the grey square overlays existing now.

LAGIdiot:
Thank you for creating this suggestion. I think it is good idea to make these blocks look distinct.

ALongStringOfNumbers:
This is done as of #1607

PR #1607, opened 2021-05-16
 Make differing input/output overlays on buses/hatches

galyfray:
**What:**
Implemented : #1344 

**How solved:**
Added small arrow head indicating if it's input or output, the arrows are in different place to ease recognize if it's in or out.
**Outcome:**
Implemented : #1344 

**Additional info:**
placed in the world : 

in the player hotbar/inventory :

None


LAGIdiot:
I expect that arrows pointing in is input and arrows pointing out is output bus right?

DStrand1 (changes requested):
Changing these textures has an unintended side effect of also changing the textures on all MetaTileEntities, so I think that these specific textures should be left as-is, and the new textures added to entirely new assets.

While you are at it, what do you think of having Item Buses use orange arrows, and Fluid Hatches using blue arrows? That way not only can you tell the difference between input and output, but also between fluid and item This is similar to the auto-output buttons in singleblock machine GUIs.

galyfray:
>Changing these textures has an unintended side effect of also changing the textures on all MetaTileEntities, so I think that these specific textures should be left as-is, and the new textures added to entirely new assets.

this change on all MetaTileEntities looks pretty nice in my opinion I don't sse any reason to go backward on this, does this change make some horrible things to some MetaTileEntities ? 

> While you are at it, what do you think of having Item Buses use orange arrows, and Fluid Hatches using blue arrows? That way not only can you tell the difference between input and output, but also between fluid and item This is similar to the auto-output buttons in singleblock machine GUIs.

good idea gonna add this

galyfray:
Some Item and fluid output overlay already exist why not edit them to get arrow add they input counterpart and leave the pipe IO as is or as before my editing ?

galyfray:
Changed existing fluid and item output overlay to use arrow and create their input  counterpart. I reverted the pipe overlay so there is no more side effect on all of the MetaTileEntities except for those who where using the existing output overlay.

the hatches are now easy to recognize between input output fluid and item. the machine is an electrolizer with auto output on fluid and item on the bottom and no auto output on the top.

Exaxxion:
> this change on all MetaTileEntities looks pretty nice in my opinion I don't sse any reason to go backward on this, does this change make some horrible things to some MetaTileEntities ?

I believe he brings this up because single-block machines would get these arrows applied when they already have their own indicators for the dedicated output face. Input arrows would also not make sense as these machines need covers (which have their own textures) for any other I/O.

It could be interesting to visually distinguish between input and output mode on a placed cover, but that seems outside of the scope of this PR targeting multiblock hatches and buses.

All that said, I do like the blue and orange arrows. I think it will be helpful. My only critique is that it's a bit subtle on the item form so it might be hard to tell on smaller GUI scales.

LAGIdiot:
Currently these textures are replacing one we are using for single block machines. Through I am not sure if it is for good or worse. I feel it is bit subtle but it would give us more space for other indicators like "Input allowed from output side".

Regarding covers, they don't interfere with texture it self so I don't think there is any problem.

galyfray:
well as the machines were already using texture named exactly how I was planing to name mine I thought it wouldn't harm to change those abstract rings to arrows tough I agree this is more subtle than the previous textures but I didn't find another good way to have both liquid and item output or input on the same block.

DStrand1:
> Currently these textures are replacing one we are using for single block machines. Through I am not sure if it is for good or worse.

Personally I think it is for worse. The machines already indicate clearly if they have auto-output specified, so I don't see any reason for them to change. But anyway, I think this PR should stay on scope and only apply this to buses and hatches, and the rest can possibly be discussed further on another issue or PR since this one has now been sitting here for two weeks with no further discussion.

If this suggested action is taken, I think the fluid and output overlays should be centered on the hatches, since they would look strange and off-center as it is on just buses/hatches which personally bothers my ocd a lot :)

galyfray:
>If this suggested action is taken, I think the fluid and output overlays should be centered on the hatches, since they would look strange and off-center as it is on just buses/hatches which personally bothers my ocd a lot :)

if using the arrow for the single block machine is not on the scope of this PR then yes the texture should and will be centered as in my first attempt. I am changing this.

galyfray:
okay so I reverted the output and input textures for the single block machines and did a little rework on the hatch arrow.
first I centered all the arrows as it isn't needed anymore to put both fluid and item overlay. I have widen a bit the arrow to make them easier to see and I keep the idea of input arrows are on top and bottom while output one are on the left and right which I hope will make them  easy to recognize with a bit of time.

they are easy to recognize when placed in the world


harder on the hot bar and in the inventory though with the habit I hope it will be easier

 


it is quite easy to recognize them in JEI which is a good thing and will save a lot of time for a lot of people.

DStrand1 (approved):
I think at this point this is good to go. I agree that it will help in inventory, and also on multiblocks in world, making it easier to see which is which when formed. There will be some casings where they blend in to the casing background, but considering the various casings addons are adding, I think this will always be a minor issue so I'm satisfied with this solution.

PR #1609, opened 2021-05-17
Update README.md

pyure:
Updating readme verbiage re. GTCE Council, per LAG's request.


LAGIdiot (approved):
Thanks for your hard work!

PR #1610, opened 2021-05-17
Updated Changelog and version bump to 1.15.0

LAGIdiot:


DStrand1:
https://github.com/GregTechCE/GregTech/blob/2ed876a782bca1993a57d09e390f625de2eb0864/src/main/java/gregtech/common/datafix/fixes/metablockid/MetaBlockIdFixHelper.java#L19

We need to make sure to update this variable to 1.15.0 before releasing this update

Exaxxion:
I would also suggest we consider adding #1485 to this release as the Crafting Station right-click voiding bug is pretty bad. @ALongStringOfNumbers recommended it for further review about a week ago.

DStrand1 (approved):

Issue #1602, opened 2021-05-15
[BUG] Multiblock output hatch sometimes don't output


galyfray:
**Describe the bug**
Sometimes for no real reason the output hatch of some multiblocks  don't output to connected pipes.

**Versions**
Forge: 14.23.5.2854
GTCE: 1.14.1.705
_Other mods_ :
Better Builder's wand : 0.13.2
BetterFPS : 1.4.87
Chamelon : 4.1.3
ChickenASM : 1.0.2.7
Code chickenLib : 1.0.2.7
COFH Core : 4.6.6.1
COFH World : 1.4.0.1
CT2 : 4.1.20.594
Garbage bin : 1.1.1
JEI : 4.16.1.302
TOP : 1.4.28
Redstone Flux : 2.1.1.1
Storage Drawer : 5.4.2
Thermal Expansion : 5.5.7.1
Thermal Foundation : 2.6.7.1

_Modpack:_ also append in the [modpack Gregicality Skyblock edition](https://www.curseforge.com/minecraft/modpacks/gregicality-skyblock-edition) in muliplayer

**Setup**
Playing Both multi and single player
tested in new world generated

**Steps To Reproduce**
1. build a large tungstenesteel large boiler 
2. start the boiler with water and whatever fuel you want
3. put large tungstensteel pipe next to the output hatch 
4. wait until the heat goes over 100°
5. no steam in the pipe

**Expected behavior**
steam get outputted

**Screenshots**
![](https://cdn.discordapp.com/attachments/763408591126265867/843172984910643230/2021-05-15_18.58.19.png)
![](https://cdn.discordapp.com/attachments/763408591126265867/843172994321874954/2021-05-15_18.58.20.png)

warjort:
Did you log out or stop the server at some point? See #1357

galyfray:
no most of the time login out and in or restarting the server fix the problem but sometime restarting the server each time someone has this kind of problem is not really possible

warjort:
Does placing a block next to the pipe fix the problem? I think that should cause the fluid pipe network to be recalculated.

The fluid output hatch auto outputs so the only way I can see that the fluid pipe would not accept inputs when empty would be if it think the fluid pipe network has 0 capacity and/or it hasn't formed properly?

[...]

PR #1612, opened 2021-05-18
Use TOP's item label to translate fuel item names

warjort:
**What:**
I noticed while looking at #1602 that the Fuelable TOP provider wasn't translating item names correctly.
This was especially true for our MetaItems where flawless coal was displayed as "item.metaitem".
But it was also true for things like a block a coal which instead of redirecting to the block name it just displayed something like "tile.coal"

**How solved:**
Made the Fuelable top provider use TOP's "itemLabel" element which invokes itemStack.getDisplayName() on the client which is always going to be correct.

**Outcome:**

**Additional info:**
My solution is kind of hacky in that it does special processing for ItemFuelInfo in the top provider.
The alternative would be let to IFuelInfo choose its own behaviour, but that would mean introducing TOP classes into
gregtech.api.capability classes. Something like
````
public interface IFuelInfo {
    // IProbeInfo is a TOP class
    void appendName(IProbeInfo probeInfo);
}
````

### DIFF
--- a/src/main/java/gregtech/api/capability/impl/ItemFuelInfo.java
+++ b/src/main/java/gregtech/api/capability/impl/ItemFuelInfo.java
@@ -17,4 +17,8 @@ public class ItemFuelInfo extends AbstractFuelInfo {
     public String getFuelName() {
         return itemStack.getTranslationKey();
     }
+
+    public ItemStack getItemStack() {
+        return this.itemStack;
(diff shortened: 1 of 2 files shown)


LAGIdiot (approved):
I think we can leave fix entirely on our side. As we are not aware of anyone needed it at the moment.

PR #1619, opened 2021-05-25
Miscellaneous Surface Rock updates

### LINKED ISSUE #1599: [BUG?] Surface rocks not generated on snow
**Describe the bug**
Apparently, surface rocks are not generated on snow (tile form, not block form).
It makes the early game very difficult when a player spawns in a snow-covered biome.
Is it intended?

**Versions**
Forge: forge-14.23.5.2855
GTCE: gregtech-1.12.2-1.14.1.705
CodeChickenLib-1.12.2-3.2.3.358-universal
NaturesCompass-1.12.2-1.8.5 (for testing)

**Setup**
Playing Solo, New world generated

**Steps To Reproduce**
1) generate a world
2) find snow biome (I used Nature's Compass for testing)
3) few surface rocks are generated

**Expected behavior**
Snow tiles are replaced into the surface rocks, which indicate the vein under the place.

ALongStringOfNumbers:
**What:**
This PR combines a couple of Miscellaneous Surface Rock fixes and updates.
First: Fixes surface rocks not generating on replaceable materials, Closes #1599 
Second: Updates the Surface rock indicator in the generation files to use the new surface rock generation syntax
Third: Changes some of the Surface rock materials to be more unique to more easily tell veins apart

**How solved:**
Adds checks for if the block is replaceable, as suggested in #1420 

**Outcome:**
Fixes surface rocks not generating on replaceable materials. Closes #1599.

**Additional info:**
To Easily follow the path of the new surface rock generation system, start in `OreDepositDefinition`, which then moves to `WorldGenRegistry#createVeinPopulator`, which then moves to `SurfaceRockPopulator` due to the registration for surface rocks, also in `WorldGenRegistry`.

I think that after this PR, the wiki section should be updated to describe the new surface rock syntax, and also state the surface blocks can be generated.

No compatibility issues expected. The change to the ore veins would only go into effect for freshly generated worlds, where existing files did not already exist.


DStrand1 (approved):
Looks like a good change to me. Will be nice to have Surface Rocks appear more consistently, and be more accurate to the veins they are trying to indicate are nearby. Also is a good change to display our newer and more powerful syntax for the surface rocks in our Ore Jsons, which is good for example code for addons and modpack creators

warjort:
I believe this reintroduces the bug where rocks generate in 1 deep water?

From MaterialLiquid
````
    public MaterialLiquid(MapColor color)
    {
        super(color);
        this.setReplaceable(); // HERE: fluids are replaceable
        this.setNoPushMobility();
    }
````

**Comment by warjort** [comment]
I also think that the Diamond veins being indicated by coal rocks is deliberate?

It encourages using the scanner to see whether a vein is really coal or diamond.

Your change means people can use TOP to see that it is graphite and therefore diamond.

File: src/main/resources/assets/gregtech/worldgen/overworld/molybdenum_vein.json [hunk 1]
```diff
@@ -3,8 +3,10 @@
   "density": 0.5,
```
LAGIdiot, line 3: I think we should rename this file to `molybdenite_vein.json` to correctly represent what is generated.

LAGIdiot (changes requested):
Honestly I had no idea that our vein files were in such inconsistent state. Thanks for taking look at it. I like changes in material for surface rocks as it makes more sense now.
There is only that one file I think should be changed.

src/main/resources/assets/gregtech/worldgen/overworld/molybdenum_vein.json
ALongStringOfNumbers: I have renamed this file.

warjort:
If you rename a worldgen file, users with existing worlds will have both the old and new files, leading to increased ore weights for that file.
Gregicality also has some processing to replace these files:
https://github.com/Gregicality/gregicality/blob/b81f376e730d67e7ed0ee83929b40e46d1a83758/src/main/java/gregicadditions/worldgen/WorldGenRegister.java#L60
Molybdenite isn't one of them.

LAGIdiot:
> If you rename a worldgen file, users with existing worlds will have both the old and new files, leading to increased ore weights for that file.

New files will not be extracted as long as `worldgen_extracted` exists under `config\gregtech`. By manually removing it new files will be roll out and it is not our responsibility to handle what users had there (they should not remove that file in first place if they did not want new files).

> Gregicality also has some processing to replace these files:
> https://github.com/Gregicality/gregicality/blob/b81f376e730d67e7ed0ee83929b40e46d1a83758/src/main/java/gregicadditions/worldgen/WorldGenRegister.java#L60
> Molybdenite isn't one of them.

Our vein files are not part of API (even though API is copying them out). So any modification to them is in our discretion. If anyone is doing something with them later it is their problem to keep up.

LAGIdiot (approved):
Thanks for your hard work!

src/main/resources/assets/gregtech/worldgen/overworld/monazite_vein.json
```diff
@@ -3,8 +3,10 @@
   "density": 0.4,
   "max_height": 40,
   "min_height": 20,
-  "surface_stone_material": "monazite",
-
+  "vein_populator": {
+    "type": "surface_rock",
+    "material": "monzanite"
```
warjort, line 8: Monazite is misspelt here. Doh!

PR #1625, opened 2021-05-27
Add vanilla dusts to dustRegular

### LINKED ISSUE #1616: [Suggestion] Add dustRegular oredict to Redstone
Redstone not having the dustRegular oredict makes it hard to fully automate ore processing, as redstone now cannot be easily removed with oredict filters.

Add a dustRegular oredict to redstone, like all other dusts have

warjort:
**What:**
Vanilla "dusts" don't have a "dustRegular" ore dictionary entry, meaning they are harder to differentiate from dustTiny,dustSmall, etc. using an ore dictionary filter.

**How solved:**
Changed OreDictionaryLoader for all vanilla dusts that are registered to OrePrefix.dust to also add them to "dustRegular".

**Outcome:**
Closes #1616 

The following are now part of a dustRegular ore dictionary:

* blaze powder
* bone meal
* glowstone
* gunpowder
* redstone
* sugar

**Additional info:**
#1616 is about ore processing and only redstone has an ore in default GTCE. 
These new entries can be used for other processing, e.g. using an ore dictionary filter to store full sized dusts seperately.

I made "dustRegular" a constant in OrePrefix and updated the existing usage.

None expected, since dustRegular is a GTCE invention.


DStrand1 (approved):
I’m not sure if I like the constant in OrePrefix since it could be confusing to someone looking at the file, but I agree that it should be a constant value, so I don’t have issue with it. Otherwise I agree that dustRegular should be on all of these.

Issue #789, opened 2019-04-30
Granite block mining becomes harder with ores embedded


pyure:
Ore + Granite seems to always have diamond-pick level mining for no reason.

Granite itself seems fine.  Ores seems fine.  Combination of them makes no sense.

https://github.com/pyure/infitech-prototype/issues/499

Archengius:
because black granite has diamond level and it's not going to change

pyure:
That's dumb.

WarlordWossman:
@Archengius you know this is about the vanilla granite with ores in it right?

Archengius:
Which is still baked to the black granite material. As i said earlier, i'd better remove these ore variants altogether instead of building special cases like that

pyure:
Re-opening this per conversation with LAGIdiot.  We want this to be revisited.

Issue #1589, opened 2021-05-07
Stone type of Granite is registered as GraniteBlack


tomasbrod:
Causing inconsistency in harvest levels.

https://github.com/GregTechCE/GregTech/blob/0983830269bb63290391cf5d18a5e84f822a454f/src/main/java/gregtech/api/unification/ore/StoneTypes.java#L25

ArchitektRadim:
Can confirm

warjort:
I am guessing this is a typo introduced during this refactoring a long time ago and the material should be Materials.Granite

https://github.com/GregTechCE/GregTech/commit/695097ef48b75415fcdc59656e93fae8c0aea18c#diff-20e691d6c95655a0f08b746b8d42d0b37cbe74f0fd757844ae686de84a34f68f

The two materials are not equivalent:
````
    public static DustMaterial Granite = new DustMaterial(449, "granite", 0xCFA18C, ROUGH, 2, of(), NO_SMASHING);
    public static DustMaterial GraniteBlack = new DustMaterial(251, "granite_black", 0x0A0A0A, ROUGH, 3, of(new MaterialStack(SiliconDioxide, 4), new MaterialStack(Biotite, 1)), NO_SMASHING);
````

AFAICT, the only use of this configuration is in BlockOre where it uses it to determine the tool level for mining 
````
    @Override
    public int getHarvestLevel(IBlockState state) {
        StoneType stoneType = state.getValue(STONE_TYPE);
        if (material != null) {
            int toolQuality = material.harvestLevel;
            return Math.max(stoneType.stoneMaterial.harvestLevel, toolQuality > 1 ? toolQuality - 1 : toolQuality);
        }
        return 1;
    }
````

BlackGranite is 3, while normal Granite is 2 - which is the subject of this bug report.

This seems to be a popular area for complaints, cf  
id shift #1549, 
dust byproducts #1553, 
tool harvest level #1567

tomasbrod:
Moreover, the vanilla stone has the same harestLevel as vanilla granite, which is 1, but Granite material is registered as harestLevel 2. Which, although inconsistent, is IMO fair.
https://github.com/GregTechCE/GregTech/blob/0983830269bb63290391cf5d18a5e84f822a454f/src/main/java/gregtech/api/unification/material/Materials.java#L348

PR #1627, opened 2021-05-27
Correct the material type for plain granite ore blocks

warjort:
**What:**
The material for StoneType plain granite is referencing black granite instead of plain granite.

**How solved:**
Changed Material.GraniteBlack to Material.Granite for plain granite stone type.
This looks like a copy and paste error during refactoring.

**Outcome:**
Addresses the main issue in #1589 
Fixes: #789 

You now only need a diamond level pick instead of an obsidian level pick for plain granite ores.

**Additional info:**
As #1589 points out. The granite stone block only requires an iron level pick so Material.Granite technically has the wrong harvest level (2 instead of 1), the same looks to be true for andesite and diorite.
This PR does not address that issue which could have more far reaching consequences. 
The change made only affects the mining levels of plain granite ores.


ALongStringOfNumbers (approved):

LAGIdiot (approved):
I agree with this change as these Granite types are not same. But at same time I am against change of mining level.

Also I will link #1589 to this as closed. If author of issue feels that mining level should be consistent new issue for it should be open.

LAGIdiot:
Also this fixes: #789

PR #1632, opened 2021-05-29
FuelRecipeLogic Rework

galyfray:
### What:
reworked the `FuelRecipeLogic` class adding methods and changing other methods to allow better flexibility of the class.

### How solved:

**methods added** :

- calculateRecipeDuration // equivalent to calculateRecipeDuration(lastRecipe)
- calculateFuelAmount // equivalant to calculateFuelAmount(this.previousRecipe)
- canProduceEnergy // to allow energy production when there is no recipe runing or without fuel consuption
- calculateFuelConsumptionMultiplier // applied in calculateFuelAmount(FuelRecipe).
- calculateStaticEnergyEfficiency // applied once on recipe start
- calculateDynamicEnergyEfficiency // applied every tick before producing energy
- calculateRecipeOutputVoltage // by default behave just like  `getRecipeOutputVoltage()`  but will include `calculateDynamicEnergyEfficiency` and return effective power output.
- calculateRecipeDurationMultiplier() // applied in calculateRecipeDuration(lastRecipe)
- hasRecipeEnded // return this.recipeDurationLeft <= 0 but can be overrided to include different element. 
- canConsumeFuel // used in the update method

#### Methods changed : 

- calculateFuelAmount // included calculateFuelConsumptionMultiplier
- calculateRecipeDuration // included calculateRecipeDurationMultiplier
- startRecipe  // included calculateStaticEnergyEfficiency

the `update` method has changed  in form but not in default behavior. the test `energyContainer.get().getEnergyCanBeInserted()>= calculateRecipeOutputVoltage() || shouldVoidExcessiveEnergy()` has been moved to `canProduceEnergy` by default `canConsumeFuel` return `canProduceEnergy` leading to the same behavior concerning power production and fuel consumption as before. New recipe is searched when the current has ended as before.

**Outcome**:

Make FuelRecipeLogic more flexible and easier to implement custom behavior.

### Additional info:

see #1604 for additional information on why and for what those changes are made

### Possible compatibility issue:

Possible compatibility issues with every multiblock using the FuelRecipeLogic as workable handler  or base for their workable handler. I will write reports on compatibility for each multiblock and provide as much proof has possible about the existence or not of compatibility issue. Help would be appreciated especially from add-on devs as I don't know each of them and they will know their code better than I could.

### DIFF
--- a/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
@@ -113,8 +113,15 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
 
         if (workingEnabled) {
 
-            boolean canConsume = canConsumeFuel();   // we need to compute those two before doing anything
-            boolean canProduce = canProduceEnergy(); // to avoid loosing 1 tick of production
+            /* we need to compute those two before doing anything
+             * to avoid loosing 1 tick of production
+             */
+
+            boolean canProduce = canProduceEnergy() &&
+                    (energyContainer.get().getEnergyCanBeInserted() >= calculateRecipeOutputVoltage() ||
+                            shouldVoidExcessiveEnergy());
+
+            boolean canConsume = canConsumeFuel() && (shouldConsumeWhenNotProducing() || canProduce);
 
             if (canConsume) {
                 --this.recipeDurationLeft;
@@ -139,6 +146,10 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
 
     protected boolean isReadyForRecipes() {
         return true;
+    }
+
+    protected boolean shouldConsumeWhenNotProducing() {
+        return false;
     }
 
     protected boolean shouldVoidExcessiveEnergy() {
@@ -248,13 +259,13 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
     }
 
     public boolean canProduceEnergy() {
-        return !hasRecipeEnded() &&
-                (energyContainer.get().getEnergyCanBeInserted() >= calculateRecipeOutputVoltage() ||
-                        shouldVoidExcessiveEnergy());
+        return !hasRecipeEnded();
     }
 
     public boolean canConsumeFuel() {
-        return canProduceEnergy();
+        return !hasRecipeEnded() &&
+                (energyContainer.get().getEnergyCanBeInserted() >= calculateRecipeOutputVoltage() ||
+                        shouldVoidExcessiveEnergy());
     }
 
     /**
--- a/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
@@ -88,8 +88,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
             if (fuelInfo == null) {
                 fuelInfo = new FluidFuelInfo(tankContents, fuelRemaining, fuelCapacity, amountPerRecipe, fuelBurnTime);
                 fuels.put(tankContents.getUnlocalizedName(), fuelInfo);
-            }
-            else {
+            } else {
                 fuelInfo.addFuelRemaining(fuelRemaining);
                 fuelInfo.addFuelBurnTime(fuelBurnTime);
             }
@@ -99,10 +98,10 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
 
     @Override
     public <T> T getCapability(Capability<T> capability) {
-        if(capability == GregtechTileCapabilities.CAPABILITY_CONTROLLABLE) {
+        if (capability == GregtechTileCapabilities.CAPABILITY_CONTROLLABLE) {
             return GregtechTileCapabilities.CAPABILITY_CONTROLLABLE.cast(this);
         }
-        if(capability == GregtechCapabilities.CAPABILITY_FUELABLE) {
+        if (capability == GregtechCapabilities.CAPABILITY_FUELABLE) {
             return GregtechCapabilities.CAPABILITY_FUELABLE.cast(this);
         }
         return null;
@@ -114,7 +113,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
         if (workingEnabled) {
             if (recipeDurationLeft > 0) {
                 if (energyContainer.get().getEnergyCanBeInserted() >=
-                    recipeOutputVoltage || shouldVoidExcessiveEnergy()) {
+                        recipeOutputVoltage || shouldVoidExcessiveEnergy()) {
                     energyContainer.get().addEnergy(recipeOutputVoltage);
                     if (--this.recipeDurationLeft == 0) {
                         this.wasActiveAndNeedsUpdate = true;
@@ -132,7 +131,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
     }
 
     protected boolean isReadyForRecipes() {
-       return true;
+        return true;
     }
 
     protected boolean shouldVoidExcessiveEnergy() {
@@ -211,6 +210,46 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
         return currentRecipe.getRecipeFluid().amount * getVoltageMultiplier(getMaxVoltage(), currentRecipe.getMinVoltage());
     }
 
+    public int calculateRecipeDuration() {
+        return calculateRecipeDuration(this.previousRecipe);
+    }
+
+    public int calculateFuelConsumption() {
+        return calculateFuelAmount(this.previousRecipe);
+    }
+
+    public boolean hasRecipeEnded() {
+        return !(this.recipeDurationLeft > 0);
+    }
+
+    public boolean canProduceEnergy() {
+        return !hasRecipeEnded();
+    }
+
+    public boolean canConsumeFuel(){
+        return !hasRecipeEnded();
+    }
+
+    protected double calculateFuelConsumptionMultiplier() {
+        return 1.0;
+    }
+
+    protected double calculateStaticEnergyEfficiency() {
+        return 1.0;
+    }
+
+    protected double calculateDynamicEnergyEfficiency() {
+        return 1.0;
+    }
+
+    public long calculateRecipeOutputVoltage() {
+        return (long) (getRecipeOutputVoltage() * calculateDynamicEnergyEfficiency());
+    }
+
+    protected double calculateRecipeDurationMultiplier() {
+        return 1.0;
+    }
+
     protected int calculateRecipeDuration(FuelRecipe currentRecipe) {
         return currentRecipe.getDuration();
     }
@@ -243,6 +282,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
             writeCustomData(1, buf -> buf.writeBoolean(active));
         }
     }
+
 
     @Override
     public void setWorkingEnabled(boolean workingEnabled) {
--- a/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
@@ -121,9 +121,9 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
                     (energyContainer.get().getEnergyCanBeInserted() >= calculateRecipeOutputVoltage() ||
                             shouldVoidExcessiveEnergy());
 
-            boolean canConsume = canRecipeProgress() && (shouldConsumeWhenNotProducing() || canProduce);
+            boolean canProgress = canRecipeProgress() && (shouldRecipeProgressWhenNotProducingEnergy() || canProduce);
 
-            if (canConsume) {
+            if (canProgress) {
                 --this.recipeDurationLeft;
             }
 
@@ -148,7 +148,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
         return true;
     }
 
-    protected boolean shouldConsumeWhenNotProducing() {
+    protected boolean shouldRecipeProgressWhenNotProducingEnergy() {
         return false;
     }
 
--- a/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
@@ -354,7 +354,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
         NBTTagCompound compound = new NBTTagCompound();
         compound.setBoolean("WorkEnabled", this.workingEnabled);
         compound.setInteger("RecipeDurationLeft", this.recipeDurationLeft);
-        if (recipeDurationLeft > 0) {
+        if (!hasRecipeEnded()) {
             compound.setLong("RecipeOutputVoltage", this.recipeOutputVoltage);
         }
         return compound;
@@ -367,10 +367,10 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
             this.workingEnabled = compound.getBoolean("WorkEnabled");
         }
         this.recipeDurationLeft = compound.getInteger("RecipeDurationLeft");
-        if (recipeDurationLeft > 0) {
+        if (!hasRecipeEnded()) {
             this.recipeOutputVoltage = compound.getLong("RecipeOutputVoltage");
         }
-        this.isActive = recipeDurationLeft > 0;
+        this.isActive = !hasRecipeEnded();
     }
 
 }
--- a/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
@@ -30,6 +30,11 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
 
     private int recipeDurationLeft;
     private long recipeOutputVoltage;
+    private long outputVoltage;
+    private boolean canProduceEnergy;
+    private boolean canProgress;
+    private int recipeDuration;
+    private int fuelAmount;
 
     private boolean isActive;
     private boolean workingEnabled = true;
@@ -43,8 +48,25 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
         this.maxVoltage = maxVoltage;
     }
 
+    /**
+     * Deprecated please use {@link #getOutputVoltage} to get the output voltage.
+     * overrides of this method should get moved to {@link #calculateRecipeOutputVoltage()}
+     */
+    @Deprecated
     public long getRecipeOutputVoltage() {
-        return calculateRecipeOutputVoltage();
+        return this.outputVoltage;
+    }
+
+    public long getOutputVoltage() {
+        return this.outputVoltage;
+    }
+
+    public boolean hasProducedEnergy() {
+        return this.canProduceEnergy;
+    }
+
+    public boolean hasRecipeProgressed() {
+        return this.canProgress;
     }
 
     @Override
@@ -117,18 +139,19 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
              * to avoid loosing 1 tick of production
              */
 
-            boolean canProduce = canProduceEnergy() &&
+            this.canProduceEnergy = canProduceEnergy() &&
                     (energyContainer.get().getEnergyCanBeInserted() >= calculateRecipeOutputVoltage() ||
                             shouldVoidExcessiveEnergy());
 
-            boolean canProgress = canRecipeProgress() && (shouldRecipeProgressWhenNotProducingEnergy() || canProduce);
+            this.canProgress = canRecipeProgress() && (shouldRecipeProgressWhenNotProducingEnergy() || this.canProduceEnergy);
 
-            if (canProgress) {
+            if (this.canProgress) {
                 --this.recipeDurationLeft;
             }
 
-            if (canProduce) {
-                energyContainer.get().addEnergy(calculateRecipeOutputVoltage());
+            if (this.canProduceEnergy) {
+                this.outputVoltage = calculateRecipeOutputVoltage();
+                energyContainer.get().addEnergy(outputVoltage);
             }
 
             if (hasRecipeEnded()) {
@@ -190,16 +213,17 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
             }
         }
         if (currentRecipe != null && checkRecipe(currentRecipe)) {
-            int fuelAmountToUse = calculateFuelAmount(currentRecipe);
-            if (fluidStack.amount >= fuelAmountToUse) {
+            this.fuelAmount = calculateFuelAmount(currentRecipe);
+            if (fluidStack.amount >= this.fuelAmount) {
                 this.recipeDurationLeft = calculateRecipeDuration(currentRecipe);
-                this.recipeOutputVoltage = startRecipe(currentRecipe, fuelAmountToUse, recipeDurationLeft);
+                this.recipeDuration = this.recipeDurationLeft;
+                this.recipeOutputVoltage = startRecipe(currentRecipe, fuelAmount, recipeDurationLeft);
                 if (wasActiveAndNeedsUpdate) {
                     this.wasActiveAndNeedsUpdate = false;
                 } else {
                     setActive(true);
                 }
-                return fuelAmountToUse;
+                return this.fuelAmount;
             }
         }
         return 0;
@@ -235,7 +259,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
     }
 
     public int getFuelAmount() {
-        return calculateFuelAmount(this.previousRecipe);
+        return this.fuelAmount;
     }
 
     protected double calculateRecipeDurationMultiplier() {
@@ -247,7 +271,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
     }
 
     public int getRecipeDuration() {
-        return calculateRecipeDuration(this.previousRecipe);
+        return this.recipeDuration;
     }
 
     /**
@@ -296,7 +320,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
      *
      * @return Effective power output of the machine.
      */
-    public long calculateRecipeOutputVoltage() {
+    protected long calculateRecipeOutputVoltage() {
         return (long) (this.recipeOutputVoltage * calculateDynamicEnergyEfficiency());
     }
 
--- a/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
@@ -129,11 +129,6 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
         return null;
     }
 
-    protected boolean hasRoomForEnergy() {
-        return (energyContainer.get().getEnergyCanBeInserted() >= calculateRecipeOutputVoltage() ||
-                shouldVoidExcessiveEnergy());
-    }
-
     @Override
     public void update() {
         if (getMetaTileEntity().getWorld().isRemote) return;
@@ -144,9 +139,9 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
              * to avoid loosing 1 tick of production
              */
 
-            this.canProduceEnergy = canProduceEnergy(hasRoomForEnergy());
+            this.canProduceEnergy = canProduceEnergy();
 
-            this.canProgress = canRecipeProgress() && (shouldRecipeProgressWhenNotProducingEnergy() || this.canProduceEnergy);
+            this.canProgress = canRecipeProgress();
 
             if (this.canProgress) {
                 --this.recipeDurationLeft;
@@ -279,12 +274,13 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
         return this.recipeDurationLeft <= 0;
     }
 
-    public boolean canProduceEnergy(boolean hasRoomForEnergy) {
-        return !hasRecipeEnded() && hasRoomForEnergy;
+    public boolean canProduceEnergy() {
+        return !hasRecipeEnded() && energyContainer.get().getEnergyCanBeInserted() >= calculateRecipeOutputVoltage() ||
+                shouldVoidExcessiveEnergy();
     }
 
     public boolean canRecipeProgress() {
-        return !hasRecipeEnded();
+        return !hasRecipeEnded() && (shouldRecipeProgressWhenNotProducingEnergy() || canProduceEnergy());
     }
 
     protected long calculateRecipeOutputVoltage() {
--- a/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
@@ -121,7 +121,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
                     (energyContainer.get().getEnergyCanBeInserted() >= calculateRecipeOutputVoltage() ||
                             shouldVoidExcessiveEnergy());
 
-            boolean canConsume = canConsumeFuel() && (shouldConsumeWhenNotProducing() || canProduce);
+            boolean canConsume = canRecipeProgress() && (shouldConsumeWhenNotProducing() || canProduce);
 
             if (canConsume) {
                 --this.recipeDurationLeft;
@@ -262,7 +262,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
         return !hasRecipeEnded();
     }
 
-    public boolean canConsumeFuel() {
+    public boolean canRecipeProgress() {
         return !hasRecipeEnded();
     }
 
--- a/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
@@ -263,9 +263,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
     }
 
     public boolean canConsumeFuel() {
-        return !hasRecipeEnded() &&
-                (energyContainer.get().getEnergyCanBeInserted() >= calculateRecipeOutputVoltage() ||
-                        shouldVoidExcessiveEnergy());
+        return !hasRecipeEnded();
     }
 
     /**
--- a/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
+++ b/src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
@@ -129,6 +129,11 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
         return null;
     }
 
+    protected boolean hasRoomForEnergy() {
+        return (energyContainer.get().getEnergyCanBeInserted() >= calculateRecipeOutputVoltage() ||
+                shouldVoidExcessiveEnergy());
+    }
+
     @Override
     public void update() {
         if (getMetaTileEntity().getWorld().isRemote) return;
@@ -139,9 +144,7 @@ public class FuelRecipeLogic extends MTETrait implements IControllable, IFuelabl
              * to avoid loosing 1 tick of production
              */
 
-            this.canProduceEnergy = canProduceEnergy() &&
-                    (energyContainer.get().getEnergyCanBeInserted() >= calculateRecipeOutputVoltage() ||
-                            shouldVoidExcessiveEnergy());
+            this.canProduceEnergy = canProduceEnergy(hasRoomForEnergy());
 
             this.canProgress = canRecipeProgress() && (shouldRecipeProgressWhenNotProducingEnergy() || this.canProduceEnergy);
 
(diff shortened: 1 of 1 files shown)


galyfray:
concerning the Large Turbines :

Report in progress

galyfray:
concerning the Diesel Engines:

As no changes has been done concerning the usage of overrided methods in `DieselEngineWorkableHandler` no issues have been found and none are expected.

galyfray:
Concerning Large Boilers : 

only use unchanged static methods, no issues have been found and none are expected.

galyfray:
concerning the `FueledMultiblockController` : 
`getRecipeOutputVoltage()` is used and should get replaced by `calculateRecipeOutputVoltage` soon or late to avoid showing wrong data. No changes will be done has `getRecipeOutputVoltage()` can be overrided.

A compatibility solution should be found to allow smooth transition.

galyfray:
concerning the `SimpleGeneratorMetaTileEntity`

No issues have been found and none are expected.

warjort:
I don't understand why you have changed everything to calculateXXX().

I suggested changing policy methods to be called calculateXXX() to differentiate them from accessors that are used to expose metrics on displays, getXXX().

warjort:
I also don't like the hasRecipeEnded(), canProduceEnergy() and canConsumeFuel()

I know you are just trying to reproduce the old logic in a way that can be overridden, but this abstraction doesn't look very clean.

e.g. if I want to override canProduceEnergy() I also need to reproduce the insertion energy check in my subclass.
Previously this logic was dictated by the shouldVoidExcessEnergy() policy without me having to worry about the mechanics of how it was done.

You should make thes policy methods "do one thing well" and not mix things together.

galyfray:
>I don't understand why you have changed everything to calculateXXX().
I suggested changing policy methods to be called calculateXXX() to differentiate them from accessors that are used to expose metrics on displays, getXXX().

I didn't correctly understand what you was meaning when saying that. For me a getXXX method is a getter which return a stored value while calculateXXX means you have calculation behind it.

can you give me a list or something more detailed on what do think should be a getXXX or a calculateXXX ?

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
```diff
@@ -207,21 +213,90 @@ public long getMaxVoltage() {
...
     protected int calculateFuelAmount(FuelRecipe currentRecipe) {
-        return currentRecipe.getRecipeFluid().amount * getVoltageMultiplier(getMaxVoltage(), currentRecipe.getMinVoltage());
+        return (int) (currentRecipe.getRecipeFluid().amount * getVoltageMultiplier(getMaxVoltage(), currentRecipe.getMinVoltage()) * calculateFuelConsumptionMultiplier());
+    }
+
+    public int calculateFuelAmount() {
+        return calculateFuelAmount(this.previousRecipe);
+    }
+
+    protected double calculateRecipeDurationMultiplier() {
```
warjort, line 228: This is policy that somebody can override and it will change the behaviour of the machine.
So calculateX() is correct.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
```diff
@@ -207,21 +213,90 @@ public long getMaxVoltage() {
...
+        return 1.0;
     }
 
     protected int calculateRecipeDuration(FuelRecipe currentRecipe) {
-        return currentRecipe.getDuration();
+        return (int) (currentRecipe.getDuration() * calculateRecipeDurationMultiplier());
+    }
+
+    public int calculateRecipeDuration() {
```
warjort, line 236: This is a convenience accessor so subclasses can display info, so getXXX()
Overridding this method will do nothing to machine's behaviour.

galyfray: okay get it I guess

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
```diff
@@ -207,21 +213,90 @@ public long getMaxVoltage() {
...
+    protected double calculateFuelConsumptionMultiplier() {
+        return 1.0;
+    }
+
     protected int calculateFuelAmount(FuelRecipe currentRecipe) {
-        return currentRecipe.getRecipeFluid().amount * getVoltageMultiplier(getMaxVoltage(), currentRecipe.getMinVoltage());
+        return (int) (currentRecipe.getRecipeFluid().amount * getVoltageMultiplier(getMaxVoltage(), currentRecipe.getMinVoltage()) * calculateFuelConsumptionMultiplier());
+    }
+
+    public int calculateFuelAmount() {
```
warjort, line 224: This is an accessor that can be used to display information, so getFuelAmount()
Does not change behaviour.

galyfray:
>e.g. if I want to override canProduceEnergy() I also need to reproduce the energy insertion check in my subclass.
Previously this logic was dictated by the shouldVoidExcessEnergy() policy without me having to worry about the mechanics of how it was done.
You should make these policy methods "do one thing well" and not mix things together.

The fact of looking if there is room for energy is a part of the "can you produce power" for me. still I understand your point and changing this isn't a big deal for energy production but for fuel consumption it will stay not as clean as you may want.

warjort:
> The fact of looking if there is room for energy is a part of the "can you produce power" for me

But as I said, there is already a policy for that, I shouldn't have to override it again, or even worse copy/paste the implementation details to override some other policy.

If you are mixing things together or forcing people to reproduce implementation details in their override it is a sign that the abstraction is wrong and that you need some other policy override option or maybe multiple policy overrides?

galyfray:
>But as I said, there is already a policy for that, I shouldn't have to override it again, or even worse copy/paste the implementation details to override some other policy.

that's why I said I understand your point I'm working on fixing this at least for the energy part and I'm thinking of a way to do so for fuel

galyfray:
The abstraction has been cleaned up `canProduce` and `canConsume` are bonded by `shouldConsumeWhenNotProducing` allowing to override this behavior easily. the voiding energy policy has been moved again in the update function making the canProduceEnergy much clearer as well as the canConsumeFuel

warjort:
The canConsumeFuel() is confusing to me.

I think of consuming fuel as the using of fuel in the input hatch that is consumed when starting the recipe.

I think this flag would be better named as something like canProgressRecipe() since that is what it is actually controlling.

galyfray:
>The canConsumeFuel() is confusing to me.
I think of consuming fuel as the using of fuel in the input hatch that is consumed when starting the recipe.
I think this flag would be better named as something like canProgressRecipe() since that is what it is actually controlling.



I was looking for a new name for this one since a long time ...

warjort:
The same is true for the shouldConsumeWhenNotProducing().

Is this shouldProgressRecipeWhenNotProducingEnergy()

Phew, that's a long name :-)

galyfray:
>Is this shouldProgressRecipeWhenNotProducingEnergy()

ups forget this one and I think RecipeProgress instead of ProgressRecipe makes more sense

warjort:
You should change the canConsume local variable as well.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
```diff
@@ -44,7 +44,7 @@ public FuelRecipeLogic(MetaTileEntity metaTileEntity, FuelRecipeMap recipeMap, S
     }
 
     public long getRecipeOutputVoltage() {
-        return recipeOutputVoltage;
+        return calculateRecipeOutputVoltage();
```
Archengius, line 47: Do you understand the difference between getXXX and calculateXXX?
getXXX methods are supposed to return cached values
calculateXXX methods are used to actually calculate values and then cache them into the fields

So why do you call calculateRecipeOutputVoltage inside of getRecipeOutputVoltage

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
```diff
@@ -111,18 +110,32 @@ public int getNetworkID() {
...
+            boolean canConsume = canConsumeFuel() && (shouldConsumeWhenNotProducing() || canProduce);
+
+            if (canConsume) {
+                --this.recipeDurationLeft;
             }
-            if (recipeDurationLeft == 0 && isReadyForRecipes()) {
-                tryAcquireNewRecipe();
+
+            if (canProduce) {
+                energyContainer.get().addEnergy(calculateRecipeOutputVoltage());
```
Archengius, line 131: You should probably call calculateRecipeOutputVoltage every tick and write result into the backing field, then use it's value here and in the getOutputVoltage method.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
```diff
@@ -207,21 +224,88 @@ public long getMaxVoltage() {
...
+    protected double calculateFuelConsumptionMultiplier() {
+        return 1.0;
+    }
+
     protected int calculateFuelAmount(FuelRecipe currentRecipe) {
-        return currentRecipe.getRecipeFluid().amount * getVoltageMultiplier(getMaxVoltage(), currentRecipe.getMinVoltage());
+        return (int) (currentRecipe.getRecipeFluid().amount * getVoltageMultiplier(getMaxVoltage(), currentRecipe.getMinVoltage()) * calculateFuelConsumptionMultiplier());
+    }
+
+    public int getFuelAmount() {
```
Archengius, line 235: Same issue like with getOutputVoltage: getters should return cached values, not calculate new ones

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
```diff
@@ -207,21 +224,88 @@ public long getMaxVoltage() {
...
+        return 1.0;
     }
 
     protected int calculateRecipeDuration(FuelRecipe currentRecipe) {
-        return currentRecipe.getDuration();
+        return (int) (currentRecipe.getDuration() * calculateRecipeDurationMultiplier());
+    }
+
+    public int getRecipeDuration() {
+        return calculateRecipeDuration(this.previousRecipe);
```
Archengius, line 248: Same issue like with all your getters again

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
```diff
@@ -207,21 +224,88 @@ public long getMaxVoltage() {
...
+    }
+
+    /**
+     * this method aim to tell if the current recipe has ended in order to check for a new recipe see {@link #isActive}
+     * or {@link #isWorkingEnabled} if you want to know if the machine is running.
+     *
+     * @return true if no recipe is running false otherwise
+     */
+    public boolean hasRecipeEnded() {
```
Archengius, line 257: This method is pointless and badly designed. Why would you EVER want to override it? 
If you want it to prevent machine from picking up new recipe, you have isReadyForRecipes
If you want to stop machine from producing energy, you have canProduceEnergy
If you want to affect recipe duration, you have calculateRecipeDuration

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
```diff
@@ -207,21 +224,88 @@ public long getMaxVoltage() {
...
+        return !hasRecipeEnded();
+    }
+
+    /**
+     * Compute a multiplier used by default in {@link #startRecipe(FuelRecipe, int, int)}  startRecipe} each time the recipe start.
+     * fast changing efficiency could be implemented in {@link #calculateDynamicEnergyEfficiency() calculateDynamicEnergyEfficiency} for better responsiveness of the changes.
+     *
+     * @return power multiplier
+     */
+    protected double calculateStaticEnergyEfficiency() {
```
Archengius, line 275: What is the point of this method? What is the point of having "static" energy efficiency in general?
What can you achieve using "static" energy efficiency that you cannot achieve by returning constant value in calculateDynamicEnergyEfficiency?

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
```diff
@@ -207,21 +224,88 @@ public long getMaxVoltage() {
...
+    }
+
+    /**
+     * Apply {@link #calculateDynamicEnergyEfficiency() calculateDynamicEnergyEfficiency} multiplier
+     * to the outputVoltage returned by {@link #getRecipeOutputVoltage(), getRecipeOutputVoltage}
+     * in order to get the effective power output used in the {@link #update() update} method.
+     *
+     * @return Effective power output of the machine.
+     */
+    public long calculateRecipeOutputVoltage() {
```
Archengius, line 297: If you can override this method, why do you even have calculateDynamicEnergyEfficiency?
Again, what prevents you from just doing super.calculateRecipeOutputVoltage() * whateverYouWant in your child class?

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
```diff
@@ -207,21 +224,88 @@ public long getMaxVoltage() {
...
+        return (long) (this.recipeOutputVoltage * calculateDynamicEnergyEfficiency());
     }
 
     /**
      * Performs preparations for starting given recipe and determines it's output voltage
+     * the {@link #calculateStaticEnergyEfficiency() calculateStaticEnergyEfficiency} multiplier is by default applied in this method.
      *
      * @return recipe's output voltage
      */
     protected long startRecipe(FuelRecipe currentRecipe, int fuelAmountUsed, int recipeDuration) {
```
Archengius, line 307: What's even the point of this method returning any value if you have it recomputed every tick?

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
```diff
@@ -111,18 +110,32 @@ public int getNetworkID() {
...
+
+            /* we need to compute those two before doing anything
+             * to avoid loosing 1 tick of production
+             */
+
+            boolean canProduce = canProduceEnergy() &&
+                    (energyContainer.get().getEnergyCanBeInserted() >= calculateRecipeOutputVoltage() ||
+                            shouldVoidExcessiveEnergy());
+
+            boolean canConsume = canConsumeFuel() && (shouldConsumeWhenNotProducing() || canProduce);
```
Archengius: Both of these seem to be bad abstractions to build logic upon. Why do you even have logic additionally to calling canProduceEnergy here? You should literally move it to default implementation of canProduceEnergy and just use canProduceEnergy without doing any additional checks

Same applies to canConsumeFuel

Archengius (changes requested):
Literally all abstractions reproduced by this PR fall into either of 3 categories:
* Duplicated functionality (mostly from other abstractions added in this PR)
* Unnecessary and bloated functionality (like dynamic + static efficiency multipliers)
* Functionality that is still not flexible enough and has hardcoded conditions (like with canConsumeFuel or canProduceEnergy)

I heavily recommend reworking it from the ground while trying to keep things as simple as possible and avoid introducing unnecessary functionality

warjort:
You have changed the logic for wasActiveAndNeedsUpdate.

Previously this was only set to true when the recipeDurationLeft was changed to zero.
Now it is always being set to true if the recipeDurationLeft has the value zero

I am unsure what this would change but you should try to keep the old logic to avoid introducing subtle bugs.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: simple yet it requires to read every comment this is to fix  the `FueledMultiblockController` problem caused by the large turbine overriding this methods. If you find a better way to handle possible modification of this methods not breaking what is showed in the GUI go ahead and tell this in comment.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: >This method is pointless and badly designed. Why would you EVER want to override it? 

>Additional info:
see #1604 for additional information on why and for what those changes are made

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: large turbine overrides this method because it was written... before that PR?
With code introduced in this PR, getters should be literally made final, and large turbine should override calculateRecipeOutputVoltage

galyfray: yes but that may break code which use the same behavior in addon this is why I don't touch at any other class to fix issue in this class this way I am sure I won't be obviously breaking things

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: I don't see this method overriden in large turbine in this PR. So why is it needed?

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: We literally have all addon developers in touch with GTCE changes, doing changes like that is not an issue at all.

warjort: The point of calling it getX() instead of calculateX() is to differentiate policy from metrics.

It is very confusing to have:

calculateSomething(Recipe r) // override this to change behaviour
calculateSomething() { return calculateSomething(lastRecipe)) // doesn't change behaviour, just exposes info for display

If you believe getX() should be reserved for javabean accessors then some other prefix should be used that makes it obvious it is not a policy override

But the rational for the extra dance in this location is backwards compatibility:
https://github.com/GregTechCE/GregTech/pull/1604#issuecomment-846636440
This is to workaround the hack where LargeTurbine changes the contract of getRecipeOutputVoltage() to be every tick instead of at the start of the recipe and the possiblity that an addon has overridden it.

galyfray: >We literally have all addon developers in touch with GTCE changes, doing changes like that is not an issue at all.

LAG Idiot 12 day ago:
>We should try not break our addons even if they are using common part of mod.

when talking about a small and easy to fix changes in the common part of GTCE . here we are in the API I rather have a method with a name a little weird when looking at what it does than breaking things in the API.

Archengius: the point of calling it getX is to indicate that it retrieves metrics, indeed.
problem is, in that exact case it does NOT retrieve metrics, it literally recomputes value in place instead of providing value that was ACTUALLY used this tick for adding energy into the buffer.

getX methods usually do not indicate a policy to override, even if they are not marked as final. But yeah, to avoid confusion, you can literally mark them final to prevent people from overriding them instead of calculateX methods.

Again, like I said above, we have literally all addon developers in touch with GTCE development, so even if it breaks any functionality in addons, fix should not take much time to implement.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
warjort: No, because then the subclass then has to reproduce these implementation details to override the policy and will get out of sync when new policy is added to the default implementation.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: why would I ever do that ? I mean yes that would work but how would you me to name that ? and why would a getter called getXXX returning something else than XXX but YY a value created from another methods ? it doesn't makes sense

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: > when talking about a small and easy to fix changes in the common part of GTCE . here we are in the API I rather have a method with a name a little weird when looking at what it does than breaking things in the API.
If you really want to keep compatibility that much, you should actually deprecate this method and provide a proper getter method which does not recompute value

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: >I don't see this method overriden in large turbine in referenced PR. So why is it needed?

read the comments not the code this methods doesn't yet exist of course it can't be overriden. this part has been extensively discussed

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: Because ideally you want value that was ACTUALLY used to generate energy this tick exposed via getRecipeOutputVoltage, and not recomputed every time getter is called, potentially producing unwanted side effects and giving incorrect results when other fields are changed.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: once again go and read the comment in #1604 I pointed that in the additional context.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: You realize you can just call super method to follow parent policy and then && with your custom condition, right?
Hard-coding conditions like that just to force parent implementation is stupid.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: well let's say backward compatibility ? no because you see in #1604 I ignored the value returned at one time and a wise guy told me it wasn't a good idea

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: Still makes no sense to introduce method for something so niche and resolvable by other means

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: read warjot comment please this has implemented like this because he pointed out that doing this in the way you want was messy and I agree

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: Still makes no sense to introduce method for something so niche and resolvable by other means

galyfray: >Still makes no sense to introduce method for something so niche and resolvable by other means

find a way to resolve the problem it was intended to fix and post it as a comment in #1604 if you are able to do so then this method will become useless I'll be glad to not to have to change RecipeDurationLeft to a float

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: I repeat, backwards compatibility is really no an issue, especially for that particular method

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: >You realize you can just call super method to follow parent policy and then && with your custom condition, right?
Hard-coding conditions like that just to force parent implementation is stupid.

It's not stupid it's clearer you can easily change canProduce without minding you can change the voiding policy without minding or risking affecting the other policy

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: Once again, your comment on that PR does not advocate adding this method
You gave only a single argument, which is "performance", but performance difference is literally nonexistent in that particular case, and it makes no sense to introduce additional useless method for it

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
warjort: Only for simple overrides. 
If the default implementation is doing 5 steps and I want to change logic at step 3, super isn't going to help.
The user should not have to reproduce implementation details (that will change) 
for anything beyond a BEFORE or AFTER injection.
Like I said above, these policy overrides should concentrate on "doing one thing well".

Archengius: Yeah, you do not risk changing it because, you know, you cannot change it at all with your implementation?
If you consider calling super implementation such a difficult aspect to be mindful about, I have some bad news for you

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: >You gave only a single argument, which is "performance", but performance difference is literally nonexistent in that particular case, and it makes no sense to introduce additional useless method for it

okay in this case I don't see any reason keeping those methods separated

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: > Only for simple overrides.
If the default implementation is doing 5 steps and I want to change logic at step 3, super isn't going to help.
The user should not have to reproduce implementation details (that will change)
for anything beyond a BEFORE or AFTER injection.

Yeah, and the funniest thing is, do you actually realize *current* implementation only allows simple overrides? It literally does && with the value returned from canProduceEnergy(), you cannot change anything after this point.
Keeping it like that does not solve the issue of injecting into original logic, it literally makes it even worse because you just CANNOT change parent logic at all.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: Alright, judging from the code in your turbine PR, you want to compare it against getRecipeDurationMultiplier() and not zero. But the question is, why can't you just let it tick for one more time and decrease remaining ticks by getRecipeDurationMultiplier(), making it zero and not needing any changes?
I really don't see why it should REALLY stop when value is less than multiplier and cannot work for one more tick and then stop when it's actually zero or below.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: >If you really want to keep compatibility that much, you should actually deprecate this method and provide a proper getter method which does not recompute value

in deed I should, I didn't really think about this.

> it literally recomputes value in place instead of providing value that was ACTUALLY used this tick for adding energy into the buffer.

No need for caps locks.

yes values used to compute could have changed depending on the calling order of block updates this should get fixed

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: no need for caps lock.

yes that is a  solution and a problem. please when you are proposing changes which fix a problem that haven't been pointed out yet point it out first then propose the solution this way makes more sense.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
warjort: > Yeah, and the funniest thing is, do you actually realize current implementation only allows simple overrides? It literally does && with the value returned from canProduceEnergy(), you cannot change anything after this point.

Not true because there are 2 policy options in play here;

canProduceEnergy()
shouldVoidExcessEnergy()

You can override these policies independently without worrying about implementation details.
You don't have worry about updating your override if somebody introduces a 3rd piece of logic (possibly overriden by an intermediate class in the hierarchy).

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: how about `calculateFuelAmount` and `calculateRecipeDuration` and their respective multiplier ?
for me this is more about make things clear, what methods does what. Each things does one thing and do it clearly and properly. 

`calculateRecipeOutputVoltage` get all the different multipliers and apply them in his way but doesn't touch to how are those multipliers are computed.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: you can still override and not worry about implementation details in parent, granted you just call super
Like really, this logic being moved to canProduceEnergy does not change anything, but allows you to completely override it if you really need to

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
Archengius: I understand your desire to keep things clean, but you just end up making them more bloated than they should be.
I really think in that particular case it's better to have very few methods for changing functionality, granted they're all flexible and simple to use and understand.

Archengius: Like, I would literally just leave 3 methods for overrides: calculateRecipeDuration, calculateRecipeOutputVoltage and calculateFuelAmount, plus have their respective getters as final methods. That way it's very minimalistic and simple to understand.
Obviously that does not cover all functionality and does not allow overriding fuel consumption/energy production conditions, so that's why we introduce canProduceEnergy and shouldConsumeFuel, which ideally now contain logic that used to be hard-coded inside of the update() method
Plus it's probably a good idea to keep some of the utility methods for customizing default functionality without fully rewriting it, like shouldVoidExcessiveEnergy (or whatever that method was called)

That's it, I would literally keep only these few methods and nothing else

galyfray: I think that the energy voiding policy should be placed in an other method to keep things splitted this way warjot will be happy because those two policy will be in different methods and won't be hard coded what do think of that ? 

considering reducing the number of methods I think both ways are acceptable I prefer to have different methods to keep things splitted . I will ask others council member their opinion concerning this question.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
warjort: If you are completely overridding in a sledgehammer way, you are probably extending the wrong class?
I still disagree about having to redo implementation details (which is fairly easy to show) but this argument could go on forever.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: the previous code was using `== 0` not `<= 0` so yeah. 
This method would still be useful to stop the current recipe on other conditions than just recipe duration.

>I really don't see why it should REALLY stop when value is less than multiplier and cannot work for one more tick and then stop when it's actually zero or below.

just to not overproducing by one tick yes this is little but I don't want the new implementation to contain known and ignored bugs.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: >but performance difference is literally nonexistent in that particular case, and it makes no sense to introduce additional useless method for it

just asking but if you have a relatively  expensive computation needed only each recipe start and the recipe is very long could it be worth to keep things splitted ?

galyfray:
As pointed by @Archengius getter should reworked as they can send value different from what was used in the computation.
quick fix : store those values in protected field when computed. there is probably no need to store those new value in the NBT.

serenibyss (changes requested):
I will say that I have not fully reviewed this PR, nor the original one focused on Turbines specifically. However, to me it seems very overcomplicated, and as someone who has not read all of the discussion as it is quite long, and is simply looking at the list of API method changes, I find this less productive than before, simply because of how complicated it is. Personally, I think that Archengius’s suggestion  [here](https://github.com/GregTechCE/GregTech/pull/1632#discussion_r641966902) makes the most sense to me as the path forward.

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: >If you really want to keep compatibility that much, you should actually deprecate this method and provide a proper getter method which does not recompute value

just think about that but is it really needed to replace this method with a new one ? is there a reason why you would need the original recipe duration and not the effective one ?

src/main/java/gregtech/api/capability/impl/FuelRecipeLogic.java
galyfray: as Dstrand1 agree with you I will apply your changes

warjort:
> Personally, I think that Archengius’s suggestion here makes the most sense to me as the path forward.

I am reserving judgement until I see how the implementation of the throttle now works.

Without the multiplier injection exposed explicitly, I can see it being harder to implement a generic throttle that can be shared across generator implementations including those that have their own duration, consumption and energy modifications like the diesel generator?

This was kind of the first step so the throttle can reason about (and affect) all the generators in a generic way.

I think the main reason for adding the other methods was to try to integrate the turbine behaviour back into FuelRecipeLogic so it didn't have to "corrupt" the contract of quite a few important methods, again allowing the throttle to reason consistently.

It seems to me that with these more broad brush overrides we will again see kludges but now it will be in the new methods? :-)

warjort:
The way I envisioned this feature working was that you;

* have a "widget" that you can add to your multiblock display or single block generator that implements a generic throttle
* You can initialise the throttle with tunable parameters/functions to dictate the efficiency curve(s)
* You wire up your workable handler to the throttle, e.g. the multiplier overrides are just delegated to the throttle

That's it. The generators don't need invasive changes into their energy production logic. 
The throttle works with the FuelRecipeLogic as a kind of back channel.

I did suggest that for the multiblocks the throttle could be a multiblock part with its own multiblock ability.
That would make the integration pretty much automatic for all MB generators just by changing the pattern to allow it.
But that has issues with how to inject the GUI unless it is actually in the multiblock part and not the controller.

You could do the same for single blocks if the throttle was a cover.

galyfray:
>You wire up your workable handler to the throttle, e.g. the multiplier overrides are just delegated to the throttle
you want the throttle widget to override the multiplier by itself ?

warjort:
> you want the throttle widget to override the multiplier by itself ?

Is there a reason why it can't? 

It's not really by itself if you allow the calculation to be configurable on the throttle.
Though I would expect the throttle to have a default implementation.

And if you really needed to modify the behaviour in a gross way you can always do:

````
protected double calculateMultiplier() {
    double throttleResult = throttle.calculateMultiplier();
    return doWhateverYouWant(throttleResult);
}
````

But @Archengius said he doesn't like this multiplier policy being seperated so these things are more difficult to achieve.

galyfray:
@Archengius @warjort I applied some changes I'm not sure about the utility of some getters like the `hasRecipeProgressed` one
can you give me feedback ?

galyfray:
I changed the `canProduceEnergy` in a way that allow full customization and doesn't require to redo implementation detail. If this way satisfy anyone I will change the `canRecipeProgress` to behave this way too.

warjort:
> I changed the canProduceEnergy in a way that allow full customization and doesn't require to redo implementation detail.

I don't think your compromise will satisfy either @Archengius or myself. :-)

@Archengius doesn't want a proliferation of methods in the interface that are just there on the slight chance that they might be useful, something I agree with.

While I want a proper "separation of concerns" - the "do one thing well" principle.

Let me give a concrete example loosly based on the code you are modifying.

````
public boolean FuelRecipeLogic.canProduceEnergy() {
    return energyContainerCheck || voidable();
}
@Override
public boolean RotorLogic.canProduceEnergy() {
   return super.canProduceEnergy() && rotorLogic
}
````
Scenario 1: I want to override the rotor logic in the turbine so I have to reproduce the implementation detail:
````
@Override
public boolean TurbineLogic.canProduceEnergy() {
    // Notice I can't use super
    return (energyContainerCheck || voidable()) && changedRotorLogic;
}
````

Scenario 2: Somebody now wants a new check to FuelRecipeLogic
````
public boolean FuelRecipeLogic.canProduceEnergy() {
    return energyContainerCheck || voidable() || newLogic;
}
````
I now have to know the turbine logic needs changing as well (assuming it is not in an addon project)

My argument is that the stuff I wrote in FuelRecipeLogic.canProduceEnergy() above doesn't even need to overridable. There is already the well defined voidable() api to change the behaviour.
The override should be tailored to the specific behaviour.

````
boolean canProduce = canProduceEnergy() && (energyContainerCheck || voidable())
protected boolean FuelRecipeLogic.canProduceEnergy() {
    return true;
}
@Override
protected boolean RotorLogic.canProduceEnergy() {
   return rotorLogic();
}
protected boolean RotorLogic.rotorLogic() {
   // whatever
}
@Override
protected boolean TurbineLogic.rotorLogic() {
   // Override what you want to change rather than mixing different concerns
}
````

Notice in my version I don't need to call super or copy/paste code. And it is easier to reason about what is happening.
And I can add newLogic with no problems without having to worry that somebody might not be invoking super.
````
boolean canProduce = canProduceEnergy() && (energyContainerCheck || voidable() || newLogic)
````

galyfray:
my actual version giving (energyContainerCheck || voidable()) as a parameter for canProduceEnergy is resilient to the addition of new behavior  like the newLogic thing, if the main problem is the addition of a new method I can remove it and simply put the expression directly into the call as a parameter reducing useless methods while keeping a  proper "separation of concerns" as there will be just a float to take care of if you want.

warjort:
> or should we call a third person to decide which way will be used

@Archengius has review authority, I don't

warjort:
> if the main problem is the addition of a new method I can remove it and simply put the expression directly into the call as a parameter reducing useless methods while keeping a proper "separation of concerns"

I don't think it is a real "separation of concerns" because in that newLogic example you are going to have to || it with
hasRoomForEnergy. But that newLogic might not have anything to do with the hasRoomForEnergy state.
````
public boolean canProduceEnergy(boolean hasRoomForEnergy) {
       return !hasRecipeEnded() && hasRoomForEnergy;
}
````

galyfray:
@Archengius  is this okay now ?

galyfray:
I won't work on this anymore, I don't even remember what I was doing in this PR, so I close it

PR #1633, opened 2021-05-29
Add Component Assembler recipes

### LINKED ISSUE #1605: [Suggestion] Assembler Recipes for components
## Main Idea

I am opening this issue as a request for feedback before I take a look at implementing this feature in GTCE, so any feedback on my proposed idea and possible recipes would be appreciated.

This idea came about due to the fact that both addon mods SoG and Gregicality currently implement Component (Robot arm, Motor, etc) recipes in the Assembler, although the recipes added by both SoG and Gregicality could use some revamping (As agreed upon both both dan and I). Currently, the only component that has an Assembler recipe in GTCE is the Field Generator.

While talking about the potential revisions that could be made to the recipes, I brought up the fact that it might be better to have these recipes in GTCE directly, so that they can stop being added through addon mods, unless they want to modify the recipes.

Some things that were brought up in the discussion were:

- The speed of the recipe has to compete with Molecular Assemblers, to be favored over just a plain crafting recipe.

Currently, aside from component recipes with tools, Molecular Assemblers (or the Refined Storage equivalent) can make these recipes pretty fast. Therefore, to be competitive, the assembler recipes should be somewhat fast as well. The currently agreed upon time is a flat 5 seconds per component. This is the main point that needs to be revised from both addon mods, as currently the recipes take a flat 30 seconds.

- Recipes should have on-tier power requirements to make their components.

To incentivize progressing to the assembler for each tier, the component recipes should have power requirements in the assembler matching the tier that they are. EG. MV component takes MV power. This would allow for GTCE to implement recipes for up to IV components, which matches well with both addon mods, because components of the LuV tier and higher are required to be made in the Assembly line in both addon mods.

- Assembler Recipes should be somewhat different from the Crafting Table recipe

For example, the Pump Recipe. Currently the recipe takes rubber rings (Both regular rubber and Styrene Butadiene rubber can be used, and SoG adds a recipe with Silicone Rubber). When transitioned to an Assembler recipe, these rings could be replaced with fluids of the respective type, Rubber, SBR, and Silicone Rubber. (However, the recipe with paper rings in LV would be left as crafting table only). In addition, it might be possible to drop the required cable or some other part in the assembler recipe (Here is where I am looking for some feedback)

Another example is the Motor recipe. Currently, the motor requires increasing sizes of copper (Annealed copper for EV and IV) wires. These copper wires could be transitioned to a fluid component in the assembler recipe.

Another example is that the Emitter recipes could have their rods transferred to fluids of the same material.

### Summary

I am opening this issue to gather feedback on implementing Component recipes in the assembler, and possible recipes for them. These recipes should be somewhat different from the regular crafting table recipes, I and have stated some possible recipes for a few components in the section above.

Edit:
As an addition to adding recipe for components, I feel like the assembler recipes for Field Generators should be updated and brought in line with the added recipes, namely so they do not take 90 seconds (These component recipes are added by GTCE).

ALongStringOfNumbers:
**What:**
Adds assembler recipes for components, as described in #1605, with a few minor changes. The recipes in the assembler are mostly the exact same as the hand crafting recipes, except in the case of conveyors and pumps, which have had their rubber components transitioned to fluids. 

For the pumps specifically, I have added the ability for pumps to be made with Silicone Rubber in the assembler, and also introduced fluid cost savings depending on the fluid type used, with the fluid tiers following the same as for cables.

In addition, this PR changes the existing assembler recipes for EV and IV Field Generators to use on tier circuits, instead of tier + 1 circuits. This was done to match the crafting recipes, and the fact that other components taking on tier circuits

**How solved:**
Adds a new class `ComponentRecipes`, which handles all of the assembler component recipes. Moves the Field Generators into this class as well.

**Outcome:**
Adds assembler recipes for components. Closes #1605.

**Additional info:**
As it would be too long to display all the new recipes here, I will simply state that all assembler recipes are the same as the crafting table recipes, with the exception of the recipes specifically mentioned in the opening section.

Addon mods could stop adding these recipes, but there is no issue for addon mods, just temporary duplicate recipes. For modpack makers, they could use these recipes or remove them with crafttweaker.

### DIFF
--- a/src/main/java/gregtech/loaders/recipe/ComponentRecipes.java
+++ b/src/main/java/gregtech/loaders/recipe/ComponentRecipes.java
@@ -1,0 +1,387 @@
+package gregtech.loaders.recipe;
+
+import gregtech.api.recipes.RecipeMaps;
+import gregtech.api.unification.material.MarkerMaterials;
+import gregtech.api.unification.material.Materials;
+import gregtech.api.unification.ore.OrePrefix;
+import gregtech.common.items.MetaItems;
+import net.minecraftforge.fluids.FluidStack;
+
+import static gregtech.api.GTValues.L;
+
+public class ComponentRecipes {
+
+    private static final FluidStack[] pumpFluids = {Materials.Rubber.getFluid(L), Materials.StyreneButadieneRubber.getFluid((L * 3)/4), Materials.SiliconeRubber.getFluid(L/2) };
+
+    public static void register() {
+
+        //Field Generators start ---------------------------------------------------------------------------------------
+        RecipeMaps.ASSEMBLER_RECIPES.recipeBuilder()
+                .input(OrePrefix.dust, Materials.EnderPearl, 1)
+                .input(OrePrefix.circuit, MarkerMaterials.Tier.Basic, 4)
+                .fluidInputs(Materials.Osmium.getFluid(L * 2))
+                .outputs(MetaItems.FIELD_GENERATOR_LV.getStackForm())
+                .duration(100).EUt(30).buildAndRegister();
+
+        RecipeMaps.ASSEMBLER_RECIPES.recipeBuilder()
+                .input(OrePrefix.dust, Materials.EnderEye, 1)
+                .input(OrePrefix.circuit, MarkerMaterials.Tier.Good, 4)
+                .fluidInputs(Materials.Osmium.getFluid(L * 4))
+                .outputs(MetaItems.FIELD_GENERATOR_MV.getStackForm())
+                .duration(100).EUt(120).buildAndRegister();
+
+        RecipeMaps.ASSEMBLER_RECIPES.recipeBuilder()
+                .inputs(MetaItems.QUANTUM_EYE.getStackForm())
+                .input(OrePrefix.circuit, MarkerMaterials.Tier.Advanced, 4)
+                .fluidInputs(Materials.Osmium.getFluid(L * 8))
+                .outputs(MetaItems.FIELD_GENERATOR_HV.getStackForm())
+                .duration(100).EUt(480).buildAndRegister();
+
+        RecipeMaps.ASSEMBLER_RECIPES.recipeBuilder()
+                .input(OrePrefix.dust, Materials.NetherStar, 1)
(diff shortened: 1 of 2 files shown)


src/main/java/gregtech/loaders/recipe/ComponentRecipes.java
```diff
@@ -0,0 +1,387 @@
...
+
+import static gregtech.api.GTValues.L;
+
+public class ComponentRecipes {
+
+    private static final FluidStack[] pumpFluids = {Materials.Rubber.getFluid(L), Materials.StyreneButadieneRubber.getFluid((L * 3)/4), Materials.SiliconeRubber.getFluid(L/2) };
+
+    public static void register() {
+
+        //Field Generators start ---------------------------------------------------------------------------------------
```
LAGIdiot, line 18: I like start comment telling what is created. Though it should start with `S` to be consistent

src/main/java/gregtech/loaders/recipe/ComponentRecipes.java
```diff
@@ -0,0 +1,387 @@
...
+                .duration(100).EUt(1920).buildAndRegister();
+
+        RecipeMaps.ASSEMBLER_RECIPES.recipeBuilder()
+                .inputs(MetaItems.QUANTUM_STAR.getStackForm())
+                .input(OrePrefix.circuit, MarkerMaterials.Tier.Elite, 4)
+                .fluidInputs(Materials.Osmium.getFluid(L * 32))
+                .outputs(MetaItems.FIELD_GENERATOR_IV.getStackForm())
+                .duration(100).EUt(7680).buildAndRegister();
+
+        //Field Generators End -----------------------------------------------------------------------------------------
```
LAGIdiot, line 54: I don't think we need end comment as it is pretty obvious that if another section is starting previous is ending.

LAGIdiot (changes requested):
This seems like reasonable addition to me. I think that timing and energy consumption is adequate. Also having this in file is really good idea.

There are just some minor things to fix.

LAGIdiot (approved):
Thanks for your hard work!

PR #1634, opened 2021-05-30
Migrate Json Recipes to Java Code

serenibyss:
**What:**
This PR migrates our json-based recipes to Java code, into `CraftingRecipeLoader` with the other crafting table recipes. Personally, I have found it confusing and annoying when looking through the source code for a specific recipe, only to find that it was hiding in the resources folder in a json file. Now all recipes are in Java code, making them all much easier to find and change if needed.

This PR looks like a large amount of code changes, but it is only one file changed, with the rest being json deletions.

**How solved:**
I moved all recipes from json files to Java code, ensuring that the recipes remained completely unchanged, and that the recipe names match exactly. As a result, there should be no noticeable difference to players, addon developers, and modpack authors, except for two things:
- IV Robot Arms were incorrectly using EV circuits. This was changed to IV circuits.
- Pumps of all tiers had json recipes for Rubber, Styrene-Butadiene Rubber, and Silicone Rubber Rings. However, for whatever reason, the Silicone Rubber ring recipes were failing to be added, most likely to an improperly typed name for Silicone Rubber, though I did not investigate the reason very much since it would be migrated to Java anyway. As a result, Pumps of all tiers (LV-IV) now have a recipe to use Silicone Rubber Rings. If this change is undesired, I can remove it.

Otherwise all recipes are identical in name, inputs, and outputs. I verified all recipes manually against the latest release of GTCE, and can guarantee that all recipes (aside from the couple above) are identical.

**Outcome:**
- Move json recipes to Java code
- Add Silicone Rubber Ring Pump recipes
- Fix IV Robot Arm using EV circuit

None expected, as these recipes will appear to have no change to anyone outside of GTCE source code.


LAGIdiot:
This seems reasonable but I would like to leave it for next release. So we get #1633 out first. Also I think we should make separate files for larger groups like molds, coins, components (use that one from #1633 with some modifications).

ALongStringOfNumbers (dismissed):
This looks good to me as is. We can probably wait on moving the shapes/molds recipe and coins recipes to new files until such time as a general reorganizing.

LAGIdiot (approved):

PR #1640, opened 2021-05-30
Boilers fixes

### LINKED ISSUE #1628: [BUG] Large boiler incredible efficiency loose when throttling
**Describe the bug**
Large boiler loose way more efficiency when throttling than what was expected. when comparing 100% and 20% throttle leading to 51% of efficiency expected or 49% of loose we indeed have 86.7% of loose. the main problem is not really about the efficiency loose (even if ohh deer ) but the two fact that are : 

- both boiler run the same amount of time with the same amount of fuel
- the efficiency showed in the GUI is far from reality .

When throttling I'm pretty sure merely everyone thought as me that when reducing the throttle to 20% the output is 20%*maxOutput*efficiencyInGUI and the consumption is 20%*maxConsumption or something close to this like output= 20%*maxOutput and consumption = 20%*maxConsumption + (100% - efficiencyInGUI)*maxConsumption

**Versions**
Forge: 14.23.5.2854
GTCE: 1.14.1.705
_Addons:_ None

**Setup**
Playing Solo
New world generated yes

**Steps To Reproduce**
build two large boiler of same type here two bronze. 
warm them to 100° (optional should not change anything).
fuel them with the same amount of the same fuel.
store all of the steam produced.
see.

**Expected behavior**
the showed efficiency is 51% so it is expected to get roughly 50% less steam with the boiler at 20% over the one at 100%

**Screenshots**
both boiler burned 48B of creosote after achieving 100° each enderium tank hold 500B with a total of 5788.5 B of steam produced at 100% of throttle versus 765.4B at 21% instead of something close to 2900B.

### LINKED ISSUE #1637: [BUG] Large Boiler temperature wiggle
**Describe the bug**
The current implementation of the Large boilers make them cooling and heating again all the time which is frustrating and reduce the effective steam output why the boiler doesn't just keep at max temp when possible ?

**Setup**
Playing Both
_New world generated_ yes

**Steps To Reproduce**
1) build a large boiler any of them
2) make it warm up to max temp
3) see it periodically decrease and increase his temparature

**Expected behavior**
keeping itself at max temperature

**Additional context**
the temparature change is used to consume fuel but the consumption can be done in an other way really easily

galyfray:
**What:**
refactored a little the update method to keep everything as is except not reducing the temperature when at max temp. And removed unwanted throttle efficiency loss.

**How solved:**
Before the update when at max temp the boiler was consuming 1 burn tick every two tick in average this behavior has been kept using `getOffsetTimer%2` as a condition. nothing changed for temperature increase. Decrease append only when there is no more fuel for 20 ticks.

concerning the throttle efficiency I just removed the multiplier in the duration computation. Making the Large boilers consuming less when not at full throttle unlike before. The older behavior was completely useless knowing that the boiler trash it's excess production and was requiring the same amount of fuel no matter how much you decrease the throttle.

**Outcome:**
Fixes : #1637
Fixes : #1628

None expected


warjort:
While you are modifying this, can you also add a markDirty() where it changes the throttle value:
https://github.com/GregTechCE/GregTech/blob/c76e2e3c84479c58e9d5bbbfdaef12da3ca3f2ce/src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityLargeBoiler.java#L206

This will make sure it saves the user's throttle changes (absent another change to the controller causing a save).

galyfray:
for sure but can you explain to me why the current code doesn't save it for sure and what does markDirty because when i'm lokking into the code it seems to me that the throttle is saved correctly and I don't really see what markDirty does.

warjort:
markDirty() is minecraft's way of saying a block needs to be saved.
It adds it to a "queue" of things to be saved the next time the background save timer runs.
In this case, the new throttlePercentage needs to be persisted in the NBT for the tile.

As far as I can tell this handleDisplayClick() method is invoked directly from the network handling code to respond to the client.
So there is no other code that will automatically say the tile has changed and needs saving.

Of course for an active boiler there are likely other changes that are causing the block to marked as dirty.

LAGIdiot:
> Before the update when at max temp the boiler was consuming 1 burn tick every two tick in average
How exactly did you get to this number? I don't see any math that will explicitly show this.

Throttle now looks correctly but I expect that you will provide real numbers and screenshot from testing. Best way would be to do same test as in #1628 so that values are comparable.

galyfray:
>Before the update when at max temp the boiler was consuming 1 burn tick every two tick in average
How exactly did you get to this number? I don't see any math that will explicitly show this.

This indeed is not trivial but not that hard to explain : 
let's assume a basis state were the boiler is at maxTemp-1 and timer%20 = 0
fuel get consumed and the heat increase we are now at maxTemp.
from timer%20 =1 to 19 no burn time is consumed as we are at maxTemp
when timer%20 =0 we still don't consume burn time but we reduce the temparature we are now at maxTemp-1
from timer %20 =1 to 19 we consume one burn time per tick
when timer %20=0 we are return to the first step.

to sum up burn time is consumed every tick for 20 tick and then not consumed at all for 20 tick again and again when we are near maxTemp which give us an average of 1 burn time consumed every two ticks

galyfray:
concerning the test part : 
I have a small problem : the boiler efficiency gets really high (around 150% or more) when throttling while heating up.

this append because I don't touch the burnTime with the throttle anymore and this not a big deal at max temp everything work as expected but not when heating up because the heating doesn't depend on the throttling. I first thought that this won't be a big issue but I wasn't expecting this amount of over efficiency.

I am having a bit of trouble to fix this I have either a big efficiency or inefficiency while the boiler is heating up . I am testing the throttle at maximum temp just for being sure that at least this part work as intended.

galyfray:
and work as intended I have an efficiency of around 55% but I have my tests in really bad conditions better tests will be done later

warjort:
This change modifies the recipe duration calculations in setupRecipeAndConsumeInputs() but it does not modify the corresponding calculations in getFuels() for display in top ?

galyfray:
yes I was planing to fix this on another PR because the current getFuel implementation is already wrong I will fix that after finishing up the throttle

warjort:
I have been trying to understand what this throttle processing in the LargeBoiler actually does and why.

Before this change there has been no major change to the logic since it was first introduced (except for a fix to incorrect rounding), so that doesn't give much insight.

Here is my summary of how I understand it.
Below "normal" will be used for the boiler operating at 100% throttle for a given recipe and boiler type so the equations below can be simplified to remove superfluous details.

Below throttle is a number between 0.2 and 1.0

For non-solid fuels the calculation is:
consumption = normal * throttle
duration = normal * throttle

For solid fuels it is:
consumption = recipe amount
duration = normal / throttle

After these calculations an additional heat efficiency (based on boiler type and how close to max temperature you are)
increases the duration but this is not directly affected by the throttle.

Finally the downside to the throttle is the reduced output. This is

output = normal * throttle * (1 + 0.3 * log(throttle))

This multiplier is a curve passing through approx (0.2, 0.1) and (1, 1) becoming steeper as the throttle is reduced.
Some other approx values for reference (throttle, output) (0.5, 0.4) (0.75, 0.685) (0.9, 0.87)

The first thing to notice is that for non solid fuels the throttle does nothing except reduce your steam output.
The fluid consumption and duration are changed in step meaning your overall fuel consumption doesn't change.

For the solid fuels, it does make the fuel last longer (notice the division instead of multiplication).
So 50% throttle means you get twice the burn time but only get 80% of the steam (twice as long * 40%).
20% throttle is 5 times as long for 50% steam (5 * 10%)

This PR actually flips this behaviour, By removing all the throttle multipliers from *all* the durations the non-solid fuels now last longer on reduced throttle, but the throttle has no affect on solid fuel consumption.

But the big question for me is whether this is correct behaviour?
By just extending the duration (or equivalently just reducing the fuel consumed) you are actually getting more ++temperature changes for the same amount of fuel consumed as you reduce the throttle.
The ++temperature changes are done every 20 ticks (yes it is a bit more complicated than that) and so are proportional to the burn time.

This means it takes a lot less fuel to get the boiler upto max temperature where you will get the better temperature multipliers.
These are the indirect affects I mentioned above with the heat efficiency, but it also applies to the steam output.

warjort:
> the current getFuel implementation is already wrong

It looks like it is not including the getHeatEfficiencyMultiplier() on the duration.
I must have missed this because it is not in the same place as the rest of the calculations.

galyfray:
>It looks like it is not including the getHeatEfficiencyMultiplier() on the duration.
I must have missed this because it is not in the same place as the rest of the calculations.

and it doesn't include the efficiency boost when at max temp

galyfray:
>By just extending the duration (or equivalently just reducing the fuel consumed) you are actually getting more ++temperature changes for the same amount of fuel consumed as you reduce the throttle.

yes I know and I indicate that my first thought was "that should be negligible and should be okay" but it isn't the case and i'm fixing this 

>This PR actually flips this behaviour, By removing all the throttle multipliers from all the durations the non-solid fuels now last longer on reduced throttle, but the throttle has no affect on solid fuel consumption.

yup I didn't see that the implementation was inconsistent between solid and liquid I will fix this to get a nice behavior thanks for pointing this out

galyfray:
>This means it takes a lot less fuel to get the boiler upto max temperature where you will get the better temperature multipliers.
These are the indirect affects I mentioned above with the heat efficiency, but it also applies to the steam output.

all of those different retroactive loop are quite complex and the main reason why I have problems achieving a good balance.

warjort:
> all of those different retroactive loop are quite complex and one of the main reason why I have problems achieving a good balance. the other one is it is difficult to achieve correct experiment.

Would making it check every (20 / throttle) ticks for a ++currentTemperature fix this?
That would have the affect of making the temperature increase more slowly as you would expect on reduced throttle.
It shouldn't affect the max temperature behaviour.

galyfray:
>Would making it check every (20 / throttle) ticks for a ++currentTemperature fix this?
That would have the affect of making the temperature increase more slowly as you would expect on reduced throttle.
It shouldn't affect the max temperature behaviour.

tested and we get a very low efficiency also tested by dividing by the efficiency and we get too much. Also tryed /sqrt(throttle) and we get too much too.

I'm making a small simulator to rapidly try some computation and find a good candidate.

warjort:
> tested and we get a very low efficiency

Did you apply the change to just the ++currentTemperature or to the steam output check as well?

To me it wouldn't be the end of the world if trying to heat up the boiler on reduced throttle is inefficient.
The important part is that it doesn't become more efficient, like your test.

I think it would be acceptable if to get optimal efficiency you should bootstrap your boiler on high throttle and only throttle back when you have it running at high temperature?

galyfray:
>I think it would be acceptable that to get optimal efficiency you should bootstrap your boiler on high throttle and only throttle back when you have it running at high temperature?

I think so still I think not using the optimal way should be viable and it is currently not the case

warjort:
I was wondering what happens if you change the throttle to do something like I suggested for the turbine.

That is instead of the throttle modifying the output directly, it modifies the maximum temperature that can be reached.

This would modify the steam output indirectly because you can't reach higher temperatures and you won't get the better bonuses from the getHeatEfficiency() 

e.g. For the bronze boiler, it would mean 20% throttle only lets you just reach 100 celsius but your fuel would be lasting 5 times longer. i.e. the overall steam output is the "same"(1), it is just produced slower (less steam/tick).
(1) It is not actually the same because of the heat efficiency giving longer burn times at higher temperatures

Something like:
````
diff --git a/src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityLargeBoiler.java b/src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityLargeBoiler.java
index 557ce961..c449804a 100755
--- a/src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityLargeBoiler.java
+++ b/src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityLargeBoiler.java
@@ -207,17 +207,13 @@ public class MetaTileEntityLargeBoiler extends MultiblockWithDisplayBase impleme
     protected void addDisplayText(List<ITextComponent> textList) {
         super.addDisplayText(textList);
         if (isStructureFormed()) {
-            textList.add(new TextComponentTranslation("gregtech.multiblock.large_boiler.temperature", currentTemperature, boilerType.maxTemperature));
+            textList.add(new TextComponentTranslation("gregtech.multiblock.large_boiler.temperature", currentTemperature, boilerType.maxTemperature / getThrottleMultiplier()));
             textList.add(new TextComponentTranslation("gregtech.multiblock.large_boiler.steam_output", lastTickSteamOutput, boilerType.baseSteamOutput));

             ITextComponent heatEffText = new TextComponentTranslation("gregtech.multiblock.large_boiler.heat_efficiency", (int) (getHeatEfficiencyMultiplier() * 100));
             withHoverTextTranslate(heatEffText, "gregtech.multiblock.large_boiler.heat_efficiency.tooltip");
             textList.add(heatEffText);

-            ITextComponent throttleText = new TextComponentTranslation("gregtech.multiblock.large_boiler.throttle", throttlePercentage, (int)(getThrottleEfficiency() * 100));
-            withHoverTextTranslate(throttleText, "gregtech.multiblock.large_boiler.throttle.tooltip");
-            textList.add(throttleText);
-
             ITextComponent buttonText = new TextComponentTranslation("gregtech.multiblock.large_boiler.throttle_modify");
             buttonText.appendText(" ");
             buttonText.appendSibling(withButton(new TextComponentString("[-]"), "sub"));
@@ -242,7 +238,7 @@ public class MetaTileEntityLargeBoiler extends MultiblockWithDisplayBase impleme

     @Override
     protected void updateFormedValid() {
-        if (fuelBurnTicksLeft > 0 && currentTemperature < boilerType.maxTemperature) {
+        if (fuelBurnTicksLeft > 0 && currentTemperature < boilerType.maxTemperature / getThrottleMultiplier()) {
             --this.fuelBurnTicksLeft;
             if (getOffsetTimer() % 20 == 0) {
                 this.currentTemperature++;
@@ -270,7 +266,7 @@ public class MetaTileEntityLargeBoiler extends MultiblockWithDisplayBase impleme
                 }
                 this.hasNoWater = false;
                 if (currentTemperature >= BOILING_TEMPERATURE) {
-                    double outputMultiplier = currentTemperature / (boilerType.maxTemperature * 1.0) * getThrottleMultiplier() * getThrottleEfficiency();
+                    double outputMultiplier = currentTemperature / (boilerType.maxTemperature * 1.0);
                     int steamOutput = (int) (boilerType.baseSteamOutput * outputMultiplier);
                     FluidStack steamStack = ModHandler.getSteam(steamOutput);
                     steamOutputTank.fill(steamStack, true);
@@ -381,10 +377,6 @@ public class MetaTileEntityLargeBoiler extends MultiblockWithDisplayBase impleme
         return throttlePercentage / 100.0;
     }

-    private double getThrottleEfficiency() {
-        return MathHelper.clamp(1.0 + 0.3*Math.log(getThrottleMultiplier()), 0.4, 1.0);
-    }
-
     private void replaceFireboxAsActive(boolean isActive) {
         BlockPos centerPos = getPos().offset(getFrontFacing().getOpposite()).down();
         for (int x = -1; x <= 1; x++) {
````

But this does have the downside that you don't get immediate feedback of a throttle efficiency on the display, instead you see the modified max temperature.

The other more important "but" is that you reach the "max temperature" bonus at a lower temperature.

**Comment by galyfray** [comment]
>Would making it check every (20 / throttle) ticks for a ++currentTemperature fix this? 

thanks to the mall simulation code I've made I can have better view of the overall effectiveness of a throttle method and this one seems to be a good starting point but while it does better in general than the two others solution (the one in the actual state of the PR and the current behavior in the main branch) the values fluctuate a lot.

I found a error in the test setup i have used when checking if this solution was a good one so I'm currently re testing this to see if my simulation is a good approximation and if this solution can be acceptable. 

from what I can see with 24B of creosote it work perfectly

**Comment by galyfray** [comment]
here are the graph produced with the little simulation program : 





due to various rounding the behavior is not as smooth as I would like it to be but the differences are at most 3% and from what I have tested the actual behavior seems smoother than what the simulation produce.

from 24B of creosote the boiler at 100% produced around 404 B of steam while the 20% one produced around 202B of steam using much more time, the difference can  be easily explain by some inaccuracy in my protocol as I manually stop the output when the boiler stop because steam produced while the boiler is cooling down make the measurement wrong.

**Comment by warjort** [comment]
I think that jerkiness you see in the last graph is partly because you are not doing
````
(int) Math.ceil(20 / getThrottleMultiplier())
````
to round up to the nearest integer?

If that really is the cause, you see a more step like graph with the above change.

**Comment by galyfray** [comment]
I have fixed the errors in the getFuel and updated it with the current changes

>(int) Math.ceil(20 / getThrottleMultiplier()) 

ceil doesn't round to the nearest interger but to the smallest integer greater than the value. anyway no matter how you round things in this case it will stay as is because you are not changing the number of step but only where they are

**Comment by warjort** [comment]
> ceil doesn't round to the nearest interger but to the smallest integer greater than the value

I said round UP, which is the same as what you said in different words :-)

**Comment by galyfray** [comment]
>I said round UP, which is the same as what you said in different words :-)

whups missed the up sorry xD

**Comment by ALongStringOfNumbers** [comment]
I was taking a look at this PR and noticed that it dropped some of the changes introduced in #1630, namely the Textures being passed in to the new, non-deprecated enum. These should not be dropped, so the PR needs to be updated to not drop those changes.

File: src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityLargeBoiler.java [hunk 1]
```diff
@@ -242,10 +236,12 @@ private double getHeatEfficiencyMultiplier() {
 
     @Override
     protected void updateFormedValid() {
-        if (fuelBurnTicksLeft > 0 && currentTemperature < boilerType.maxTemperature) {
-            --this.fuelBurnTicksLeft;
-            if (getOffsetTimer() % 20 == 0) {
-                this.currentTemperature++;
+        if (fuelBurnTicksLeft > 0) {
+            if (getOffsetTimer() % 2 == 0 || currentTemperature < boilerType.maxTemperature) {
+                --this.fuelBurnTicksLeft;
```
ALongStringOfNumbers, line 241: Since you changed to `getOffsetTimer() % 2`, don't you need to subtract 2 from the `fuelBurnTicksLeft`, since this is no longer being called every tick, but rather every other tick?

src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityLargeBoiler.java
```diff
@@ -313,7 +309,7 @@ private int setupRecipeAndConsumeInputs() {
                     fluidTank.drain(fuelAmountToConsume, true);
                     long recipeVoltage = FuelRecipeLogic.getTieredVoltage(dieselRecipe.getMinVoltage());
                     int voltageMultiplier = (int) Math.max(1L, recipeVoltage / GTValues.V[GTValues.LV]);
-                    return (int) Math.ceil(dieselRecipe.getDuration() * CONSUMPTION_MULTIPLIER / 2.0 * voltageMultiplier * getThrottleMultiplier());
```
ALongStringOfNumbers, line 316: Why are you removing the throttle multiplier from this value, but not from when the fuel amount to consume is calculated, which is a couple lines above?

src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityLargeBoiler.java
```diff
@@ -72,32 +64,33 @@
...
+                MetaBlocks.BOILER_CASING.getState(BoilerCasingType.STEEL_PIPE),
+                Textures.SOLID_STEEL_CASING,
+                Textures.STEEL_FIREBOX, Textures.STEEL_FIREBOX_ACTIVE),
 
         TITANIUM(3700, 3.0f, 31, 2000,
-            MetaBlocks.METAL_CASING.getState(MetalCasingType.TITANIUM_STABLE),
-            MetaBlocks.BOILER_FIREBOX_CASING.getState(FireboxCasingType.TITANIUM_FIREBOX),
-            MetaBlocks.BOILER_CASING.getState(BoilerCasingType.TITANIUM_PIPE),
-            Textures.STABLE_TITANIUM_CASING,
-            Textures.TITANIUM_FIREBOX, Textures.TITANIUM_FIREBOX_ACTIVE, Textures.LARGE_TITANIUM_BOILER),
```
ALongStringOfNumbers, line 93: The removing of the provided texture for the boilers should be reverted

src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityLargeBoiler.java
```diff
@@ -562,14 +556,13 @@ public boolean onRightClick(EntityPlayer playerIn, EnumHand hand, EnumFacing fac
         }
         for (int slotIndex = 0; slotIndex < itemImportInventory.getSlots(); slotIndex++) {
             ItemStack itemStack = itemImportInventory.getStackInSlot(slotIndex);
-            int burnTime = (int) Math.ceil(TileEntityFurnace.getItemBurnTime(itemStack) / (50.0 * boilerType.fuelConsumptionMultiplier * getThrottleMultiplier()));
+            int burnTime = (int) Math.ceil(TileEntityFurnace.getItemBurnTime(itemStack) / (50.0 * boilerType.fuelConsumptionMultiplier)) * getHeatEfficiencyMultiplier() * currentTemperature == boilerType.maxTemperature ? 2 : 1;
```
ALongStringOfNumbers, line 559: Why the change away from the Throttle Multiplier to determine the burn time, and why does the fuel burn longer when at maximum temperature?

src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityLargeBoiler.java
galyfray: If I remember well that's the main part of the bug the throttle was applied too many times thus acting more like a secondary efficiency than like a throttle

src/main/java/gregtech/common/metatileentities/multi/MetaTileEntityLargeBoiler.java
galyfray: Why does the fuel burn longer at max temp ? I don't know it worked like that before I changed how the boilers works ...
If I am not wrong this code is acting on the the one probe integration so that's just a fix of the showed burn time not taking this behavior in account explaining why I removed the throttle from the computation as I changed this part of how the throttle works

galyfray:
I won't work on this anymore so I close the PR

PR #1645, opened 2021-06-01
Fix wire and cable lang

TechLord22:
**What:**
This PR fixes an error in the en_us lang file. The words wire and cable for their respective items previously started with a lowercase letter, inconsistent with nearly all other lang entries. This PR corrects this error by changing the words to use capital letters for their first letter.

**How solved:**
This was solved by editing the corresponding values in the en_us lang file.

**Outcome:**
Wires and Cables have capital letters in their name types.

None are expected.


DStrand1 (approved):

Exaxxion:
Seems reasonable to me. From the description it sounds like the other lang files already do this (as applicable), and item names in the en_us locale are generally in Title Case.

Should be entirely cosmetic, no risk of breaking anything else, and it's a small change set.

PR #1651, opened 2021-06-02
Swap RECIPE_MAPS to protected rather than private

bruberu:
**What:**
This PR is a one-line change that makes RECIPE_MAPS in RecipeMap protected. Before, if one wanted to create a custom JEI wrapper for a recipe map that extended this, they would have no way to prevent the default GTCE wrapper from being put into a category.

**How solved:**
This problem is easily solved by changing the access of RECIPE_MAPS to protected, which allows for subclasses to delete themselves from that list, thus preventing those default wrappers and categories from being registered in the first place.

**Outcome:**
Swaps RECIPE_MAPS instance variable to a protected, rather than private status.

Since this particular variable wasn't even accessible before, it would be literally impossible for this to create any compatibility issues in the first place.


DStrand1 (changes requested):
I suggest reworking this PR somewhat, and keeping that List private, so that we can guarantee that it will have all RecipeMaps in it, but instead having an `isHidden` field that is checked when adding the JEI category instead. That way we can guarantee that the List has all maps, and still achieve what you desire, and even more if we ever want a fully hidden RecipeMap for whatever reason.

bruberu:
I have just addressed those changes, by making said field `isHidden` that is checked before creating a default RecipeMapCategory. I have also created another constructor for RecipeMap to allow for immediately accessing it, while sustaining the original to keep compatibility. I also set the access back to private, and so the outcome should technically now be "RecipeMaps may now have their categories hidden in JEI" instead.

src/main/java/gregtech/api/recipes/RecipeMap.java
```diff
@@ -59,6 +59,7 @@
     private final TByteObjectMap<TextureArea> slotOverlays;
     protected TextureArea progressBarTexture;
     protected MoveType moveType;
+    public boolean isHidden = false;
```
DStrand1, line 62: Since this is just a constructor parameter, it may as well be private

DStrand1 (approved):
Just one little thing to change, otherwise this PR is good to me.

Also, the force push I did was to cleanup git hierarchy issues at the request of @bruberu.

src/main/java/gregtech/api/recipes/RecipeMap.java
DStrand1: final*, apologies

DStrand1 (approved):

DStrand1:
Git hierarchy will need to cleaned up again before merge, which I can do later if need be.

bruberu:
I think I'm going to have to make a new fork, unfortunately, as well as a new PR in order to resolve the git issues.

PR #1657, opened 2021-06-04
Fix electric tools Taking damage when charged

htmlcsjs:
**What:**
This PR fixes electric tools taking damage if they have charge 

**How solved:**
I implemented a check to only apply damage to tools when the have no charge or are not electric.

**Outcome:**
Fix for electric tools taking damage when charged

None to my knolage


Archengius:
Electric tools are designed to consume both power and durability, but they have 10x more durability than normal tools at the same time. It's not a bug.

idcppl:
That not fully accurate, that is accurate for let's say, LV tools. But as you go to HV tools they are extremely inefficient. Like an HV Wrench is only 5 times as much durability.

Issue #1585, opened 2021-04-26
[BUG] Shadows on Coil Blocks


warjort:
**Describe the bug**
I noticed when looking at #1583 that the shadows for coil blocks look wrong

For other blocks the bright part is 
front: top left 
top: front right

For coil blocks the bright part is kind of opposite but not quite
front: bottom
top: back

I haven't checked other faces.

**Versions**
Forge: 14.23.5.87
GTCE: 1.14.0

**Expected behavior**
The shadows should look consistent.

I don't know what the correct fix should be.
Change the textures?
Apply some uv transformations in blockstates/wire_coil.json?

**Screenshots**
Some casings and tile entities showing the issue
![2021-04-26_14 33 47](https://user-images.githubusercontent.com/275258/116091472-c11dd980-a69c-11eb-815e-62e738fe0c8a.png)

Comparison with apatite from forestry
![2021-04-26_14 41 45](https://user-images.githubusercontent.com/275258/116092212-82d4ea00-a69d-11eb-8b0e-cff8352f55d5.png)

DStrand1:
There is no sort of "fancy render" stuff going on with these, it is just shading applied to the texture of the block itself. So I think just updating the texture would be the proper fix if we decide this needs to be addressed

PR #1658, opened 2021-06-05
new coils texture

galyfray:
**What:**
Fixes: #1585  
gonna be honest previous coils wasn't looking at all like coils 

**How solved:**
created new textures inspired by the F32 patch

**Outcome:**
Fixes: #1585

**Additional info:**

here is a link to the [base file](https://cdn.discordapp.com/attachments/763408591126265867/850699509184200734/coils.xcf) I used to produce the coils. this is a gimp file and to create new coils simply choose use color>colorize on the layer with the coil inner texture choose your color and export . I put it here so addons creator can use it if they want or in case we would like to add new coils in the future. You may use the file as you want as long as you tell from were you took it.


warjort:
Don't you think it looks weird with the background to the coils being a different color to the rest of the multiblock?

I think the reason why these textures were full blocks is because they can be used with MBs of different color.
e.g. the pyrolase oven or cracker unit.

If you want it to look more like a coil, perhaps you can do something like this?
https://ftb.fandom.com/wiki/Copper_Coil_Block

galyfray:
>Don't you think it looks weird with the background to the coils being a different color to the rest of the multiblock?

no i don't I think this looks perfectly fine

LAGIdiot:
How does this look on Pyrolysis oven?

galyfray:
it looks quite okay on pyrolyse oven and has the same problem than the older coils : we can't change the top orientation making it looking a bit weird.

ALongStringOfNumbers:
Personally, I think that changing the coils like this should be done with a texturepack, especially since we are not likely to get a unified agreement on a new coil texture. These coils look, to me at least, pretty bad, but I am sure that other think that they look fine. There are several existing texture packs that change the looks of the coils, if how they currently look in GTCE does not look good (subjectively) to a player.

galyfray:
>These coils look, to me at least, pretty bad, but I am sure that other think that they look fine. 

yes I do. but can you explain why you dislike them ? I personally don't like how the top looks.

>Personally, I think that changing the coils like this should be done with a texturepack, especially since we are not likely to get a unified agreement on a new coil texture.

So far the current coils texture are (I think most of the people will agree) pretty bad looking and have a different shadow than other blocks added by GTCE which is why I'm proposing this as a PR. The main idea of this PR is to discuss and propose textures that everyone would agree on :
- it looks like a GTCE texture
- it is better than the older texture

and I think it is possible that a large amount of people agree on both points.

warjort:
Besides the issue I raised with the background texture color, 
your texture also appears to have something of a Moire pattern.
https://en.wikipedia.org/wiki/Moir%C3%A9_pattern
at least in the top blast furnace image, my eyes want to interpret the coils as moving?

ALongStringOfNumbers:
> yes I do. but can you explain why you dislike them ? I personally don't like how the top looks.

I don't like the pastel shades of the colors that are used for the different coils, and I was never too fond of the diagonal stripe look present on the coils. 
In addition, I do not like the metal seam running between the upper and lower coils on the ebf, instead I prefer the current coils which don't have the seam in the middle.

That last point might be something that is able to be addressed through CTM, although I am unsure if that would make CTM a hard dependency for GTCE.

Getting everyone to agree on a texture that satisfies everyone is a pretty hard thing to do, which is why I recommended this be solved by resourcepacks

galyfray:
>at least in the top blast furnace image, my eyes want to interpret the coils as moving?
that's an interesting issue

>I don't like the pastel shades of the colors that are used for the different coils, and I was never too fond of the diagonal stripe look present on the coils. 

I tried to reuse has much as possible the original shade of the coils and colors looks really clone on my main monitor and not and my second I will try to find a way of fixing the colors

>In addition, I do not like the metal seam running between the upper and lower coils on the ebf, instead I prefer the current coils which don't have the seam in the middle.

seems like you are not the only one disliking this ^^

I will work on new texture in grey scale to validate the pattern and then colorize them to validate the colors used.

warjort:
> That last point might be something that is able to be addressed through CTM, although I am unsure if that would make CTM a hard dependency for GTCE.

I am no expert, but I think CTM just requires adding a
block.pgn.mcmeta
to describe alternate CTM textures/metadata.

But I don't know if/how that works if you have custom rendering.

TechLord22:
These textures have numerous issues in my eyes. First of all, the textures are too "flat." The colors are very plain and don't blend well into each other. They also honestly just don't look very good, especially in comparison to the textures in other variants of GT (namely forks of GT5u). 

I personally don't mind the metal bands running across the middle, but they are too jagged and also do not fit and blend well with the rest of the coil colors.

Furthermore, the coils do not vertically connect. The diagonals don't continue into the next when moving up or down a Y value, shown in this screenshot.
image

Frankly, I do not think GTCE needs any new textures for coils or really anything else, and I honestly think any and all coil retexturing should be left to resourcepacks.

In reply to @ALongStringOfNumbers: CTM textures would not make CTM a hard GTCE dependency, and are only optional png texture files used according to an MCMeta file, only loaded when CTM or Optifine is present. I have specifically done CTM work with coils for a resourcepack around a month ago. It is not very difficult, and is no different from using CTM with any other block.

warjort:
> Frankly, I do not think GTCE needs any new textures for coils or really anything else, and I honestly think any and all coil retexturing should be left to resourcepacks.

The issue I originally raised on the bug report is that the shadows are wrong in the current textures.

PR #1662, opened 2021-06-07
[Tool]Multi Assembler

Yefancy:
**What:**
A lot of people think `in world preview` from MTB and Gregicality should be pulled into GTCE. 
_Because Gregicality's `Freedom Wrench` could rotate multiple blocks freely, in order to reduce compatibility problems, I make this PR. (It's essentially only adding the necessary code about preview, but leaving places for gregicality to do coremode easily)_

Multi Assembler is a new tool that helps players build multi block. It has the following featrues:

1. Preview---Show a preview of the multi blocks machine
2. Perspective---Show the perspective preview of the outline (Useful when you have blocks blocking your view).
3. Debug---Shows the position that error occurred when checking the blockpattern.
4. Building---Fast build (no consumption in creative mode)

**How solved:**
`MultiAssemblerBehaviour`: item using logic.
`BlockPatternChecker`: help check the blockpattern.
`CPacketMultiBlockStructure`: client packet, send the blocks that should be built to the server.
`MultiHelperRenderer`: preview renderer.
`GTJeiOptional`: check JEI installed.

**Outcome:**
Preview:
preview

Perspective:
perspective

Debug:
debug

Building:
building

Recipe:
recipe

**Additional info:**
This feature now relies on JEI. The pattern in JEI are required for reference except debug mode. 
I considered separating the pattern from the JEI integration to support native multi-block hinting. But that required a lot of code modification, so I gave up. (It's also not necessary. I don't think anyone who plays GTCE without JEI.)

**Compatibility related:**
There is a variable in the code, `spin`, which is always `Enumfacing.North` in GTCE. It will be modified in Gregicality. In order to reduce hacking work, I reserved it.


src/main/java/gregtech/api/multiblock/BlockPatternChecker.java
```diff
@@ -0,0 +1,49 @@
...
+import gregtech.api.metatileentity.multiblock.MultiblockControllerBase;
+import net.minecraft.util.EnumFacing;
+import net.minecraft.util.math.BlockPos;
+
+public class BlockPatternChecker {
+    
+    public static BlockPos getPatternErrorPos(MultiblockControllerBase controllerBase) {
+        BlockPattern structurePattern = controllerBase.getStructurePattern();
+        if(structurePattern != null && structurePattern.checkPatternAt(controllerBase.getWorld(), controllerBase.getPos(), controllerBase.getFrontFacing().getOpposite()) == null) {
+            return new BlockPos(structurePattern.getBlockPos());
```
warjort, line 13: Why didn't you make BlockPattern do this convert to immutable?

It seems like a bad idea to expose that mutable state from BlockPattern for others to corrupt.

The same goes for structureDir (the elements will be mutable).

src/main/java/gregtech/api/net/CPacketMultiBlockStructure.java
```diff
@@ -0,0 +1,141 @@
...
+                (buf) -> {
+                    int dim = buf.readVarInt();
+                    List<ItemStack> map = new ArrayList<>();
+                    List<BlockInfo> blockInfos = new ArrayList<>();
+                    int size = buf.readVarInt();
+                    try {
+                        for (int i = 0; i < size; i++) {
+                            map.add(buf.readItemStack());
+                        }
+                    } catch (Exception ignored) {
```
warjort, line 50: Shouldn't there be some logging here instead of silently eating exceptions?
Or are these always expected for some reason producing too much noise?

src/main/java/gregtech/common/render/MultiHelperRenderer.java
```diff
@@ -0,0 +1,419 @@
...
+            }
+
+            GlStateManager.popMatrix();
+            ForgeHooksClient.setRenderLayer(oldLayer);
+
+            GlStateManager.glEndList();
+            GlStateManager.enableLighting();
+            if (mte != null) {
+                try{
+                    ObfuscationReflectionHelper.findMethod(MultiblockControllerBase.class, "checkStructurePattern", Void.TYPE).invoke(mte);
```
warjort, line 278: You can update this not to use reflection now?

src/main/java/gregtech/common/items/behaviors/MultiAssemblerBehaviour.java
```diff
@@ -0,0 +1,128 @@
...
+        NBTTagCompound nbt = player.getHeldItem(hand).getOrCreateSubCompound("GT.Detrav");
+        byte mode = nbt.hasKey("mode") ? nbt.getByte("mode") : 0;
+        if (pos != null && world.getTileEntity(pos) instanceof MetaTileEntityHolder) {
+            MetaTileEntity mte = ((MetaTileEntityHolder) world.getTileEntity(pos)).getMetaTileEntity();
+            if (mte instanceof MultiblockControllerBase && !player.isSneaking()) { // preview
+                if (world.isRemote) { // client side only
+                    if(!Loader.isModLoaded("jei")) { // check jei installed
+                        player.sendMessage(new TextComponentTranslation("metaitem.multi_assembler.jei_missing"));
+                    } else if (mode == 3) { // building mode
+                        WorldSceneRenderer worldSceneRenderer = GTJeiOptional.getWorldSceneRenderer((MultiblockControllerBase) mte);
```
warjort, line 46: Won't this crash when JEI is not installed? The annotation you used removes the method from the class when JEI is not installed.

It would be better to have an interface (that doesn't depend on JEI) that the JEI integration can implement and initialise a static instance in ClientProxy.

src/main/java/gregtech/api/multiblock/BlockPatternChecker.java
```diff
@@ -0,0 +1,49 @@
...
+        }
+        return null;
+    }
+
+    public static PatternMatchContext checkPatternAt(MultiblockControllerBase controllerBase) {
+        BlockPattern structurePattern = controllerBase.getStructurePattern();
+        return structurePattern == null ? null : structurePattern.checkPatternAt(controllerBase.getWorld(), controllerBase.getPos(), controllerBase.getFrontFacing().getOpposite());
+    }
+
+    public static EnumFacing getSpin(MultiblockControllerBase controllerBase) {
```
warjort, line 23: Not sure how you are going to override a static method? :-)

warjort:
This looks good, I haven't tried it, just skimmed the code.

Have you tried this with MultiblockTweaker? I know they have a similar feature, is it compatible with that?

They were also having problems with modded transparent blocks in the preview. Have you tried that?
https://github.com/eutro/MultiblockTweaker/issues/5

src/main/java/gregtech/api/render/MetaTileEntityTESR.java
```diff
@@ -125,10 +125,21 @@ public void renderTileEntityFast(MetaTileEntityHolder te, double x, double y, do
 
     @Override
     public boolean isGlobalRenderer(MetaTileEntityHolder te) {
-        if(te instanceof IRenderMetaTileEntity) {
-            return ((IRenderMetaTileEntity) te).isGlobalRenderer();
-        } else if(te instanceof IFastRenderMetaTileEntity) {
-            return ((IFastRenderMetaTileEntity) te).isGlobalRenderer();
+        if (te instanceof IRenderMetaTileEntity) {
```
warjort, line 128: I think you need to resync your working branch with master.
This change is already applied in #1563

src/main/java/gregtech/api/multiblock/BlockPatternChecker.java
Yefancy: I copied most of the code from Gregicality, where I created a new object to record it so it didn't have this problem there. However, you are right, I simplified the implementation, which is a real potential bug. The reason I wrote it outside the BlockPattern is that I want to keep the function name available in Gregicality. ill convert it to draft and make it more elegant

src/main/java/gregtech/api/multiblock/BlockPatternChecker.java
Yefancy: i was going to write a coremod directly . If you don't want too much abstraction in GTCE.

src/main/java/gregtech/common/items/behaviors/MultiAssemblerBehaviour.java
Yefancy: I thought about the second point, but I tried to avoid making too many changes to the original code.

src/main/java/gregtech/api/render/MetaTileEntityTESR.java
Yefancy: well, my negligence

src/main/java/gregtech/common/items/behaviors/MultiAssemblerBehaviour.java
warjort: I would recommed creating an interface called something like MultiblockRenderSource 
that the JEI integration can implement when it is present.

Yefancy: good idea

src/main/java/gregtech/api/multiblock/BlockPatternChecker.java
warjort: An abstraction that is understandable for future maintainers is better than a coremod hack. :-)

I guess this can just be an extra method in MultiblockControllerBase for people to override.

Yefancy: yeah, I'd be happy to do that too.

Yefancy:
After a day of thinking about it, I wondered if it was worth it to create an interface for multi controller to provid the multi info. JEI integration can check it and automatically create InfoPages for each multiblock.

I think it is nice, because I don't think structure info is used only in JEI. However, I should also consider compatibility with addons that register JEI using original way.  In order to be compatible with the multiblocks provided by addons, it may still need to check the JEI pattern.

In order not to rework again and again, I wonder what you think about: @LAGIdiot @DStrand1 

1. Is this PR worth to be done? 
2. Is it worthwhile to separate MulitiBlockInfoPge from the JEI integration? 
3. Do I need to keep compatibility if I want to modify existing framework?

If there is a better way, could guide me, or push it directly to this PR.

Issue #1588, opened 2021-05-03
[Suggestion] Machine Configuration Copy/Paste Item


Caedis:
**Is your feature request related to a problem? Please describe.**
Getting into the end game of Omnifactory requires the use of many solidifiers with a creative fluid tank and setting them each up manually can be very repetitive  

**Describe the solution you'd like**
Crouch+UseItem = Copy machine config (output side and fluid/item output toggle)  
UseItem = Paste machine config from copied machine  
Optional: Have config copy item in off hand to auto paste the config to the machine when placed.

**Additional context**
Think something like the Conduit Probe from Ender IO or Red Print from Thermal Foundation

LAGIdiot:
This sounds like interesting idea and I am in favor of it.

I think we should discuss this before we get to implementation as this may not be easy as it sounds.

Also we could use this tool for copying configurations to covers too. But for that we would need more then one mod on it.

galyfray:
I think the step to make this possible is to define what is a configuration and what is not a config the second step would to create a class (or a family of class) which will handle those configuration inside it something which would work a bit like an NBT tag compound

warjort:
A generic implementation of this would need to get the NBT for the tile.

But that tag contains more than just the configuration. 
It contains the state of the machine (including inventories, energy, recipe progress etc.) and also its attached covers.

So that does not seem doable in general.

An alternate approach would be to have a capability, something like:
````
public interface IConfigurable {
    // Used to match the tile entity id so configruations are not applied to the wrong machines
    ResourceLocation getId();
    // Gets the configuration subset of the tile NBT
    NBTCompoundTag getConfiguration();
    // Applies the configuration subset
    void applyConfiguration(NBTCompoundTag configuration);
}
````
It needs to be NBT because that is the only way to store additional information within the item stack of the new tool.

Once the basic plumbing has been added to MetaTileEntity, AbstractRecipeLogic and other places that have standard configuration (or pass through to something that does), then machines can "opt in" to the process.
e.g. SimpleMachine would be an easy candidate

For covers presumably a separate "screwdriver" version would be used? 
This would just extend the main tool with logic to redirect its operation to the cover like the screwdriver does.
Or maybe it can be a different mode of the same tool?

But covers have the additional complication that some of the config is stored within optional items added to the cover,
e.g. filters. They would need additional logic to detect this and indicate the problem to the user (chat message?).

[...]

PR #1663, opened 2021-06-07
Prototype configuration tool

warjort:
**What:**
This is a prototype configuration tool for #1588

**How solved:**
Added 2 new capabilities
IConfigurable implemented by machines and covers that  gets and sets a subset of the NBT that is just the configuration.
IConfigurator a tool that can retrieve this config from one machine/cover and apply it to another

**Outcome:**
WIP

**Additional info:**
This has been implemented as "opt in", each individual machine and cover has to expose the IConfigurable capability.
This is because each machine/cover needs to decide what is configuration and what is other state.

For now the SimpleMachine and Conveyor cover (including item filters) have been implemented.
NOTE: item filters as covers themselves have not been implemented (though it would be easy to do).
It is only a subset so people can try it without me having to unpick everything if people don't like it.

The configurator tool is actually the FILE (this is because it is only a prototype and the file has no other real use).
It has 2 modes, you change it (like similar tools in other mods) by pointing away from blocks and right clicking.

COVER - this is the normal mode that respects the machine grid and lets you configure covers
MACHINE - this ignores the grid and lets you see through covers - it is mainly for people who are having difficulty finding somewhere to click to change their machine

Some issues:
1) This is all server side. How to store something in the configurator nbt so something can be displayed in its tooltip about the configuration stored (and it would also be useful for context in chat/error messages)? You can use /ct hand to see what is stored in the item but thats not user friendly :-)
2) I left some comments in the code about edge cases - e.g. what happens if a machine says it is configurable but then returns null for its configuration - for now it is just ignored with a message in chat
3) Should the real configurator take durability - for now it does - maybe it should be an electric item?
4) I haven't attempted to do the magic wand/auto item filter switch stuff asked for by eutro
5) LongStringNumbers suggested (4) should be on an advanced version of the item - having a gui that lets you select from multiple saved configurations could also be an advanced feature?
6) The normal (de)serializeNBT is not quite the same api as this new copy/pasteConfiguration - the new api includes validation and a player so that errors/warnings can be reported - this makes reuse between the two harder


src/main/java/gregtech/api/metatileentity/MTETrait.java
```diff
@@ -39,6 +43,26 @@ public NBTTagCompound serializeNBT() {
...
     }
 
+    @SuppressWarnings("static-method")
+    protected boolean isConfigurable() {
+        return false;
+    }
+
+    @Override
+    public ResourceLocation getConfigurationID() {
+        throw new AssertionError("unused");
```
eutro, line 53: Wouldn't it be better to just not implement `IConfigurable` for classes that... don't implement it?

warjort: I did it this way, so it is easy for me to the refactor while it is being developed. 

I don't intend for MTETrait or the item filter handler to really implement that interface since they don't expose the capability directly.

warjort:
I have done an implementation for points 4/5 above.

Rather than try to guess/define what advanced features should be enabled, I have done this by making the ConfigurationContext passed into the IConfigurable api an ICapabilityProvider.
This means the "caller" can decide what advanced features to enable by providing a suitable capability.

For the configurator item this has been implemented as proxying the player capabilities when it is the advanced version with the  basic version having no capabilities.

Enabling this for the item is as simple as:
````
addComponents(new ConfiguratorItemStat().advanced());
````

I also made the isAdvanced() available on the ConfigurationContext and IConfiguratorItem in case somebody wants to do some advanced behaviour that does not neatly map to a  capability or a capability would be overkill.

warjort:
With the above changes, the "magic wand" behaviour now works for an advanced configurator.

The ItemFilterContainer will now check if there is an IItemHandler capability and use that as the inventory to check to obtain filters when the one in the cover is wrong (or place an unwanted filter into it if no filter is required in the new configuration).

I also made these features usable outside the IConfigurable api. 
Although that is of limited use for the swapFilter() method since other methods to actually configure the new filter are not public.

warjort:
I am marking this as ready for review, even though there are obviously missing pieces like;

* the real configurator item
* updating other machines/covers to work with the new api

These require feedback before proceeding.

PR #1664, opened 2021-06-10
Rework Chargers, add inventory charging capability

ALongStringOfNumbers:
**What:**
This PR reworks chargers to make them a bit more viable, and also adds in an inventory charging feature to the chargers, in an attempt to address #542.

**How solved:**
Reworks the chargers so that they will pull up to 4 amps at a time (Depending on the number of batteries in the charger), and split the charge evenly across all batteries in the charger.

For the wireless charging, a 2 block space in front of the front face of the charger has been transformed into a wireless charging zone. When in this zone, `IElectricItems` in the players inventory will be charged, taking evenly from all batteries in the charger

**Outcome:**
Reworking the Charger and adds minimal wireless charging to it

**Additional info:**
screenshots and other information to come when the PR is finalized

This will be fine for addons if they are not overriding part of the charger logic


warjort:
> For the wireless charging, a 2 block space in front of the front face of the charger has been transformed into a wireless charging zone. When in this zone, IElectricItems in the players inventory will be charged, taking evenly from all batteries in the charger

Wouldn't it be more immerisive to implement this as a "charge pad" that could be connected anywhere on the EnergyNet?
The charge pad responding to acceptEnergyFromNetwork() by checking for players standing on top it.

I also think for "real" wireless charging, a cover (placeable on a cable) would be better fit?
This would allow a tiered system where the range of the charging can be increased as you progess.

warjort:
For the player charging itself, I think this would be better implemented using the AttachCapabilitiesEvent
to make a player an IEnergyContainer?
This would make other methods to charge the player easier to do.

src/main/java/gregtech/common/metatileentities/electric/MetaTileEntityCharger.java
```diff
@@ -87,18 +90,274 @@ protected void initializeInventory() {
...
+        BlockPos endPos;
+        //If not facing up or down, add an additional translation, to translate the charging region to be centered on a block,
+        //rather than centered between blocks
+        if (!(frontFacing == EnumFacing.UP || frontFacing == EnumFacing.DOWN)) {
+            endPos = this.getPos().offset(frontFacing, 2).add(0, 1, 0).add(0, 0, 1);
+        } else {
+            endPos = this.getPos().offset(frontFacing, 2);
+        }
+        //Create a list of all players within the AABB created by the charge position and the end position of the charging region
+        List<EntityPlayer> playersAbove = this.getWorld().getEntitiesWithinAABB(EntityPlayer.class, new AxisAlignedBB(this.getPos(), endPos));
```
warjort, line 133: I think you need to use Predicate version of the getEntitiesWithinAABB() here to check to see if the player is dead?

The player entity can still exist until the user clicks "respawn".

warjort:
I am trying to understand how you reworked the charging.

In the old method, the acceptEnergyFromNetwork() charged the internal buffer.
Then the tick charged the items.

You seem to be charging the items in both places now?

PR #1665, opened 2021-06-11
Fix name overlaps in GUIs

### LINKED ISSUE #1223: [BUG] Assembler Machine name overlaps with the mahine slots
**Describe the bug**
The machine name for the assembling machines overlaps the top slots of the assembly machine, somewhat obscuring what is present in those slots.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.10.0.546

**Setup**
Playing Solo (only tested solo)
_New world generated_ if applicable: yes

**Steps To Reproduce**
1) Cheat in and place an Assembling Machine of any tier 
2) Ope the GUI and notice that the machine name overlaps with the top row of input slots

**Expected behavior**
The machine name does not overlap with the input slots

**Screenshots**
overlay

TechLord22:
**What:**
This PR changes the way machine GUI templates are created so that a machine's name does not overlap its slots in the GUI, making UIs more clean and readable for players.

**How solved:**
Overloaded methods were added to `SimpleMachineMetaTileEntity`, `RecipeMap`, `ModularUI`, and `RecipeMapGroupOutput` in order to offset all of a given machine's UI information lower down on an extended background, and create adequate space for the machine's name to be clearly displayed. This y-offset is exactly 9 units, as that is the standard height of the minecraft font in the `FontRenderer` class. The fix is only applied if a machine has more than 6 input or output slots of any type, as that would result in a 3 or more rows of slots and thus a name overlap. The original UI size and GUI positions is left alone for any machine that does not require an adjustment, and any other methods utilizing this functionality will not need to be modified.

**Outcome:**
Machine names will no longer overlap with other GUI elements. Fixes #1223.

**Additional info:**
Old, Overlapping Assembler GUI:
image

New, Non-Overlapping Assembler GUI: 
image

I do not expect any compatibility issues. All method changes were done in the overloaded versions, and the existing methods now utilize their counterparts in a manner that does not change their functionality.

### DIFF
--- a/src/main/java/gregtech/api/gui/ModularUI.java
+++ b/src/main/java/gregtech/api/gui/ModularUI.java
@@ -90,7 +90,12 @@ public final class ModularUI implements ISizeProvider {
     }
 
     public static Builder defaultBuilder() {
-        return new Builder(GuiTextures.BACKGROUND, 176, 166);
+        return defaultBuilder(0);
+    }
+
+    //to be called in order to change the gui height by a specific amount
+    public static Builder defaultBuilder(int yOffset) {
+        return new Builder(GuiTextures.BACKGROUND, 176, 166 + yOffset);
     }
 
     public static Builder borderedBuilder() {
@@ -186,6 +191,11 @@ public final class ModularUI implements ISizeProvider {
 
         public Builder bindPlayerInventory(InventoryPlayer inventoryPlayer, TextureArea imageLocation) {
             return bindPlayerInventory(inventoryPlayer, imageLocation, 8, 84);
+        }
+
+        //to be called in order to offset the player inventory from the top of the window
+        public Builder bindPlayerInventory(InventoryPlayer inventoryPlayer, TextureArea imageLocation, int yOffset) {
+            return bindPlayerInventory(inventoryPlayer, imageLocation, 8, 84 + yOffset);
         }
 
         public Builder bindPlayerInventory(InventoryPlayer inventoryPlayer, TextureArea imageLocation, int x, int y) {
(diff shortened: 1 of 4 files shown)


ALongStringOfNumbers:
Could we get a picture of the before GUI please?

src/main/java/gregtech/api/gui/ModularUI.java
```diff
@@ -188,6 +193,11 @@ public Builder bindPlayerInventory(InventoryPlayer inventoryPlayer, TextureArea
             return bindPlayerInventory(inventoryPlayer, imageLocation, 8, 84);
         }
 
+        //to be called in order to offset the player inventory from the top of the window
+        public Builder bindPlayerInventory(InventoryPlayer inventoryPlayer, TextureArea imageLocation, int yOffset) {
+            return bindPlayerInventory(inventoryPlayer, imageLocation, 8, 84 + yOffset);
+        }
```
ALongStringOfNumbers, line 199: To be consistent, the non offset version of this method should redirect to the offset version of the method, to match the changes to `defaultBuilder`

src/main/java/gregtech/api/recipes/machines/RecipeMapGroupOutput.java
```diff
@@ -40,6 +40,11 @@ public Builder createUITemplate(DoubleSupplier progressSupplier, IItemHandlerMod
         return builder;
     }
 
+    @Override
+    public Builder createUITemplate(DoubleSupplier progressSupplier, IItemHandlerModifiable importItems, IItemHandlerModifiable exportItems, FluidTankList importFluids, FluidTankList exportFluids, int yOffset) {
+        return createUITemplate(progressSupplier, importItems, exportItems, importFluids, exportFluids);
+    }
+
```
ALongStringOfNumbers, line 47: Isn't this backwards? Shouldn't the non offset one redirect to the offset method with an offset of zero?

ALongStringOfNumbers (commented):
Looks mostly good, just a few comments

ALongStringOfNumbers (approved):

PR #1666, opened 2021-06-12
Fix Surface Block generation issue, add logging

ALongStringOfNumbers:
**What:**
Fixes some bad surface block generation logic added in #1619. In addition, breaks up the conditional statement for better readability and for documentation. In addition, adds logging for when all surface rocks/surface blocks for a vein fail to generate

**How solved:**
Fixed the conditional check, and broke up the conditional checks for better readability

**Outcome:**
Fixed Surface Block/Rock Spawning logic. Add log for if all generation attempts fail for Surface Rock/Block.

None expected

### DIFF
--- a/src/main/java/gregtech/api/worldgen/populator/SurfaceRockPopulator.java
+++ b/src/main/java/gregtech/api/worldgen/populator/SurfaceRockPopulator.java
@@ -4,6 +4,7 @@ import com.google.gson.JsonObject;
 import gregtech.api.unification.OreDictUnifier;
 import gregtech.api.unification.material.type.Material;
 import gregtech.api.unification.stack.UnificationEntry;
+import gregtech.api.util.GTLog;
 import gregtech.api.worldgen.config.OreConfigUtils;
 import gregtech.api.worldgen.config.OreDepositDefinition;
 import gregtech.api.worldgen.generator.GridEntryInfo;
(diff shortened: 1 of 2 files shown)


src/main/java/gregtech/api/worldgen/populator/SurfaceRockPopulator.java
```diff
@@ -84,12 +99,32 @@ public void populateChunk(World world, int chunkX, int chunkZ, Random random, Or
...
-                    setStoneBlock(world, topBlockPos, undergroundMaterials);
+
+                //Checks if there is an air block for the surface rock indicator to spawn at
+                if(!world.isAirBlock(topBlockPos)) {
+                    continue;
                 }
+
+                //Checks if the block is a replaceable feature like grass or snow layers. Liquids are replaceable, so
+                // exclude one deep liquid blocks, for looks
+                if(!blockAtPos.isReplaceable(world, topBlockPos) || world.getBlockState(topBlockPos).getMaterial().isLiquid()) {
```
DStrand1, line 110: Shouldn't this `isReplacable()` check be done above instead of `isAirBlock()`, so that surface rocks will be able to spawn in place of Tall Grass, Snow, etc.?

ALongStringOfNumbers: I removed the isAirBlock check entirely, as it was redundant as the Air block is replaceable.

DStrand1 (approved):

PR #1668, opened 2021-06-12
Fix Ore generation Page glitches

### LINKED ISSUE #1669: [BUG] Issue with the Ore Generation Page and Advanced Rocketry
**Describe the bug**
Ore veins set to generate on AR planets do not show the dimension in the list of dimensions that the ore will generate in.

Another minor thing, Some veins are shown that they can generate in the AR "Space" dimension. I believe that this happens with veins that are set to spawn in the overworld. It should be checked if anything can even spawn in the "Space" dimension (Are asteroids a separate dimension?) and if not, then the "Space" dimension information should not be shown.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.15.1
_Modpack:_ Non3
_Addons:_ None
AR: 2.0.0-250
LibVulpes: 0.4.2-83

**Setup**
Playing Solo
_New world generated_ Yes

**Steps To Reproduce**
1) Add a vein to generate specifically in an AR dimension (dimension ids found in planetDefs.xml in the world save folder)
2) See that the dimension is not listed on the ore generation page with the vein

**Expected behavior**
For the dimension to show on the ore generation page

**Additional context**
While debugging another issue, I found that this was because at the time of loading, the registered dimensions only included the vanilla dimensions and the AR "Space" dimension. There was no list of the various planet dimensions anywhere in the list of registered dimensions, causing no dimension to be displayed

ALongStringOfNumbers:
**What:**
This PR has some slight fixes to the Ore Generation page, mostly rendering, but also how the named dimensions were stored and accessed.

The main rendering issue was when the fifth row of the ore list would be used, text would be rendered under the slot
java_2021-06-11_22-57-00

Adds support for detecting Advanced Rocketry dimensions and displaying them on the ore generation page.

**How solved:**
Adds an offset to the y position stored value, to account for the stored value being the top of the slot instead of the bottom of the slot.

Moves the dimension collection information to later on in loading, and adds support for gathering AR specific dimension information.

**Outcome:**
Fixed a rendering issue on the Ore Generation Page when the veins had large numbers of unique ores.
Adds support for displaying Advanced Rocketry dimension information. Closes #1669.

**Additional info:**
Fixed rendering for large ore veins:
java_2021-06-11_23-12-18

If any addons were using the public getter for the named dimensions, they will have to switch the order Map, but this is a minor issue as it would only affect a part of the JEI page, and I am not aware of any addons that were using this feature.

Due to moving where the dimension information was captured, some accessors are no longer available.

### DIFF
--- a/src/main/java/gregtech/api/GTValues.java
+++ b/src/main/java/gregtech/api/GTValues.java
@@ -75,7 +75,8 @@ public class GTValues {
         MODID_CT = "crafttweaker",
         MODID_TOP = "theoneprobe",
         MODID_CTM = "ctm",
-        MODID_CC = "cubicchunks";
+        MODID_CC = "cubicchunks",
+        MODID_AR = "advancedrocketry";
 
     //because forge is too fucking retarded to cache results or at least do not create fucking
     //immutable collections every time you retrieve indexed mod list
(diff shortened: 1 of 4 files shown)


ALongStringOfNumbers:
With the latest commit, I have moved when the dimension information was collected, and also added in support for the display of Advanced Rocketry dimensions.

An example can be seen here:
java_2021-06-15_15-53-22

Dimension 2 was the moon dimension in Advanced Rocketry's planetDef.xml, and I locally added an entry for it in the dimensions.json file

LAGIdiot (approved):
This change seems reasonable to me.
Because of changes on API side I will mark it as Minor

PR #1671, opened 2021-06-16
Properly localize Aluminium

Testure:
**What:**
This PR fixes a spelling issue in the en_us.lang file. Aluminium is spelled as "Aluminum" in the US and Canada but is spelled as "Aluminium" in the rest of the world. Which may cause confusion for players in the US/Canada when searching in JEI.

**How solved:**
This was solved by editing all references to "Aluminium" in the en_us.lang file and adding the en_gb.lang and en_au.lang files.

**Outcome:**
Aluminium material is spelled as "Aluminum" in the en_us.lang file and is spelled as "Aluminium" in the en_gb.lang and en_au.lang files.

None expected.


DStrand1:
Thanks for submitting this PR.

Personally, I am not in favor of this PR. First and foremost, changing its localization in `en_us` means that the most-often used locale will no longer match the Ore Dictionary name for the material, and changing this is not something we can do lightly, as it would potentially remove any Aluminium items in current worlds due to how our materials are generated. If we leave the Ore Dictionary entry as separate from the new localization (which we would), then I think that this would cause more confusion for players, wondering why its suddenly called something else. At least with how we have it currently, it is fully consistent.

I will leave this for other members to weigh in, but I think the issues this change raises are more of a problem than the current localization.

warjort:
This wikipedia entry explains the history of the spelling:
https://en.wikipedia.org/wiki/Aluminium#Etymology

10,000 ft view:
Aluminium is the correct scientific spelling. 
While the American spelling is an acceptable alternative.

Given both are correct, I don't see the need to change it and cause the confusion referred to by @DStrand1.
If you strongly disagree, you can always use a resource pack in your local installation/mod pack to modify it.

eutro:
I would agree that there's no need to change it.

PR #1672, opened 2021-06-18
Update systems to new Recipe Properties

### LINKED ISSUE #1614: [BUG] Implosion Compressor No Longer Stores Explosives as Property
**Describe the bug**
Since the addition of the new recipe properties in #1580, the implosion compressor now does not store its explosives as a property for the recipe. You can see [here in Gregicality](https://github.com/Gregicality/gregicality/blob/d1883fed1778ffec09215275ab93a8abcf1aac14/src/main/java/gregicadditions/recipes/categories/handlers/ElectricImplosionHandler.java#L59) that we are depending on this property for some of our recipe generation. As a result it is impossible to discern what is an explosive and what is an input for the Implosion Compressor. This code works for any version prior to 1.15.0.

**Versions**
Forge: 14.23.5.2847
GTCE: 1.15.0
_Addons:_ None required, but tested with Gregicality

**Expected behavior**
Implosion Compressor should store explosives as a recipe property as well as a recipe input

ALongStringOfNumbers:
**What:**
This PR aims to update existing systems to the new Recipe Property system introduced in #1580.
As mentioned in #1614, after the Recipe Property rework the implosion compressor was not properly adding its properties to the list of Recipe Properties, which is one of the motivations of this PR.

This PR also adds an `isHidden` method to the Recipe Properties, which will control if the recipe property should display its key and value on the JEI page. This defaults false to keep current behavior.

Please let me know if there are any systems that I have forgotten to update to the new system, and I will update those. I want to update all systems in one go, so that there are not any PRs that have to be made later.

**How solved:**

### Implosion Compressor
Creates the `ImplosionExplosiveProperty`. The Recipe Property type for this property was chosen to be ItemStack, to easier incorporate any explosives added to the Implosion Compressor via crafttweaker, as having just an Integer in this case would not work correctly.

Removes the override of `buildAndRegister` in `ImplosionRecipeBuilder` and moves the logic that was in the method into `build`. This was done due to the explosive itemstack being null for some reason in the `build` method when it was not null before `build` was called. In addition, this brings the Builder inline with other builders as they do not override `buildAndRegister`.

This recipe property was set to hidden, as it does not make sense to see the information in JEI when it can be seen in the recipe.

### Fusion Reactor
Creates the `FusionEUToStartProperty` and uses it in the `FusionRecipeBuilder` class.

**Outcome:**
Updates systems to new Recipe Property system. Fixes #1614.

I do not expect there to be any. Although `buildAndRegister` was removed in `ImplosionRecipeBuilder` I am not aware of any addons that were using this field. It may be safer to make the override just call `super` instead, but I will leave that up to feedback.

### DIFF
--- a/src/main/resources/assets/gregtech/lang/en_us.lang
+++ b/src/main/resources/assets/gregtech/lang/en_us.lang
@@ -2887,6 +2887,7 @@ gregtech.recipe.amperage=Amperage: %,d
 gregtech.recipe.not_consumed=Not consumed in process
 gregtech.recipe.chance=Chance: %s%% +%s%%/tier
 gregtech.recipe.blast_furnace_temperature=Temperature: %,dK (%s)
+gregtech.recipe.explosive=Explosive: %s
 gregtech.recipe.eu_to_start=Energy To Start: %,dEU
 
 gregtech.fluid.click_to_fill=§7Click with a empty fluid container to fill it from tank.
--- a/src/main/java/gregtech/api/recipes/builders/FusionRecipeBuilder.java
+++ b/src/main/java/gregtech/api/recipes/builders/FusionRecipeBuilder.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableMap;
 import gregtech.api.recipes.Recipe;
 import gregtech.api.recipes.RecipeBuilder;
 import gregtech.api.recipes.RecipeMap;
+import gregtech.api.recipes.recipeproperties.FusionEUToStartProperty;
 import gregtech.api.util.EnumValidationResult;
 import gregtech.api.util.GTLog;
 import gregtech.api.util.ValidationResult;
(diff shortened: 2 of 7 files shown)


src/main/resources/assets/gregtech/lang/en_us.lang
```diff
@@ -2887,6 +2887,7 @@ gregtech.recipe.amperage=Amperage: %,d
 gregtech.recipe.not_consumed=Not consumed in process
 gregtech.recipe.chance=Chance: %s%% +%s%%/tier
 gregtech.recipe.blast_furnace_temperature=Temperature: %,dK (%s)
+gregtech.recipe.explosive=Explosive: %s
```
DStrand1, line 2890: If we are setting the property as hidden, do we need this translation?

DStrand1 (approved):
LGTM, only one small suggestion

src/main/resources/assets/gregtech/lang/en_us.lang
ALongStringOfNumbers: The property can technically have its hidden state removed, so I figured it would be better to be safe than have an untranslated string show up in JEI.

Issue #1567, opened 2021-04-09
[Suggestion] ToolProgression and Crafttweaker modification of harvest levels


AvilionAMillion:
**Is your feature request related to a problem? Please describe.**
As it stands, the gregtech ores do not have changeable harvest levels, this is in contrast to most other ores in other mods, and, in my personal case, completely breaks progression possibilites for modpack makers. I may be mistaken and am overlooking a way to change GT ore harvest levels, and in that case my apologies, however currently it would be incredibly helpful to have them be able to be changed

**Describe the solution you'd like**
I don't know much about Java, but hopefully the solution would be as simple as asking for the harvest values before placing its own in. The outcome should be the same regardless, a way of modifying the harvest levels of GT blocks.

warjort:
I am no expert on CraftTweaker but can't you just use this?
https://docs.blamejared.com/1.12/en/Vanilla/Blocks/IBlockDefinition/#set-the-blocks-harvest-level

GTCE doesn't set the harvest level of its ores, which I believe means minecraftforge sets them to level 2 (iron)

warjort:
In GTCE the harvest level is defined on the material.
For some reason this property is read only:

````
    /**
     * Tool level needed to harvest block of this material
     */
    @ZenProperty
    public final int harvestLevel;
````

The "final" indicates it can't be changed via CraftTweaker.

I guess there is some reason why changing the harvest level of a material dynamically is a bad idea, but I don't know what that is?

One alternative is to just generate your own materials for progression.
You can set the harvest level of any materials you create:
https://docs.blamejared.com/1.12/en/Mods/GregTechCE/Material/

But obviously that would mean doing some work to change the recipes and ore generation to use your materials.
I know OmniFactory and FTB Interactions does this to some extent.

AvilionAMillion:
I did know that adding new materials and changing their harvest levels was an option, however, considering the large amount of changes I was going to make to the whole tool progression system it would effectively require a complete restructuring of every ingot in the game, so I wasn't planning on doing that

If there is any good reason as to why the harvestLevel is final then I'd like to know, if not, it seems weird that it's there

warjort:
I believe this answers your question:

https://github.com/GregTechCE/GregTech/pull/1568#issuecomment-817241916

PR #1673, opened 2021-06-19
Allow the harvest level of ores to be changed

hjae78:
**What:**
Recently I've seen some questions about how to change the harvest level of GTCE ores. Since the field was final this was not possible.

**How solved:**
I solved this by removing the final modifier of the `harvestLevel` field.

**Outcome:**
Modpack developers can now change the harvest level of DustMaterials via ZenScript.

### DIFF
--- a/src/main/java/gregtech/api/unification/material/type/DustMaterial.java
+++ b/src/main/java/gregtech/api/unification/material/type/DustMaterial.java
@@ -95,7 +95,7 @@ public class DustMaterial extends FluidMaterial {
      * Tool level needed to harvest block of this material
      */
     @ZenProperty
-    public final int harvestLevel;
+    public int harvestLevel;
 
     /**
      * Material to which smelting of this material ore will result


warjort:
See #1567, #1568 in particular:
https://github.com/GregTechCE/GregTech/pull/1568#issuecomment-817241916

warjort:
I interpret that comment as you need to prove via an analysis (including for addons) that allowing users to change a material's property doesn't break something, like recipe generation.

hjae78:
The only place this property is used in any sort of recipe is [here](https://github.com/GregTechCE/GregTech/blob/master/src/main/java/gregtech/api/items/toolitem/ToolMetaItemListener.java#L64). Obviously changing this field during the GTCE CT loading stage wouldn't result in any unexpected results here

LAGIdiot:
This is not that easy change as it may feel. Harvest level is fundamental and it is used on many places even though it may not look like it. For example by changing it you can break progression through mod. Also we would have to solve/accept some problems like change on existing playthrough and NBT of tools. There probably many other places which I did not cover by this quick look through.

PR #1683, opened 2021-06-30
Fix integer overflow for total fuel burn time

### LINKED ISSUE #1682: [BUG] Integer overflow for total fuel burn time
**Describe the bug**
See Gregicality/Gregicality#547

It's possible for the total burn time of fuels to exceed Integer.MAX_VALUE (or about 3.4 years)

**Versions**
Forge: N/A
GTCE: 1.16.0
_Addons:_ Reported against the Gregicality Naquadah Reactor Multiblock

**Expected behavior**
No Integer overflow

warjort:
**What:**
Total fuel burn time has an integer overflow when total fuel burn time exceeds Integer.MAX_VALUE.

**How solved:**
Changed the code to use a long;
* When calculating total burn time
* Storing the burn time
* Displaying the burn time

**Outcome:**
An integer overflow can now only occur at Long.MAX_VALUE
Closes #1682

I don't think there are other implementations or users of IFuelInfo, but just in case;
* I have left the old integer methods in the code and deprecated them.
* I made the new long method in IFuelInfo a default method that calls the old int method.
* The int version to get the burn time in AbstractFuelInfo now returns Integer.MAX_VALUE if the long exceeds that value. 

NOTE: This is just for backwards compatibility, none of the GTCE code now uses the int methods.


warjort:
I have left the fuel remaining and fuel capacity as ints.
If these were to ever overflow an int there would be other more important problems with the real runtime handling of the fuels.

ALongStringOfNumbers (approved):

PR #1692, opened 2021-07-09
Fix time on Yttrium Barium Cuprate recipe

ALongStringOfNumbers:
**What:**
Fixes the time on the Yttrium Barium Cuprate mixer recipe. According to @DStrand1 he made a typo when adding the recipe and accidentally typed an 8000 tick duration instead of an 800 tick duration.

**How solved:**
Reduces the time on the recipe

**Outcome:**
Fixes Yttrium Barium Cuprate taking a long time to make in the mixer

Only for addons or modpack makers that touched this recipe, although it will be a quick fix for them


DStrand1 (approved):

Exaxxion (approved):
Clearly an oversight, was meant to be 40s at ULV, not 400s.

PR #1693, opened 2021-07-10
Change Fusion Property to Long

ALongStringOfNumbers:
**What:**
This change is done for cleaner implementation in addon mods, as GTCE currently does not implement anything related to the Fusion Reactor besides providing its Property and Recipe Builder.

The implementation in addon mods (SoG, Gregicality) relies upon totaling the energy stored in a number of `IEnergyContainer`s, which store their energy value as a long. Therefore it would be required to compare the values of a long (Energy stored in the number of `IEnergyContainer`s) and an int (The Fusion Property).

This change simplifies things for addons somewhat, by instead storing the property as a Long.

**How solved:**
Changes the Property storage and the Recipe Builder to Store and Create recipes with a long property instead of an int

**Outcome:**
Changes Fusion Property to a long

There may be minor compatibility issues with addons as they are attempting to gather the incorrect value from property storage, although the issues can be fixed quickly by changing the type of value that is retrieved.

### DIFF
--- a/src/main/java/gregtech/api/recipes/builders/FusionRecipeBuilder.java
+++ b/src/main/java/gregtech/api/recipes/builders/FusionRecipeBuilder.java
@@ -12,14 +12,14 @@ import org.apache.commons.lang3.builder.ToStringBuilder;
 
 public class FusionRecipeBuilder extends RecipeBuilder<FusionRecipeBuilder> {
 
-    private int EUToStart;
+    private long EUToStart;
 
     public FusionRecipeBuilder() {
     }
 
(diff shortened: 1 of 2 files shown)


Exaxxion (approved):
Code changes look fine to me. That I know of, SoG and possibly Gregicality would be the only mods affected, and as stated it's pretty trivial to adapt to.

This is a desirable change for us SoG devs as the consistent types will remove the need for the casting we currently have to do for the mixed types.

DStrand1 (approved):

PR #1706, opened 2021-08-08
Fix GT tools not working for some mod actions

### LINKED ISSUE #1704: [BUG] Cannot chop down trees with GTCE axes
**Describe the bug**
You can't break trees from the DynamicTrees TFC mod

**Versions**
Forge: 2855
GTCE: 1.17.0
_Modpack:_ TerraFirmaGreg-1.12.2-R2.7.5
_Addons:_ -

**Setup** + **Steps To Reproduce**
1. Install both mods or install modpack.
2. Enter in the world.
3. Get yourself a GT axe..
4. Try to break tree.
5. Nothing happenned.

I also found a problem on GitHub Dynamic trees TFC, but they say that this is a GTCE problem.
https://github.com/Gaelmare/dynamictreestfc/issues/31

serenibyss:
**What:**
This PR fixes GT tools not working in some cases for other mods. For example as detailed in #1704, trees from the Dynamic Trees mod cannot be chopped with a GT Axe or Saw.

**How solved:**
I overrode `getToolClasses` in `ToolMetaItem`, and specified tool classes in each tool type as applicable.

**Outcome:**
Closes #1704

**Additional info:**
As we are now marking "extra" tools as their proper classes, it is possible that additional compatibility with other mods will achieved. Particularly our Wrench and Crowbar.

### DIFF
--- a/src/main/java/gregtech/common/tools/ToolHardHammer.java
+++ b/src/main/java/gregtech/common/tools/ToolHardHammer.java
@@ -15,9 +15,15 @@ import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
 
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 public class ToolHardHammer extends ToolBase {
+
+    private static final Set<String> HAMMER_TOOL_CLASSES = new HashSet<String>() {{
+        add("hammer"); add("pickaxe");
+    }};
 
     @Override
     public boolean canApplyEnchantment(ItemStack stack, Enchantment enchantment) {
@@ -86,5 +92,10 @@ public class ToolHardHammer extends ToolBase {
     @Override
     public void addInformation(ItemStack stack, List<String> lines, boolean isAdvanced) {
         lines.add(I18n.format("metaitem.tool.tooltip.hammer.extra_drop"));
+    }
+
+    @Override
+    public Set<String> getToolClasses(ItemStack stack) {
+        return HAMMER_TOOL_CLASSES;
     }
 }
--- a/src/main/java/gregtech/common/tools/ToolJackHammer.java
+++ b/src/main/java/gregtech/common/tools/ToolJackHammer.java
@@ -21,11 +21,13 @@ import net.minecraft.util.math.RayTraceResult;
 import net.minecraft.world.World;
 import net.minecraftforge.common.util.FakePlayer;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 
 public class ToolJackHammer extends ToolDrillLV {
+
+    private static final Set<String> HAMMER_TOOL_CLASSES = new HashSet<String>() {{
+        add("pickaxe"); add("hammer");
+    }};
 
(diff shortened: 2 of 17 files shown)


src/main/java/gregtech/common/tools/ToolHardHammer.java
```diff
@@ -15,10 +15,16 @@
...
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 public class ToolHardHammer extends ToolBase {
 
+    private static final Set<String> HAMMER_TOOL_CLASSES = new HashSet<String>() {{
+        add("hammer"); add("pickaxe");
+    }};
```
Exaxxion, line 26: Probably better to use `ImmutableSet.of("hammer", "pickaxe")` or similar here. If the Set itself is mutable then a caller could modify it, and the double-brace approach will also create an anonymous inner class which is probably not desirable.

src/main/java/gregtech/common/tools/ToolJackHammer.java
```diff
@@ -21,12 +21,14 @@
...
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 
 public class ToolJackHammer extends ToolDrillLV {
 
+    private static final Set<String> HAMMER_TOOL_CLASSES = new HashSet<String>() {{
+        add("pickaxe"); add("hammer");
+    }};
```
Exaxxion, line 30: Same thing as other comment, perhaps also should be called `JACKHAMMER_[...]` rather than `HAMMER_[...]`

Exaxxion:
Besides the review comments and some more cases of the same issues, the code looks fine.

src/main/java/gregtech/common/tools/ToolJackHammer.java
warjort: It uses "hammer" in canMineBlock, so it should be consistent:
https://github.com/GregTechCE/GregTech/blob/d9e7ed4570b990ed389a6a96589636ced78edd34/src/main/java/gregtech/common/tools/ToolJackHammer.java#L117

warjort: All these if statements could be changed to something like:
````
return HAMMER_TOOL_CLASSES.contains(tool) ||
````
It will be slightly less efficient, but easier to maintain in future.
i.e. you won't have subtle bugs introduced by inconsistencies between getToolClasses() and canMineBlock()

Exaxxion: > It uses "hammer" in canMineBlock, so it should be consistent:

I clearly meant the name of the constant, not the value of the strings in the Set.

> All these if statements could be changed to something like:

Not sure what this has to do with this block of code; perhaps create a separate "conversation" linked to the code you're talking about so we have the appropriate context.

src/main/java/gregtech/common/tools/ToolHardHammer.java
LAGIdiot: I also think these should be immutable from our side to keep them safe. Regarding double braces they have to be addressed too.

PR #1709, opened 2021-08-14
Move fetching dimensions to catch late registrations

ALongStringOfNumbers:
**What:**
This PR moves the dimension fetching for the ore generation page to catch late registrations. This should expand the number of dimensions that the ore generation page can work across, while at the same time dropping the specialized logic to catch stuff from AR.

**How solved:**
Moves the call to find the registered dimensions to when the page is opened. When debugging, I noticed that the declaration at the top of the class did not contain all registered dimensions, due to how late some dimensions were registered. Moving fetching the registered dimensions to when the page is opened ensures that all dimensions will be registered.

**Outcome:**
Improves dimension fetching of Ore Generation Page

None expected

### DIFF
--- a/src/main/java/gregtech/integration/jei/GTOreCategory.java
+++ b/src/main/java/gregtech/integration/jei/GTOreCategory.java
@@ -35,7 +35,6 @@ public class GTOreCategory extends PrimitiveRecipeCategory<GTOreInfo, GTOreInfo>
     protected List<Integer> dimensionIDs;
     protected final int FONT_HEIGHT = Minecraft.getMinecraft().fontRenderer.FONT_HEIGHT;
     protected final Map<Integer, String> namedDimensions = WorldGenRegistry.getNamedDimensions();
-    private Map<DimensionType, IntSortedSet> dimMap =  DimensionManager.getRegisteredDimensions();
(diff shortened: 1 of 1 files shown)


Exaxxion:
Is there a particular reason for the change to filtering of AR dimensions by type?

If this operation is costly, it may also make sense to cache the result rather than recomputing it every time the page is opened. However, if we anticipate that new dimensions with ores could be registered after the player is able to interact with JEI, then it would not be safe to assume that we can cache the result.

ALongStringOfNumbers:
The changes to the AR block were in fact me moving away from filtering the AR dimensions by type, as previously I had hardcoded AR compat. This was because the AR dimensions were not in the populated list of registered dimensions.

However, with the move of fetching the registered dimensions to when the page opens, the AR dimensions are properly populated, and so I removed the fetching of AR dimensions by type. The only one I left was the dimension for "Space" which is the dimension AR adds for its space stations, and can have nothing spawning it in. Previously I was removing this dimension from the page, due to the reason that nothing can spawn in it, so I just left this removal in.

As for mods adding new dimensions at runtime, there is RFTools, which does such a thing, so I am not sure if caching the Map will be 100% reliable.

Exaxxion (approved):

LAGIdiot (approved):

PR #1710, opened 2021-08-15
Add warning to controller if Turbine is obstructed

ALongStringOfNumbers:
**What:**
Adds a warning to the controller GUI of the Turbine multiblocks if the front facing is obstructed. I have seen players confused on why their turbine was not functioning, and so I though a helpful warning would be useful.

**How solved:**
Checks if the front face is free, and adds a warning if it is not

**Outcome:**
Adds a warning to Turbines if their front face is obstructed

**Additional info:**
java_4NHVbA4y6h

None


serenibyss (approved):

Exaxxion:
This sounds like a useful enhancement as yeah, we get the occasional confused player we eventually determine through screenshots has set up his turbine incorrectly.

It might be clearer to state "turbine rotor obstructed" or "air intake obstructed" as the current wording is a little ambiguous. Otherwise, looks fine to me.

ALongStringOfNumbers:
> It might be clearer to state "turbine rotor obstructed" or "air intake obstructed" as the current wording is a little ambiguous. Otherwise, looks fine to me.

The reason I did not say "Turbine Rotor Obstructed" is because that might make players think that only the rotor holder needs to be clear, which is not the case for the turbines

LAGIdiot (approved):

PR #1713, opened 2021-08-25
Fix multi dimensions

### LINKED ISSUE #1712: [BUG] Multiple dimensions specified by ID in a world gen json do not work
**Describe the bug**
When specifying multiple dimensions in a world gen json file using IDs, it only works with the first listed dimension. Dimension ranges still work but separate dimensions do not.

**Versions**
Forge: 14.23.5.2855
GTCE: 1.17
Addons: None

**Steps To Reproduce**
1) Specify multiple dimensions with the dimension filter in a ore generation json like so:
`"dimension_filter": ["dimension_id:0", "dimension_id:1", "dimension_id:-1"],`
You can use any dimension ID those are just the ones I used. Dimension names are mapped in dimensions.json.
2) View in-game and see that ores do not spawn in dimensions past the first one and the JEI ore spawn information also does not show the ore spawning in dimensions past the first one.

**Expected behavior**
Each dimension ID listed generates the relevant ores and JEI ore spawn information notes it correctly too.

Synthitic:
**What:**
Fixes declaring multiple dimensions separately in a worldgen json file.

**How solved:**
Use only 1 return to prevent skipping over all other dimensions.

**Outcome:**
Fixes #1712 and also multiple dimensions being specified by name since it was also not working.
One can also declare dimensions in different ways consecutively:
`"dimension_filter": ["dimension_id:0","dimension_id:1","dimension_id:-1"],` works
`"dimension_filter": ["dimension_id:-1","name:the_end","name:*(overworld)"],` also works
`"dimension_filter": ["dimension_id:0:1","dimension_id:-1"],` also works

**Additional info:**
There may be a better way to do this. I found this to be the fastest way to fix this.

None so far after testing in different ways.

### DIFF
--- a/src/main/java/gregtech/api/worldgen/config/WorldConfigUtils.java
+++ b/src/main/java/gregtech/api/worldgen/config/WorldConfigUtils.java
@@ -36,18 +36,18 @@ public class WorldConfigUtils {
                 allPredicates.add(wp -> wp.isNether() || wp.getDimensionType() == DimensionType.NETHER);
                 continue;
             }
-            Function<WorldProvider, String> stringSupplier;
+            Function<WorldProvider, String> stringSupplier = null;
             if (stringValue.startsWith("dimension_id:")) {
                 String filterValue = stringValue.substring(13);
                 if (filterValue.indexOf(':') == -1) {
                     int dimensionId = Integer.parseInt(filterValue);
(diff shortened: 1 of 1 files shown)


LAGIdiot (approved):
This change looks reasonable to me.

PR #1716, opened 2021-09-06
Fix Fluid Regulator Keep Exact

### LINKED ISSUE #1705: [BUG] Fluid Regulators not respecting Keep Exact
Fluid Regulators appear to be ignoring their mode setting and just transferring the maximum amount per second they're allowed to.

**Versions**
GTCE: d9e7ed45 via `./gradlew runClient`

**Setup**
Playing Solo (probably doesn't matter though), new superflat world created

**Steps To Reproduce**
1. Place down an autoclave next to a quantum tank
2. Place a Fluid Regulator on the tank on the side adjacent to the autoclave
3. Set the Fluid Regulator to Export, Flow Rate 1mB/s, Keep Exact: 2mB
4. Fill the quantum tank with some palladium
5. Observe that 1mB of palladium is entering the autoclave each second and it will continue until all of the palladium has entered the autoclave.

**Expected behavior**
The fluid regulator will move 1mB of palladium into the autoclave then stop transferring until the autoclave has no palladium left in its fluid slot.

**Screenshots**
image
image

ALongStringOfNumbers:
**What:**
Fixes Fluid Regulators failing to follow Keep Exact mode as described in #1705 

**How solved:**
Previously, the destination fluid was failing to be populated, due to a failure during the `drain` call. This would leave `destFluid` as null, meaning that `amountToDrainAndFill` would always be populated, meaning that fluid would always be transferred. 

To fix this, I have changed the logic somewhat, initializing the amount to drain and fill to the amount it would be if no fluid was found in the destination inventory.

I then check the destination inventory to see if there are any fluids that match the fluid in the Source Inventory. If there is, the destination fluid is updated, and the amount to drain and fill is updated based on the amount of fluid in the destination tank. 

Note: I did cap `amountToDrainAndFill` to `0` here on the lower end, with a new call of `max`. This was because I was seeing negative values during debugging.

After checking for a matching fluid, I then check if the Fluid amount in the destination already exceeds the desired amount.

As noted in the comment, I considered putting a `break` after finding a valid destination fluid, but that could have a weird edge case if there was a Fluid Inventory with two tanks of the exact same fluid in the destination inventory. However, since fluid handlers are recipe based, I deemed this a very small chance edge case, but let me know your thoughts. I am leaning towards adding a `break` after finding a valid fluid.

**Outcome:**
Fix Fluid Regulator "Keep Exact" mode. Closes #1705 

None expected

### DIFF
--- a/src/main/java/gregtech/common/covers/CoverFluidRegulator.java
+++ b/src/main/java/gregtech/common/covers/CoverFluidRegulator.java
@@ -89,7 +89,8 @@ public class CoverFluidRegulator extends CoverPump {
                 if(destProperties.getContents() != null && destProperties.getContents().isFluidEqual(sourceFluid)) {
                     destFluid = destProperties.getContents();
                     amountToDrainAndFill = Math.min(Math.max(0, keepAmount - destFluid.amount), fluidLeftToTransfer);
-                    // Should we break here? If we do, we will only allow interaction with the first tank found,
+                    break;
+                    // Breaking here will only allow interaction with the first tank found,
                     // which could hit the edge case of having the same fluid in multiple tanks. However, this will be
                     // a rare edge case, because Fluid Handlers are limited by recipe.
                 }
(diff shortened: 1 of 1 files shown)


LAGIdiot:
Thanks for you hard work. I will rather use #1717 as it is more comprehensive.

PR #1717, opened 2021-09-15
New algorithm for Fluid Regulator Keep Exact Mode

### LINKED ISSUE #1705: [BUG] Fluid Regulators not respecting Keep Exact

Exaxxion:
Fixes #1705, Supersedes #1716 

**What:**
The existing Keep Exact algorithm was buggy and the proposed fix by ALSON - while it seemed to function fine - was hard to follow the logic due to the way it was written to work without substantially changing the original implementation.

**How solved:**
I came up with a more straightforward algorithm to completely replace the existing one, in the hopes that it would be easier to understand and maintain.

I have also created some tests to demonstrate the functionality of the revised algorithm and ensure that the behavior is the same as before, besides the bug.

**Outcome:**
Fluid Regulator Keep Exact mode works properly and the algorithm is simpler so it should be easier to verify and maintain.

As currently proposed, a misbehaving fluid tank implementation may cause an exception to intentionally be thrown; this will crash the game but is useful for diagnostic purposes during mod development and is not anticipated to be encountered when interacting with properly implemented code from other mods.

### DIFF
--- a/src/main/java/gregtech/common/covers/CoverFluidRegulator.java
+++ b/src/main/java/gregtech/common/covers/CoverFluidRegulator.java
@@ -190,7 +190,6 @@ public class CoverFluidRegulator extends CoverPump {
               .map(IFluidTankProperties::getContents)
               .filter(Objects::nonNull)
               .filter(fluidTypeFilter)
-              .filter(Objects::nonNull) // objects can and have been null here, shut up IntelliJ
               .forEach(fs -> {
                   summedFluids.putIfAbsent(fs, 0);
                   summedFluids.computeIfPresent(fs, (k,v) -> v + fs.amount);
--- a/src/main/java/gregtech/common/covers/CoverFluidRegulator.java
+++ b/src/main/java/gregtech/common/covers/CoverFluidRegulator.java
@@ -17,8 +17,9 @@ import net.minecraft.util.text.event.HoverEvent;
 import net.minecraftforge.fluids.FluidStack;
 import net.minecraftforge.fluids.capability.IFluidHandler;
 import net.minecraftforge.fluids.capability.IFluidTankProperties;
+import org.apache.logging.log4j.message.FormattedMessage;
 
-import java.util.List;
+import java.util.*;
 import java.util.function.Predicate;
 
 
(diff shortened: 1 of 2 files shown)


src/main/java/gregtech/common/covers/CoverFluidRegulator.java
```diff
@@ -73,32 +74,129 @@ protected int doTransferExact(int transferLimit, IFluidHandler sourceHandler, IF
...
+                                                     Predicate<IFluidTankProperties> tankTypeFilter,
+                                                     Predicate<FluidStack> fluidTypeFilter) {
+
+        final Map<FluidStack, Integer> summedFluids = new HashMap<>();
+        Arrays.stream(handler.getTankProperties())
+              .filter(tankTypeFilter)
+              .map(IFluidTankProperties::getContents)
+              .filter(Objects::nonNull)
+              .filter(fluidTypeFilter)
+              .filter(Objects::nonNull) // objects can and have been null here, shut up IntelliJ
```
PrototypeTrousers, line 193: filter(Objects::nonNull) is being called twice. guess you really hate your nulls (:

Exaxxion: While testing the code I encountered nulls there. Not sure how exactly since I agree with IntelliJ, it shouldn't have been possible at that point. It's possible the accompanying test case was set up wrong at the time and subsequently fixed; that was one of the first methods I wrote while working on this.

ALongStringOfNumbers (approved):
Not a fan of throwing the exceptions, which will crash, instead of simply cancelling the operation, probably voiding some fluid, and then logging a `error` to the log.

However, the logic looks good and it functions as expected

Exaxxion:
The exceptions are thrown because those outcomes are exceptional circumstances. Save for some bizarre circumstance like a mod mutating its state from another thread, it clearly indicates a flaw with the implementation of the fluid tank.

It is typically safer for system integrity to fail immediately rather than blast the log with error messages (as it may occur every tick) and simply hope for the best.

src/main/java/gregtech/common/covers/CoverFluidRegulator.java
LAGIdiot: Do you need first nonNull filter shouldn't last one catch everything?

src/test/java/gregtech/common/covers/CoverFluidRegulatorTest.java
```diff
@@ -0,0 +1,227 @@
...
+				new FluidTankList(false,
+				                  new FluidTank(new FluidStack(FluidRegistry.WATER, 100), 64000),
+				                  new FluidTank(64000)),
+				new FluidTankList(false));
+
+		// accept any fluid this time
+		int amountTransferred = cfr.doKeepExact(10000, source, dest, fs -> true, 144);
+
+		// expect that 44mB of water and 144mB of lava will be moved
+		assertEquals("Wrong fluid amount moved", 44+144, amountTransferred);
```
LAGIdiot, line 139: In this case you should also check content of destination to be really sure that correct amount of both fluids was moved.

src/test/java/gregtech/common/covers/CoverFluidRegulatorTest.java
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	@Test
+	public void doKeepExact_respects_transfer_limit() {
```
LAGIdiot, line 143: I think this case can be simplified

LAGIdiot (changes requested):
New implementation looks correct and tests looks to confirm it's functionality.

Though I have some comment to code which I would like you to address.

Regarding use of exception I think it is OK in this case as these would be abnormal situation and my corrupt us later.

src/test/java/gregtech/common/covers/CoverFluidRegulatorTest.java
```diff
@@ -0,0 +1,227 @@
...
+import net.minecraft.util.*;
+import net.minecraftforge.fluids.*;
+import net.minecraftforge.fluids.capability.*;
+import org.junit.*;
+
+import java.util.function.*;
+
+import static org.junit.Assert.*;
+
+public class CoverFluidRegulatorTest {
```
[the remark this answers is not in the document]
Exaxxion: The tabs/spaces thing is probably just because it was a new file, and I got used to relying on IntelliJ's detection of existing indent and didn't notice it was different. I can fix that shortly.

src/main/java/gregtech/common/covers/CoverFluidRegulator.java
Exaxxion: As I told Proto 15 days ago, I figured I wouldn't have needed it but I got a NPE without it. I could try it again and see if it was an unrelated problem.

LAGIdiot (approved):
Thanks for addressing my comments.

PR #1772, opened 2025-02-25
Update zh_cn.lang

60TPLewandowskiego:
Replaced some wrong translations and added more.


Exaxxion:
Thank you for your interest, @60TPLewandowskiego . This project has been inactive for 3 years, with developer interest having overwhelmingly moved to [GTCEu](https://github.com/GregTechCEu/GregTech).

Few packs currently in development continue to use GTCE. Nomifactory for example has switched to using [its own fork](https://github.com/Nomifactory/GregTech) of this project. It is my understanding that Technological Journey also did the same.

